PR #2, opened 2015-12-31
3777: Creates models/tables needed for divisions

meltheadorable:
* creates organizations, currencies, and divisions
* updates factories where necessary
* updates models where necessary
* adds `closure_tree` gem to manage division hierarchies

### DIFF
--- a/app/models/division.rb
+++ b/app/models/division.rb
@@ -1,4 +1,6 @@
 class Division < ActiveRecord::Base
-  has_many :subdivisions, class_name: 'Division', foreign_key: 'SuperDivision'
-  belongs_to :super_division, class_name: 'Division', foreign_key: 'SuperDivision'
+  has_closure_tree
+  alias_attribute :super_division, :parent
+
+  belongs_to :organization
 end
--- a/db/migrate/20151231160304_create_organizations.rb
+++ b/db/migrate/20151231160304_create_organizations.rb
@@ -1,0 +1,15 @@
+class CreateOrganizations < ActiveRecord::Migration
+  def change
+    create_table :organizations do |t|
+      t.string :display_name
+      t.string :sector
+      t.string :industry
+      t.string :referral_source
+      t.float :woman_ownership_percent
+      t.float :poc_ownership_percent
+      t.integer :environmental_impact_score
+
+      t.timestamps null: false
+    end
+  end
+end
--- a/db/migrate/20151231160304_create_organizations.rb
+++ b/db/migrate/20151231160304_create_organizations.rb
@@ -5,9 +5,6 @@ class CreateOrganizations < ActiveRecord::Migration
       t.string :sector
       t.string :industry
       t.string :referral_source
-      t.float :woman_ownership_percent
-      t.float :poc_ownership_percent
-      t.integer :environmental_impact_score
 
       t.timestamps null: false
     end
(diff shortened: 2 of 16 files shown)


app/models/division.rb
```diff
@@ -1,4 +1,6 @@
 class Division < ActiveRecord::Base
-  has_many :subdivisions, class_name: 'Division', foreign_key: 'SuperDivision'
-  belongs_to :super_division, class_name: 'Division', foreign_key: 'SuperDivision'
+  has_closure_tree
```
linesarefuzzy: That's nice :)

db/migrate/20151231160304_create_organizations.rb
```diff
@@ -0,0 +1,15 @@
+class CreateOrganizations < ActiveRecord::Migration
+  def change
+    create_table :organizations do |t|
+      t.string :display_name
+      t.string :sector
+      t.string :industry
+      t.string :referral_source
+      t.float :woman_ownership_percent
+      t.float :poc_ownership_percent
+      t.integer :environmental_impact_score
```
linesarefuzzy: As we discussed today, these demographic fields will be in their own table, so we should probably leave them out here.

db/migrate/20151231160846_create_currencies.rb
```diff
@@ -0,0 +1,13 @@
+class CreateCurrencies < ActiveRecord::Migration
+  def change
+    create_table :currencies do |t|
+      t.string :code
+      t.decimal :current_exchange_rate
+      t.datetime :exchange_rate_date
```
linesarefuzzy: We may want to do the exchange rate stuff a different way. There are gems. Leave out for now? Er, maybe we need it for data migration purposes?

db/migrate/20151231161821_create_divisions.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDivisions < ActiveRecord::Migration
+  def change
+    create_table :divisions do |t|
+      t.references :organization, index: true, foreign_key: true
```
linesarefuzzy: Are database-level constraints built into Rails now?

linesarefuzzy:
Beautiful! Thanks for updating the specs and factories. You can go ahead and merge this when ready.

db/migrate/20151231160304_create_organizations.rb
meltheadorable: Yep, I created all this before the discussion on that. I'll get rid of 'em.

db/migrate/20151231160846_create_currencies.rb
meltheadorable: There are gems, but I'm not sure whether the gems actually absolve us from needing to keep track of this information (though this might be something that we need to denormalize in certain places, like transactions, so that we can account for exchange rates at specific points in time besides "the last time we updated the exchange rate"). 

I was just using the fields that were laid out in the schema document, but I can squash them for now.

db/migrate/20151231161821_create_divisions.rb
meltheadorable: It knows how to create certain ones on databases that support them. Indexes and Foreign Keys are supported. The options on that line creates these lines in `schema.rb`:

`add_index "divisions", ["organization_id"], name: "index_divisions_on_organization_id", using: :btree`

`add_foreign_key "divisions", "organizations"`

linesarefuzzy: <old man voice> Back in my day, we had to use foreigner for that </old man voice>

PR #4, opened 2016-01-05
3760 loans index [WIP]

linesarefuzzy:


app/models/loan.rb
```diff
@@ -1,20 +1,18 @@
 class Loan < ActiveRecord::Base
   include TranslationModule, MediaModule
 
-  belongs_to :cooperative, :foreign_key => 'CooperativeID'
-  belongs_to :division, :foreign_key => 'SourceDivision'
+  belongs_to :division
+  belongs_to :primary_agent, class_name: 'Person'
+  belongs_to :secondary_agent, class_name: 'Person'
+  belongs_to :currency
+  belongs_to :representative, class_name: 'Person'
```
linesarefuzzy: Shouldn't it still belong to an organization (the loan recipient)?

db/migrate/20151231182601_create_loans.rb
```diff
@@ -0,0 +1,27 @@
...
+    create_table :loans do |t|
+      t.references :division, index: true, foreign_key: true
+      t.string :name
+      t.references :primary_agent, references: :people
+      t.references :secondary_agent, references: :people
+      t.string :status
+      t.decimal :amount
+      t.references :currency, index: true, foreign_key: true
+      t.decimal :rate
+      t.integer :length_months
```
linesarefuzzy: Is this redundant with target_end_date?

spec/features/loan_index_spec.rb
```diff
@@ -1,13 +1,14 @@
...
   before { @loans = create_list(:loan, 3, :active) }
-  context 'on the home page' do
-    before { visit root_path }
+  context 'on the loan index page' do
+    before { visit loans_path }
 
     it 'shows active loans' do
-      active_loans = @loans.select{ |loan| loan.status == I18n.t(:loan_active) }
+      active_loans = @loans.select{ |loan| loan.status == 'active' }
+      expect(active_loans).to be_present
```
linesarefuzzy: What does `be_present` mean? How does that work with a model instance?

meltheadorable: `be_present` means it responds `true` to the method `.present?`, I tossed that in there so it would fail in cases where it was empty, because `.each` will just iterate gladly over an empty collection without an error.

app/models/loan.rb
meltheadorable: I could be mistaken or misremembering, but I think there's a join model that I haven't built yet that's supposed to model that relationship. I double & triple checked the schema document while I was creating all these but I could have somehow missed that one.

db/migrate/20151231182601_create_loans.rb
meltheadorable: Possibly, it might depend on whether `target_end_date` ever moves or if it's permanently fixed at the length of the loan. Presumably the number of months the loan term is for is a fixed part of the terms, a target end date could potentially be a timeline that changes as circumstances do, but I'm not sure if that's how it's used. Both fields were in the schema document I was working from.

app/models/loan.rb
linesarefuzzy: It has `organization_id` as one of the fields. I think there is always one organization that receives a loan. I don't see the business logic for a many-to-many.

spec/features/loan_index_spec.rb
linesarefuzzy: Ah.

linesarefuzzy:
Looks great. You're a rockstar. "WIP"—you have more to do on this?

app/models/loan.rb
meltheadorable: Misremembering then, no idea how I botched that one ^_^;;

meltheadorable:
Looks like I originally opened this PR for the wrong branch. I've fixed the mentioned issues with this one, but I'm going to close it and open it correctly.

linesarefuzzy:
Actually, I closed yours and opened a new one referencing the other branch as a base, to see the changes just for this issue. But no matter.

meltheadorable:
Oh! Fancy that. Either way, needed to re-open it against develop since the other branch was merged in. :smile:

PR #7, opened 2016-01-05
[3738] - migration of divisions

elliottjf:
Main omission right now unit tests.  I'll continue to work on that, but figured I'd get an initial and small code review opened up.


app/models/division.rb
```diff
@@ -3,4 +3,22 @@ class Division < ActiveRecord::Base
...
+  # end
+
+
+  # For now the id of a special system root node.
+  # Currently convient as an owning divison of migrated orgs and people, but may not be needed in the long run.
+  # Will revisit once full requirements are more clear.
+  ROOT_ID = 99
+
+
 end
```
linesarefuzzy: Great code commenting!

app/models/legacy/division.rb
```diff
@@ -0,0 +1,50 @@
...
+      data = {
+          id: id,
+          parent_id: parent_id,
+          name: name,
+          description: description,
+          # note, no well defined division currency in current model (aside from the global AR$ default)
+          # currency_id:
+      }
+      data
+    end
```
linesarefuzzy: Why bother instantiating this data var? You could just make the hash the last thing and it will be returned.

app/models/division.rb
```diff
@@ -3,4 +3,22 @@ class Division < ActiveRecord::Base
...
+  #   t.integer :parent_id
+  #   t.references :currency, index: true, foreign_key: true
+  #   t.timestamps null: false
+  # end
+
+
+  # For now the id of a special system root node.
+  # Currently convient as an owning divison of migrated orgs and people, but may not be needed in the long run.
+  # Will revisit once full requirements are more clear.
+  ROOT_ID = 99
```
linesarefuzzy: What about a class method instead, like `Division.root`?

db/migrate/20160103040147_populate_root_division.rb
```diff
@@ -0,0 +1,12 @@
+class PopulateRootDivision < ActiveRecord::Migration
+  def up
+    # note, we'll likely want to create a parent division specific to The Working World, and may not eventually
+    # need this system wide root.  Will revisit once full requirements are more clear.
+    Division.create({id:Division::ROOT_ID, name:'Root Division'})
+    Division.connection.execute("select setval('divisions_id_seq', greatest((select max(id)+1 from divisions), 100))")
```
linesarefuzzy: "100" should probably be `Division::ROOT_ID + 1`, right?

lib/tasks/migrate_data.rake
```diff
@@ -0,0 +1,112 @@
+namespace :tww do
+
+  desc "migrate TWW data from legacy mysql to new postgres db"
+  task :migrate_legacy => :environment do
+
+    Legacy::Division.migrate_all
```
linesarefuzzy: This is so elegant.

lib/tasks/migrate_data.rake
```diff
@@ -0,0 +1,112 @@
...
+    # Note.delete_all
+    #
+    # puts "notes logs: #{Legacy::Note.where('NotedTable' => 'Cooperatives').count}"
+    # Legacy::Note.where('NotedTable' => 'Cooperatives').each &:migrate
+    # Note.connection.execute("SELECT setval('notes_id_seq', (SELECT MAX(id) FROM notes)+1000)")
+
+  end
+
+  desc "purge target (postgres) data"
+  task :purge_migrated => :environment do
```
linesarefuzzy: This is for undoing the migration? Why not just nuke the DB?

app/models/legacy/division.rb
elliottjf: as a matter of style, i often introduce extra vars to make debugging easier when needed.

app/models/division.rb
elliottjf: sure.  i could use a class method.  Division.root should perhaps return the instance object, and Division.root_id  could be the id

db/migrate/20160103040147_populate_root_division.rb
elliottjf: yeah.   i had added that constant later.  i'll clean that up.

linesarefuzzy:
Looks good. Thanks for chunking :). Consider these few comments, but either way go ahead with integrating the rest of your migration stuff.

lib/tasks/migrate_data.rake
elliottjf: It's been useful to do piecemeal reversals as I've developed.  I've typically commented / uncommented different sections of this method.   But yeah, for a total wipe a db:drop db:create db:migrate is faster and more reliable

PR #8, opened 2016-01-06
3752 - translatable concern with dependent models

elliottjf:
okay. i decided to abandon my previous branch which was pulling in the wider list of loan related models since it was going to be a painful merge with the 3760 branch.  
but here is a small pull request with the minimum to get the translatable summary and details attributes working on the current Loan model.


app/models/concerns/translatable.rb
```diff
@@ -0,0 +1,99 @@
...
+    # def foo
+    #   get_translation('foo', I18n.language_code)
+    # end
+    #
+    # def foo=(text)
+    #   set_translation('foo', text, I18n.language_code)
+    # end
+    #
+    # def get_foo(text, language_code)
+    #   get_translation('foo', text, language_code)
```
linesarefuzzy: Extraneous "text" param here?

elliottjf: woops.  yeah.

db/migrate/20160101021129_populate_languages.rb
```diff
@@ -0,0 +1,10 @@
+class PopulateLanguages < ActiveRecord::Migration
+  def up
+    Language.create({id:1,name:'English',code:'EN'})
+    Language.create({id:2,name:'Español',code:'ES'})
```
meltheadorable: Why's this being done as a migration rather than in `seeds.rb`? I usually try to avoid putting seed data in the migrations, it can mess with the test suite which cleans out the database between test runs. If it were in the `seeds.rb` file that data could be required for the specs that rely on its presence.

app/models/concerns/translatable.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  included do
+    has_many :translations, as: :translatable
+  end
+
+  module ClassMethods
+
+    #
+    # define convenience methods for assigning and fetching the named attribute
+    #
```
linesarefuzzy: This is so cool. And thanks for the great documentation.

app/models/concerns/translatable.rb
```diff
@@ -0,0 +1,99 @@
...
+    hash = {}
+    translations.where({translatable_attribute: attribute_name}).each{|t| hash[t.language.code] = t.text}
+    hash
+  end
+
+  # returns all translations as an ordered list of [code,text]
+  def translations_list(attribute_name)
+    #todo: propertly sort
+    translations.where({translatable_attribute: attribute_name}).map{ |t| {code: t.language.code, text: t.text} }
+  end
```
linesarefuzzy: This seems maybe a little bit redundant with translations_map, no? Couldn't you just as easily iterate over that? `foo_map.each { | language_code, text | ... }`

db/migrate/20160101021129_populate_languages.rb
elliottjf: That would explain some oddness that I was seeing when running my unit tests earlier today!  Because the previous rails work that I've done was not test-driven.  Thanks for the pointer in the right direction and I'll fix that!

app/models/concerns/translatable.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def set_translation(attribute_name, text, language_code = nil)
+    t = get_translation_obj(attribute_name, language_code)
+    if t
+      t.text = text
+      t.save
+    else
+      language_id = Language.resolve_id(language_code)
+      translations.create(translatable_attribute: attribute_name, language_id: language_id, text: text)
```
linesarefuzzy: `resolve_id` could return nil, right? Is that handled here?

app/models/concerns/translatable.rb
```diff
@@ -0,0 +1,99 @@
...
+  # returns all translations as an ordered list of [code,text]
+  def translations_list(attribute_name)
+    #todo: propertly sort
+    translations.where({translatable_attribute: attribute_name}).map{ |t| {code: t.language.code, text: t.text} }
+  end
+
+
+end
+
+
```
linesarefuzzy: Beautiful. I love it!

app/models/concerns/translatable.rb
elliottjf: i wasn't sure would be most convenient, so i added both for now.  we can drop one version once we figure out how we're consuming, but i didn't see too much harm in providing both.

app/models/language.rb
```diff
@@ -1,2 +1,30 @@
...
+  # create_table :languages do |t|
+  #   t.string :name
+  #   t.string :code
+  #
+  #   t.timestamps
+
+
+  # returns language_id for given code
+  def self.resolve_id(code = nil)
+    code ||= I18n.language_code
```
linesarefuzzy: Why not just put `code = I18n.language_code` in the method args?

app/models/language.rb
```diff
@@ -1,2 +1,30 @@
...
+    Language.where({code: code.upcase}).first.try(:id)  #todo: use a cache to optimize
+  end
+
+  #todo: use a cache here also
```
linesarefuzzy: Do you know about memoizing using `||=`?

app/models/loan.rb
```diff
@@ -7,7 +8,12 @@ class Loan < ActiveRecord::Base
   belongs_to :secondary_agent, class_name: 'Person'
   belongs_to :currency
   belongs_to :representative, class_name: 'Person'
-  has_many :repayments, :foreign_key => 'LoanID'
+  # has_many :repayments, :foreign_key => 'LoanID'
+
+
+  # define accessor like convenience methods for the fields stored in the Translations table
+  attr_translatable :summary, :details
```
linesarefuzzy: So beautiful!

app/models/concerns/translatable.rb
elliottjf: yeah.  i'll raise an explicit exception if the language code fails to resolve.

linesarefuzzy:
Looks great. Merge away after reviewing the comments.

app/models/language.rb
elliottjf: even if 'nil' is explicitly passed in as a param.  i wanted it to default to the default language.

app/models/concerns/translatable.rb
meltheadorable: If it's turning a hash into a paired and sorted array, I actually think this could be done with `translations_map.sort_by { |code, text| code }`

app/models/language.rb
elliottjf: i didn't know that term. but i use that technique for single value caching.  in this case, i can memoize a cached map as a class var, which is then used to perform a local lookup. 
i still prefer to defer these optimization until things have stabilized a bit.

meltheadorable:
I didn't see any changes to the shared example specs for translatable, they're under `spec/support/modules/translatable.rb`, they were written for the old translation system and I'm not sure how extensive the updates to them will need to be to get them working again with the new one. 

Do you want me to handle those as part of my loan work or do you wanna try tackling updating the example group and its specs here? You'll wanna take `it_should_behave_like 'translatable'` call out of the `pending` block in `models/loan_spec.rb` to test that it's working.

elliottjf:
regarding the unit tests.  i put that on my plate and take a pass at it once i finish integrating my other outstanding work.

app/models/concerns/translatable.rb
elliottjf: I'm actually thinking that the Language.resolve_id() should raise an exception of given an invalid code.  I'm also curious if you have a convention around what kind of exception to raise.

PR #9, opened 2016-01-06
3738 modelling updates for loans and dependencies

elliottjf:
3738 - modelling updates and migration handling for Loans and the core classes they depend on: org, person, country, currency, org snapshots.  loans and core dependencies can now be imported from legacy database.


db/migrate/20160106051255_recreate_people.rb
```diff
@@ -0,0 +1,54 @@
+class RecreatePeople < ActiveRecord::Migration
+  def change
+    # note, wasn't able to just drop and re-add table because of foreign key constraints
+    # (and apparently with postgres, there is no way to temporarily suspend the constraints)
+    # drop_table :people
+    # create_table :people do |t|
+    #   ## base Contact fields
+    #   t.references :division, index: true
+    #   t.string :name   # note agreeing with mel's change of this from 'display_name' as a more consistent field to hold the full name
```
meltheadorable: That actually wasn't the change I was making.

Based on [this guide](http://www.w3.org/International/questions/qa-personal-names#fielddesign) and many similar resources, I was guessing that 

1. We might not *need* to capture first and last names separately
2. Any suspected reason to do so should be closely examined for alternatives
3. If we really do need it, we should split into given/family names instead of "first" and "last" and not require both

I left it at just "name" at creation time since it would be easy to add extra fields whenever we actually needed them.

Poorly localized name forms based around the assumptions of English-speaking software developers cause a pretty big headache for a lot of people, I'd love for us to not be the cause of headaches on this issue. 

My preference would be for two fields, `name`/`display_name`/`nickname` for what we should call the person and `legal_name`/`official_name` for the purposes of any official/government/financial paperwork. If we need to capture and store given & family name separately, then maybe we should provide a dropdown for populating `legal_name` that gives a few different options for how to compose Given & Family names (I think wordpress does this) or if none are correct allows manual entry.

app/models/country.rb
```diff
@@ -1,4 +1,20 @@
 class Country < ActiveRecord::Base
-  belongs_to :language, :foreign_key => 'LanguageID'
+
+  # create_table :countries do |t|
```
meltheadorable: I appreciate the thought behind putting the schema information in comments here, but we might just want to add the [annotate models](https://github.com/ctran/annotate_models) gem which will do this automatically. I usually add it to new projects but left it out of this one until things had stabilized a bit more because it can cause unnecessary merge conflicts in these early stages (as would manually updating these large comment blocks, which would still be better than letting them get out of date). I wonder whether it makes more sense to drop these for now until the schema settles a bit more.

spec/factories/people.rb
```diff
@@ -1,6 +1,8 @@
 FactoryGirl.define do
   factory :person do
-    name { Faker::Name.name }
+    division
+    first_name { Faker::Name.name }
```
meltheadorable: `Faker::Name` has `.first_name` and `.last_name` methods for this. `.name` will generate full names including occasional titles/qualifiers.

README.md
```diff
@@ -0,0 +1,18 @@
+# The Madeline System
```
meltheadorable: Thanks for adding this :+1:

db/migrate/20160106040041_alter_loans.rb
```diff
@@ -0,0 +1,12 @@
+class AlterLoans < ActiveRecord::Migration
+  def change
+    add_column :loans, :name, :string
```
meltheadorable: This fails because loans already has a name column.

app/models/country.rb
elliottjf: Cool.  I didn't know about the annotate_models gem.  That sounds like a great solution!

spec/factories/people.rb
elliottjf: cool.  will update!

db/migrate/20160106040041_alter_loans.rb
elliottjf: Odd.  I see now that it already exists.  Not sure why I thought it  was missing before.  It didn't seem to be fatal on my system, but I'll rip it out and make sure things are still working.

db/migrate/20160106051255_recreate_people.rb
elliottjf: Great!  Sounds like we're on the same page.  I hadn't seen a well articulated discussion of this before. but I've always personally thought a single name field was a better approach than first/last fields, but that has generally been a lonely position for me, so I didn't even bother suggesting on this project.  
We'll likely still need to figure out the best way to handle a version sortable by last name, but we can deal with that at a later iteration.

db/migrate/20160106040041_alter_loans.rb
elliottjf: Okay.  I see now that I was confused about what db:setup does versus db:create db:migrate db:seed.  So I wasn't properly regression testing the full set of migrate scripts.  Looks like that was the only problem though.
One more thought.  It seems that it would be useful in general to put system data population in both seeds.rb and migrate scripts.  That way developers don't need to worry about knowing when to reapply the db:seed step.

meltheadorable: This is part of the reason that I try to keep `seeds.rb` exceptionally minimal. If you can open the software and navigate the functionality without the data in `seeds.rb` then it shouldn't be in `seeds.rb`, and overall you *should* be able to do that. There are very few data-level dependencies that should be so deeply embedded that the software can't cope with their absence. Keeping disciplined with what goes into `seeds.rb` makes it easy to know when you should run it because most of the time it shouldn't be changing.

README.md
linesarefuzzy: :+1:

app/models/legacy/cooperative.rb
```diff
@@ -0,0 +1,57 @@
...
+  def migrate
+    data = migration_data
+    puts "#{data[:id]}: #{data[:display_name]}"
+    ::Organization.create(data)
+  end
+
+  def self.migrate_all
+    puts "cooperatives: #{ self.count }"
+    self.all.each &:migrate
+    ::Organization.connection.execute("SELECT setval('organizations_id_seq', (SELECT MAX(id) FROM organizations)+1000)")
```
linesarefuzzy: Why +1000?

app/models/legacy/translation.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def migrate
+    if [1,2].include?(language)
+      data = migration_data
+      puts "#{data[:id]}: #{data[:translatable_type]}[#{data[:translatable_id]}].#{data[:translatable_attribute]}"
+      ::Translation.create(data)
+    else
+      # note, there is a lot of orphaned translation data in the current production db
+      puts "Warning, ignoring Translation[#{id}] with invalid Language reference: #{language}"
```
linesarefuzzy: Looks like the 3s are all French. We can keep those (when the time comes). Not sure what to do about the 6s. They're all mixed up.

app/models/legacy/loan.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+
+  def self.migrate_all
+    puts "loans: #{self.count}"
+    self.all.each &:migrate
+    ::Loan.connection.execute("SELECT setval('loans_id_seq', (SELECT MAX(id) FROM loans)+1000)")
+
+    puts "loan translations: #{ Legacy::Translation.where('RemoteTable' => 'Loans').count }"
+    Legacy::Translation.where('RemoteTable' => 'Loans').each &:migrate
```
linesarefuzzy: Why not migrate all the translations at once?

app/models/legacy/translation.rb
```diff
@@ -0,0 +1,53 @@
...
+  ATTRIBUTE_MAP = {
+      'Loan' => {'ShortDescription' => 'summary', 'Description' => 'details'},
+      'ProjectStep' => {'Summary' => 'summary', 'Details' => 'details'},
+      'ProjectLog' => {'Explanation' => 'summary', 'DetailedExplanation' => 'details',
+                       'AdditionalNotes' => 'additional_notes', 'NotasPrivadas' => 'private_notes'}
+  }
+
+end
+
+end
```
linesarefuzzy: Your migration stuff is beautiful. Very elegant. 

I'm not too worried about the location of these files. We can move or remove them once we're up and running.

app/models/loan.rb
```diff
@@ -1,20 +1,139 @@
...
+    LOAN_TYPE_OPTIONS.label_for(loan_type_option_id)
+  end
+
+  # the special name of a default step to use/create when migrating a log without a step
+  DEFAULT_STEP_NAME = '[default]'
+
+  # creates / reuses a default step when migrating ProjectLogs without a proper owning step
+  # beware, not at all optimized, but sufficient for migration
+  def default_step
+    step = project_steps.select{|s| s.summary == DEFAULT_STEP_NAME}.first
```
linesarefuzzy: You could probably use `find_or_create_by` here

app/models/loan.rb
```diff
@@ -1,20 +1,139 @@
...
+    unless step
+      logger.info {"default step not found for loan[#{id}] - creating"}
+      step = project_steps.create
+      step.update({summary: DEFAULT_STEP_NAME})
+    end
+    step
+  end
```
linesarefuzzy: Why isn't this in Legacy::Loan?

app/models/legacy/cooperative.rb
elliottjf: I was just creating some space between the legacy data and new data.

app/models/legacy/translation.rb
elliottjf: Yeah, the 3s I looked at were French, but I didn't do a complete review.  It's odd that There is nothing anymore in the legacy Language table above 2.  Shall I just go and add a record in the seeds.rb for 3->French?

app/models/legacy/loan.rb
elliottjf: it's been useful during development to break them abart.  the project log translations are hefty and take about 10 minutes on my system.

app/models/loan.rb
elliottjf: cool.  i didn't know about that method at the time.  i'll update!

app/models/loan.rb
elliottjf: I thought it might have some generic use as the requirements evolve, since it was a use case that the old system supported..  But if not, then yep, it doesn't really need to live here.

app/models/option_set.rb
```diff
@@ -0,0 +1,35 @@
...
+  def value_for(label)
+    @label_map[label]
+  end
+
+  def label_for(value)
+    @value_map[value]
+  end
+
+
+  def initialize(value_label_pairs)
```
linesarefuzzy: initialize should probably come before other method defs

app/models/option_set.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def build_label_map(options)
+    result = {}
+    options.each {|option| result[option.label] = option.value}
+    result
+  end
+
+
+
+end
```
linesarefuzzy: Nice. So this system will be easily convertible to DB storage if we decide to go that way.

app/models/organization.rb
```diff
@@ -1,2 +1,58 @@
...
+  has_many :organization_snapshots  # all historical data
+
+  validates :name, presence: true
+  validates :division_id, presence: true
+
+
+
+  def loans_count
+    loans.size
+  end
```
linesarefuzzy: Seems a bit superfluous.

app/models/organization.rb
```diff
@@ -1,2 +1,58 @@
...
+  #   t.string :sector
+  #   t.string :industry
+  #   t.string :referral_source
+  #   t.references :organization_snapshot
+
+
+  belongs_to :division
+  belongs_to :country
+  belongs_to :primary_contact, class_name: 'Person'
+  belongs_to :organization_snapshot # the latest data
```
linesarefuzzy: This should probably be `has_one`, shouldn't it?

app/models/option_set.rb
elliottjf: Exactly.  I think the interface can remain stable and we can choose to shift to a db backed version for the instances which warrant customization.

db/seeds.rb
```diff
@@ -14,3 +15,15 @@
 Language.find_or_create_by({id:2,name:'Español',code:'ES'})
 Language.connection.execute("select setval('languages_id_seq', (select max(id) from languages))")
 
+
+Currency.find_or_create_by({id:1,name:'Argentinean Peso',code:'ARS',symbol:'AR$'})
+Currency.find_or_create_by({id:2,name:'U.S. Dollar',code:'USD',symbol:'US$'})
+Currency.find_or_create_by({id:3,name:'British Pound',code:'GBP',symbol:'£'})
```
linesarefuzzy: I know it wasn't so in the old db, but we should be consistent with the symbols.

app/models/organization.rb
elliottjf: Conceptually, I was thinking of this as a calculated attribute.  It's used in the list view, and at some point we'll probably allow for those views to be customized.  We'll also need to think about how we want to handle sorting the list views by different columns.  Elastic search may be useful as we can easily index on arbitrary calculated values.

app/models/organization.rb
elliottjf: Doesn't 'has_one' place the key on the other table?  in reality organization 'has_many' snapshots (which is also represented a few lines down)
My intention of the 'belongs_to' is an explicit column on the organization table pointing to the latest record.  This is slightly denormalized as we could do a query fetch one sorted by date.  But if we want to query against the org table by the join, it would be handy to have an easy join column.  This also leads to what kind of sorting and filtering we need to support in the various list views, and how we want to approach that.

app/models/legacy/translation.rb
linesarefuzzy: I'd say so

linesarefuzzy:
This is great, Joseph! Thanks so much for following the process, and welcoming constructive feedback. Go ahead and merge when ready.

app/models/loan.rb
elliottjf: Actually, the translatable summary field makes this tricky to optimize.  I opted for simple code over performance

PR #10, opened 2016-01-06
3738 modelling for project steps, logs, notes w/ migration and basic tests

elliottjf:


spec/factories/notes.rb
```diff
@@ -0,0 +1,8 @@
+FactoryGirl.define do
+  factory :note do
+    notable { create(:person) }
```
meltheadorable: Ordinarily, you'll be creating this factory the other way around, you're more likely to want to create traits like `with_notes` on the person factory than to want to create a note that's already attached to something like this.

app/models/note.rb
```diff
@@ -0,0 +1,21 @@
...
+  include ::Translatable
+
+  # create_table :notes do |t|
+  #   t.references :notable, polymorphic: true, index: true
+  #   t.references :person, index: true
+  #   t.timestamps
+
+
+  belongs_to :notable, polymorphic: true
+  belongs_to :person
```
meltheadorable: Is this person the author/creator of the note? If so, it might make more sense to stick to the more specific terminology, especially since people are also notable.

spec/factories/notes.rb
```diff
@@ -0,0 +1,8 @@
...
+    person
+# need to make sure parent saved before assigning this
+    # text { Faker::Lorem.sentence }
```
meltheadorable: You should be able to handle this in a callback block, something like:

```
after(:create) do |note|
  note.text = Faker::Lorem.sentence
  note.save
end
```

spec/factories/project_logs.rb
```diff
@@ -1,13 +1,21 @@
...
-    detailed_explanation { Faker::Lorem.paragraphs(3) }
-    explanation { Faker::Lorem.sentences(3) }
-    notas_privadas { Faker::Lorem.paragraph }
-    paso_id { create(:project_event).id }
-    progress_metric { create(:progress_metric).id }
-    project_id { create(:basic_project).id }
-    project_table 'BasicProjects'
+    progress_metric_option_id { ProjectLog::PROGRESS_METRIC_OPTIONS.values.sample }
+
+    # need to make sure parent saved before assigning these
```
meltheadorable: Same as the notes,  you can use an `after(:create)` block to handle these

spec/factories/project_steps.rb
```diff
@@ -0,0 +1,49 @@
...
+    # project_table 'Loan'
+    # project_id { create(:loan).id }
+    project { create(:loan) }
+    person
+    # need to make sure parent saved before assigning these
+    # summary { Faker::Lorem.sentence }
+    # details { Faker::Lorem.paragraph }
+    scheduled_date { Faker::Date.between(Date.civil(2004, 01, 01), Date.today) }
+    completed_date { Faker::Date.between(Date.civil(2004, 01, 01), Date.today) }
+    is_finalized [true,false].sample  ##JE todo: is ther a better pattern to assign a boolean?
```
meltheadorable: I think this is pretty much the best way to randomly assign booleans, though as a silly nit-picky style thing I think there should be a space after the comma :stuck_out_tongue_closed_eyes:

spec/factories/project_steps.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    trait :future do
+      scheduled_date { Faker::Date.forward }
+    end
+
+    # is this even valid?
+    # trait :for_loan do
+    #   transient { loan_id 0 }
+    #   after(:build) do |event, evaluator|
```
meltheadorable: Believe it or not, this did work. It doesn't really have to work anymore though, it was built in to save space & energy on the calling side not having to manually set the correct table and stuff manually. The `for_loan: loan.id` calls are probably just going to need to be replaced, wherever they show up, with `project: loan` calls now that it's a proper association and this stuff can be deleted. 

If we wanted to keep the `for_loan` syntax though, it wouldn't look that much different:

```ruby
transient { loan_id 0 }
after(:build) do |step, evaluator|
  step.project = Loan.find(loan_id)
end
```

spec/models/note_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'rails_helper'
+
+describe Note, :type => :model do
+  it 'has a valid factory' do
```
meltheadorable: Might want to add the `it_should_behave_like 'translatable', ['text']` in here

spec/models/project_log_spec.rb
```diff
@@ -1,9 +1,12 @@
 require 'rails_helper'
 
 describe ProjectLog, :type => :model do
-  before { pending 're-implement in new project' }
-  it_should_behave_like 'translatable', ['Explanation', 'DetailedExplanation']
-  it_should_behave_like 'mediable'
```
meltheadorable: When you update this to take care of the merge conflicts, etc, you should be able to put this line back, but with `media_attachable` since that exists and is merged in now.

meltheadorable:
Overall, this looks really good! I do wanna take one more quick look at it after you've resolved the conflicts tomorrow and made the other changes, just so that I can make sure I didn't miss anything in the noise from all the already-merged commits, but I'm sure it'll be fine. :smiley_cat:

spec/factories/notes.rb
elliottjf: Okay.  I removed the notable from the notes factory.  But this leaves the default factory invalid and prevents being able to automatically test the translatable concern.  Is there a different pattern that I should be following here?

app/models/note.rb
elliottjf: Okay.  I flipped it back.  I also changed the generic 'person' relations on ProjectStep and ProjectLog to 'agent'  (I decided that I didn't like 'contact' there)
Additionally added the foreign key constraints.

spec/factories/notes.rb
elliottjf: Cool!  I used that to fix a few other places also.

spec/factories/project_steps.rb
elliottjf: Yeah, the production db was also full of references to 0 id loans.  I removed the commented code because I don't believe this is valid moving forward.  Also, ProjectSteps are inherently 'for_loan' now, so no need for a distinct trait.

spec/models/note_spec.rb
elliottjf: Tried, that except it won't work without a valid default factory.  (tested that it worked though if factory is restored to allow default notable)

elliottjf:
Note.  I merged develop and pushed that back up to the branch.  But it still reports here with a conflict.  (Note, I didn't not yet follow 'step 2' of the github instructions around resolving, as I didn't want to merge back up to develop yet.  Is there something else that I should do at this stage?)

spec/factories/notes.rb
meltheadorable: Nope, I wasn't aware of that, sorry. Carry on!

spec/factories/project_steps.rb
meltheadorable: It doesn't actually set the loan's id to 0 there, a transient attribute like that is sort of a placeholder, you need to provide a default value, I chose zero, but what it does is allow you to pass an argument for a virtual attribute in that will be evaluated and processed, in this case a loan id. 

ProjectSteps were inherently for loans then as well, it was just more difficult to construct the association in the factories because it didn't follow the rails conventions, which required manually setting the tables and ids rather than leveraging a polymorphic association setter. Like I said, we don't need this anymore but it'd be good to take note of how the transient attributes are used because it's an incredibly useful pattern for creating temporary virtual attributes used in assigning other values.

spec/models/note_spec.rb
meltheadorable: Okay! Keep the default notable. Sorry for the bad feedback :flushed:

PR #11, opened 2016-01-06
3753: Media Attachable

meltheadorable:
* Renames and modifies `mediable` spec example group to create `media_attachable`
* Adds image for spec assets
*  Changes references to `mediable` to `media_attachable`
* Updates factory for media model
* Adds `carrierwave`, `mini_magick`, and `awesome_print` to Gemfile
* Adds media model
* Makes `media` uncountable to avoid inflection weirdness in rails
* Adds media item uploader
* Creates `MediaAttachable` concern
* Adds upload directory to `.gitignore`


Gemfile
```diff
@@ -51,6 +55,9 @@ group :development, :test do
   # Report number of queries in server log
   gem 'sql_queries_count'
 
+  # Better console printing
+  gem 'awesome_print'
```
linesarefuzzy: :+1:

app/uploaders/media_item_uploader.rb
```diff
@@ -0,0 +1,91 @@
...
+  #TODO: Identify appropriate dimensions for these versions
+  version :small
+  version :medium
+  version :large
+
+  # the directory where uploaded files will be stored.
+  def store_dir
+    owner_type = model.media_attachable_type.underscore
+    owner_id = model.media_attachable_id
+    path = File.join('uploads', Rails.env, owner_type, "#{owner_id}", "#{model.id}")
```
linesarefuzzy: Why the double quotes?

app/uploaders/media_item_uploader.rb
```diff
@@ -0,0 +1,91 @@
...
+      model.kind = 'file'
+    end
+  end
+
+  def set_height_and_width_on_model
+    if file && model
+      model.item_width, model.item_height = ::MiniMagick::Image.open(file.file)[:dimensions]
+    end
+  end
+
```
linesarefuzzy: What's the purpose of setting all these things on the model?

linesarefuzzy:
Looks great. Merge when ready.

app/uploaders/media_item_uploader.rb
meltheadorable: It was either that or an explicit cast to string, I decided the interpolation was cleaner.

app/uploaders/media_item_uploader.rb
meltheadorable: By default carrierwave only persists the path to the image, when I needed to opt for it over Paperclip I needed to set and persist them manually. I don't know whether they are or will need to be used, but they were present in the schema doc and most are also present by default in Paperclip do I created them here too.

PR #13, opened 2016-01-07
2213 calendar

emitche:
Adds basic calendar to site


app/assets/javascripts/admin/calendar/dependencies.js
```diff
@@ -0,0 +1,2 @@
+//= require moment
+//= require fullcalendar
```
linesarefuzzy: Why have these dependencies in a separate file? You can put them at the top of `calendar.js.coffee`—just change the syntax to `#=` for coffeescript.

app/assets/javascripts/admin/calendar/calendar.js.coffee
```diff
@@ -0,0 +1,2 @@
+$ ->
+  $('#calendar').fullCalendar()
```
linesarefuzzy: Yay coffeescript!

linesarefuzzy:
Merge away after minor change suggested above. Thanks!

app/assets/javascripts/admin/calendar/dependencies.js
emitche: That makes sense. Will make that change.

PR #18, opened 2016-01-07
3738 - integrate admin raw crud views

elliottjf:
this provides a simplistic interface to the various model classes and relations.
factored to utilize a genericized controller implementation making it easy to add new models and change global behaviors.
each controller can be parameterized by controller namespace and base view path.  this will potentially let use one set of views for an internal admin interface, while leveraging much of the base controller infrastructure for the pretty views.


linesarefuzzy:
Hi Joseph. This branch is showing a few conflicts with develop. Can you merge develop in and resolve those conflicts when you get a chance? Thanks.

elliottjf:
okay.  latest from develop merged now

app/models/division.rb
```diff
@@ -21,7 +21,30 @@ class Division < ActiveRecord::Base
...
+
+  #JE: I like to keep a reference for the implicit db attributes here in the model class
+  # create_table :divisions do |t|
+  #   t.references :organization, index: true, foreign_key: true
+  #   t.string :name
+  #   t.text :description
+  #   t.integer :parent_id
+  #   t.references :currency, index: true, foreign_key: true
+  #   t.timestamps null: false
+  # end
```
linesarefuzzy: This is obsoleted by the gem.

config/environments/development.rb
```diff
@@ -47,4 +47,8 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+
+  Slim::Engine.set_options pretty: true, sort_attrs: false
```
linesarefuzzy: Why pretty: true?

elliottjf: It makes 'view source' html much more readable

app/controllers/concerns/base_crud_controller.rb
```diff
@@ -0,0 +1,275 @@
...
+
+#todo: consider base class vs mixin
+class BaseCrudController < ApplicationController
+  include BaseCrudHelper
+
+  layout 'raw_crud'
+
+
+  def clazz
+    raise "not implemented - clazz"
```
meltheadorable: You could implement this as `controller_name.classify.constantize`, which will raise a `NameError` on failure, and it would let you drop all the overrides in each of the other controllers. Not a big deal, but it's also a pretty easy and low-impact change.

app/models/division.rb
meltheadorable: Just make sure to yank the schema comment here since we're automating those now.

app/controllers/concerns/base_crud_controller.rb
elliottjf: cool!

meltheadorable:
Double check your `TODO:` comments, see if they're all still `TODO` or whether they've been resolved, then once you've done that and cleaned out the schema annotation comment I think you're good to go ahead and merge! :+1:

elliottjf:
Okay. Merged.  
One note. In general my 'todo' comments are 'todo later' pending discussion or wanting to wait until branches are stabilized and better unit test coverage is in place.  I use 'fixme' as a tag for things that I want to make sure to address before submitting.  
Is there a another convention that I should use for such future task code comments that is more consistent with your style?

PR #19, opened 2016-01-08
3760 loans index & translatable adjustments

meltheadorable:


meltheadorable:
Looks like I need to fix the wordpress stuff before this will work! Hold off on it for now.

app/controllers/concerns/wordpress_embeddable.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  included do
+    before_action :update_template
+    helper_method :get_division_from_url
+  end
+
+  def get_division_from_url
+    @get_division_from_url ||= Rails.configuration.x.wordpress_template[:division_urls].select { |key, val|
+      request.url.match key
+    }.values.first || default_division
```
hooverlunch: Style guide advises against multiline chaining. Could use `tap` here maybe?

meltheadorable:
closing because this will probably be split into more than one new PR once it's changed

PR #20, opened 2016-01-11
3738 - persisted option sets

elliottjf:
here is a first pass at the persisted option set implementation.
wanted to get feedback before implementing the test specs and rest of the seed data.


db/migrate/20160111013907_create_option_sets.rb
```diff
@@ -0,0 +1,11 @@
+class CreateOptionSets < ActiveRecord::Migration
+  def change
+    create_table :option_sets do |t|
+      t.references :division, index: true, null: false, foreign_key: true
+      t.string :model_type
```
meltheadorable: Is there a reason you aren't leveraging polymorphic associations here? You can see how that's done in the [media-attachable migration](madeline_system/db/migrate/20160106145757_create_media.rb). Should save you a lot of energy re-building methods rails provides in the concern.

db/seeds.rb
```diff
@@ -6,31 +6,68 @@
...
 # Person.find_or_create_by({id:0, name: 'dummy', first_name: 'dummy', division_id: Division.root_id})
+
+# handy temporary change for debugging
+# Option.destroy_all
+# OptionSet.destroy_all
+
+loan_status = OptionSet.find_or_create_by(division: Division.root, model_type: Loan.name, model_attribute: 'status')
+# beware rerunning these different position values, will result in duplicated data
+# a more verbose approach would be needed if that is important to support
+# also we can consider dropping the 'position' assignments until an ordering different than 'value' is needed
```
meltheadorable: `first_or_initialize(value: 1).update_attributes(position: 1).save` should let you do this without the duplication

db/seeds.rb
```diff
@@ -6,31 +6,68 @@
...
+loan_status.options.find_or_create_by(value: 1, position: 1).set_label_list(en: 'Active', es: 'Prestamo Activo')
```
meltheadorable: Is there a reason we aren't just using string slugs for the values (i.e. 'active', 'completed', 'frozen', 'relationship_active')? There doesn't seem to be any reason to make these values unreadable, and using slugs would eliminate the need to query through the translation table or hardcode IDs into scoping methods for filtering.

spec/factories/loans.rb
```diff
@@ -55,11 +55,12 @@
 
 
     trait :active do
-      status_option_id Loan::STATUS_OPTIONS.value_for('Active')
+      # note: can't seem to count on the seed.db data having been loaded when these are executed
+      status_option_id 1 #Loan.status_option_set.value_for_label('Active')
```
meltheadorable: Ideally the factories don't need to know the meaning of hardcoded ID values in the database in order to function correctly. Doing it this way means that this factory now has an implicit dependency on specific data being in the `Option` and `OptionSet` tables and they're not referenced in the factory at all. 

There's a few different ways we can handle this, but they all involve making sure that before these traits are used the loan factory has a valid reference to an `OptionSet`. You can see [the changes I made to the loan factories for translations](/madeline_system/blob/3760_loans_index/spec/factories/loans.rb#L67) as a starting point, but it would probably be helpful to look over some other stuff in #19 in general to see how these option sets are being used both in the specs and in the controller and see if there's a way to closely replicate the interface they're using since we'll need it for more complex filtering later.

spec/rails_helper.rb
```diff
@@ -50,6 +50,8 @@
 
   config.before(:each) do
     DatabaseCleaner.strategy = :transaction
+    # note, it's useful to temporarily switch this to 'truncation' when attempting to debug database oddities during testing
+    # DatabaseCleaner.strategy = :truncation
```
meltheadorable: If you add another block like the `before(:each, js: true)` one for `before(:each, truncate: true)` you can tag a spec with `:truncate` to get this behavior without changing the strategy globally (either temporarily for debugging purposes, or because there's a reason it's needed).

db/migrate/20160111013907_create_option_sets.rb
elliottjf: is there built in polymorphic handling against just a class, (there is not associated mode instance in this case, which is the typical pattern)

db/seeds.rb
elliottjf: my understanding is that the set of options are potentially customized per division.  we can potentially use an 'internal_name' style string instead of an id for the 'value'.  when need to think about the implications of that when designing the ability for end users to customize these lists.  
@linesarefuzzy , can you weigh in here with how you want to model this?

db/migrate/20160111013907_create_option_sets.rb
meltheadorable: Ah, yeah, I feel silly right now. This is what I get for doing code reviews late in the day. I need to think more about how this works.

spec/factories/loans.rb
elliottjf: It's definitely undesirable to have hard coded id refs.  That's why I first tried to implement using the value_for_label() logic.  What I ran into was the factory seems to be getting triggered before the seeds.rb was executing.  

I'm thinking now that there should be a way to better handle the seeds.rb invoke generically from the spec/rails_helper.rb.  Thoughts?

meltheadorable: The factories shouldn't have any dependencies on the `seeds.rb` data, ideally. What I'd like to see is something like

```ruby
after(:create) do |loan|
  option_set = create(:option_set, model_type: 'Loan', model_attribute: 'status')
  create(:option, value: 'active', option_set: option_set)
  create(:option, value: 'completed', option_set: option_set)
end

trait :active do
  status_option_id  Loan.status_option_set.options.find_by(value: 'active')
end
```

So that the factory can create everything it needs without relying on any particular database state, but the fact that it's so hard to query through this structure for pretty routine needs is making me kind of apprehensive about the approach, so I sort of want to think on this more.

elliottjf: Yeah.  I hear you.  The functional implementation was easy enough.  But the unit testing has been a can of worms.  I'll let you and @linesarefuzzy sync up and decide what makes the most sense.

app/models/loan.rb
```diff
@@ -118,23 +121,23 @@ def self.status_active_id
...
       [ [1, 'Conversion'],
         [2, 'Expansion'],
         [3, 'Start-up'],
       ])
 
-  PUBLIC_LEVEL_OPTIONS = OptionSet.new(
+  PUBLIC_LEVEL_OPTIONS = TransientOptionSet.new(
       [ [1, 'Featured'],
         [2, 'Hidden'],
       ])
```
linesarefuzzy: Are these TransientOptionSets only for migration? If so, can they live in Legacy::Loan?

config/application.rb
```diff
@@ -36,3 +36,6 @@ class Application < Rails::Application
     config.active_record.raise_in_transactional_callbacks = true
   end
 end
+
+puts "Rails.env: #{Rails.env}"
+puts "database: #{MadelineSystem::Application.config.database_configuration[::Rails.env]['database']}"
```
linesarefuzzy: For debugging? Remove?

db/seeds.rb
linesarefuzzy: Mel is going to think on this structure a bit more and see if she can come up with an improvement. My inclination at this point is that we should try to make the code agnostic of any particular options, including "active" and "completed", and to the extent that they need to be treated as special, they should be marked as special in the database. But I'm holding off on recommending that change until I hear back from Mel. 

I am concerned about this `value` column, though. I don't understand why it's there, or why you are specifying its values. I prefer not to set IDs explicitly either, but since you're doing that above anyway I don't know why you're treating these differently. If `value` is just an integer, why not use ID and leverage rails associations? If it's meaningful, why not make it meaningful, like with slugs/internal names? Again, I lean toward the former.

db/seeds.rb
```diff
@@ -6,31 +6,68 @@
...
 Division.connection.execute("select setval('divisions_id_seq', greatest((select max(id) from divisions), 100))")
 
-# note, these hardcoded id's are needed to match the migrated data
-Language.find_or_create_by({id:1,name:'English',code:'EN'})
-Language.find_or_create_by({id:2,name:'Español',code:'ES'})
-Language.find_or_create_by({id:3,name:'Français',code:'FR'})
+
+Language.find_or_create_by(id: 1, name: 'English', code: 'EN')
+Language.find_or_create_by(id: 2, name: 'Español', code: 'ES')
+Language.find_or_create_by(id: 3, name: 'Français', code: 'FR')
```
linesarefuzzy: I'm wondering out loud (as it were) if there's an easy way to avoid having to specify IDs. Perhaps using something like an old_db_id column and using that to match associations in the migration code. On the other hand, I'm also wondering if it's a big enough deal to bother setting that up. It definitely smells bad to specify IDs in the code, even in seeds I think. But I'm having trouble thinking of an actual practical problem that might arise from it.

spec/factories/loans.rb
linesarefuzzy: I agree it's not ideal for the specs to rely on seeds. Here's a partial explanation of why:
http://mikepackdev.com/blog_posts/13-seeding-data-to-your-rails-tests-with-factory-girl
It's basically the same argument as factories vs fixtures. If you haven't seen that:
https://semaphoreci.com/blog/2014/01/14/rails-testing-antipatterns-fixtures-and-factories.html
Trust me, I feel your pain. My least favorite part of TDD/BDD is writing code to generate fake data. But it's the better way to do it. And it does make it easier and nicer if you leverage FactoryGirl. The code sample Mel gave you looks really good to me. As I said above, however, I'm leaning toward removing the dependency on these `trait`s altogether, which would make it moot. :cow2:

spec/models/option_set_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  it 'has a valid factory' do
+    expect(create(:option_set)).to be_valid
+  end
+
+  it 'can fetch instance from including class' do
+    expect(OptionSet.fetch(Loan, :status)).to be
+  end
+
+  context 'seed data' do
```
linesarefuzzy: I don't think we should be testing the seed data. We don't need to. Seed data is for setting up a new environment. A failure in seeds shouldn't cause problems in production.

app/models/concerns/option_settable.rb
```diff
@@ -0,0 +1,117 @@
...
+    end
+
+  end
+
+  def option_label(attribute_name, language_code = nil)
+    self.class.resolve_option_label(self, attribute_name, language_code)
+  end
+
+
+end
```
linesarefuzzy: I love your metaprogramming. It can be easy to get carried away with that and make things overly complicated, but your code is clean, elegant, very well documented, and generally mirrors ruby/rails conventions. Awesome.

app/models/transient_option.rb
```diff
@@ -0,0 +1,10 @@
+class TransientOption
```
linesarefuzzy: Maybe these Transient models should live in legacy too, unless you foresee them being used for something in the new app.

db/seeds.rb
elliottjf: I don't like the numeric values either.  It just happened to be the simplest approach given the need to migrate data.  Here is my suggestion of how to change:
1) make `value` a string value instead of an integer and use meaningful slugs for any special system configuration data populated by seeds.rb
2) in the future when we have user created custom option values, then we'll just automatically use the ID as the value
3) we add a `migration_id` field used for resolving the correct slug value when performing the data migration

This adds a little bit of complexity to the data migration, but I think that it's worthwhile in the interest of a better designed system in the long run.  I believe that there is legitimate business logic to be attached to the "active" and "completed" statuses.  But if that logic is now tied to meaningful slugs instead of arbitrary ids, then it feels clean to me.

spec/factories/loans.rb
elliottjf: From my perspective there are two flavors of seed data.  And it's important to distinguish them.  Some of the data, like the `Currency`s and `Country`s are simply the most convenient way to handle the data migration.  It's much easier to insert a few static rows of data instead of writing business logic to migrate those from the old database.  (Building on my suggestion above, we can consider adding a `migration_id` column and remove any dependency on the primary key.  This is cleaner conceptually, but does add some complexity which I'm not entirely sure is justified, and is really a separate discussion from the thread here)  I completely agree that there should be no unit test or code dependency on data created for migration purposes that has no other specially business logic dependency.

But when there is data which is integrally tied to business logic, then I think it is legitimate for the unit tests to expect and depend on that data already existing.  I think this applies to the "active" and "completed" loan statuses.  The "root" division, a default Language (because the default translatable behavior fails with out that), and possibly a few of the other option sets like `public_level` if we have code dependencies on the "featured" and "hidden" values.

I think this set of data can be kept very minimal, left in seeds.rb and be automatically included automatically through the `spec/rails_helper.rb` logic.

Any other data (which includes the majority of what is currently in `seeds.rb` should be moved to a separate file and invoked as part of the data migration.

I think this balances things out.  We avoid the undesirable bloating of the seeds.rb as my current approach was leading us toward, but we avoid unnecessarily adding complexity and duplication to the unit test logic when there legitimate dependency on system data.  In this case, it feels more appropriate and simpler for the unit test to verify that the truly system dependent data was properly created by the seeds.rb than to duplicated its creation.  This last point is a personal preference and I don't feel too strongly about if you really think that the creation of that data should be duplicated when needed by various parts of the unit test code.

app/models/loan.rb
elliottjf: These are just the OptionSets which I had not yet converted to the persisted version, pending a finalization of the implementation approach.  Once this is complete, then we'll likely just drop the TransientOptionSets, except perhaps where it's needed for the migration, in which case I'll move over to the migration specific classes.

config/application.rb
elliottjf: I've found it very useful to have a visual confirmation of the rails env and current database during startup.  I've been burned a few times when I'd been inadvertently operating from an environment or database other than I had realized.

db/seeds.rb
elliottjf: I touched on this in the thread above.  My inclination is to only keep the English (and maybe Spanish) Language rows in the seeds.rb and to remove all of the Country and Currency entries to a migration specific file.  We could theoretically introduce a `migration_id` field on these tables, but I don't the complexity there is really justified, especially with most of that data removed from seeds.rb.

elliottjf:
Okay.  I've checked in a rework which incorporates most of what proposed in the earlier thread.  Main remaining detail to confirm is if we prefer to store the string 'slug' into a Loan.status_option_value  style column or the numeric primary key into a Loan.status_option_id style column

spec/factories/loans.rb
meltheadorable: I agree with moving the majority of what's currently in `seeds.rb` out somewhere else (I think that's what I recommended way back when it started being populated), but I'm not sure about seeding the options/option sets, I'm still trying to think through an alternative way to model this, we have a lot of different constraints on these that aren't being met well by the current schema. 

Sometimes they're just arbitrary options for populating a dropdown, sometimes they're filterable, sometimes you need to be able to filter by multiple at once, sometimes you need to filter by default, sometimes you only want a subset of the option set to be filterable by the public, but you want the full range filterable by the admin users, sometimes they're potentially security-sensitive (like public/hidden). It might be that we're dealing with two or more *different* kinds of things and we just need to split them up between which options are arbitrary and can be stored in a generic table and which have too much business logic to be meaningful in a vacuum.

Seeding the root division is okay, but I actually think we can get by without the rest of it. Even for the root division, once we start building out the specs and figure out what actual domain functionality is tied to that conceptually, we'll probably have a better idea of how to best deal with that in test suites. It might involve making sure it gets seeded from the `seeds.rb`, but it's possible that won't be necessary and that we'll want to do something more specific with that. In the meantime, all the specs that rely on a division are generating one via association, so until there's a need to test specifically for the *root* division (this will probably come when we're generating policy specs for authorization, and related functionality), we're probably fine to leave it out of the specs.

linesarefuzzy: I talked to Tom and he agrees that it’s bad practice for tests to rely on seed data. He says he’s never seen anyone do that, and it’s likely to lead to problems that could be very hard to debug. I'm curious about what bugs you say you've been encountering due to *not* using the seed data—it seems more likely to me that using it that would cause bugs. Just as one example, if you're only seeding the db once before running the test suite, and then someday you come across a test that needs to use truncate rather than transaction, the tests that run after that one may not have all the seed data. That sounds like a nightmare to debug. 

Using factories is a best practice, and it shouldn't take so much more time if we design it right. Mel's code sample is a good start, but as written it would create an additional 'active' status in the db each time an additional loan was created within the same spec. Tom suggested using a pattern similar to elmo, where at the top of the missions factory is a method that creates a new mission only if one isn't already there:

    def get_mission
      Mission.first || create(:mission)
    end

Then in a factory that needs an associated mission, `mission { get_mission }`.

That method is used all over the specs. In that case, it's only important that there be one mission. In our case it would likely need to specify certain attributes, but the same principle holds. Maybe something like this:

    def get_loan_status(status)
      Option.where(value: status) || create(:option, value: status)
    end

Using the simplified translation system Mel suggested would also remove some of the need for depending on database content, thankfully. 

So, let's please consider this settled and move on. Plan to do whatever it takes to avoid relying on seed data in the tests. Joseph, if you're ever having trouble figuring out the best way to generate the data you need (or just want to commiserate about how much generating fake data sucks), I or I'm sure Mel would be happy to help. 

I hope there are no hard feelings. You're both doing great work, you both made good arguments, and the answer here was not obvious. This kind of debate makes our products better (and me smarter :). 

On a related note, I think I've also come around to agree with both of you that certain options, like 'active', are necessary business logic and should be hardcoded in, and that we should use slugs to reference them. I suggest we put a `system` boolean field on the options table to signify options that the code relies on and thus should never be deleted.

elliottjf: Thanks for taking the time to get Tom's input.  Definitely important to move forward, so considered settled.  I'll sort out how to work most effectively within these constraints.  One more point to clarify.  It sounds like, in addition to using a generic hook which always guaranteed the availability of the seed data, that invoking the seed_data on an ad hoc basis just for the tests which needed it also considered a bad practice?

Is the Elmo codebase something which is open source or some that I can look at for more reference?  The code example around the `get_loan_status` that provided isn't quite relevant.  The challenge with the OptionSets is that they are essentially singleton objects (with dependent data) associated with the class.  But maybe something like this:
```
def ensure_loan_status_option_set
  loan_status = OptionSet.find_or_create_by(division: Division.root, model_type: Loan.name, model_attribute: 'status')
  loan_status.options.find_or_create_by(value: Loan::STATUS_ACTIVE_VALUE).set_label_list(en: 'Active', es: 'Prestamo Activo')
end
```
And put that in a `before` block for the various sections of code which depend on this configuration data.  This is essentially putting little pieces of the seeds.rb into helper methods which can be invoked when needed for very specific pieces of test code or factory setters.  Is that sufficiently capturing the spirit of your suggestion?

linesarefuzzy: Yes, relying on seeds ad hoc is also bad practice. 

The idea of using the above pattern as opposed to `find_or_create_by` is that it leverages the factories, so you don't have to specify e.g. division every time. `set_label_list` could also be invoked in the factory. The options factory could probably be modified to create an option set for itself if an appropriate one doesn't already exist using a similar method in the optionsets factory—it probably should anyway. And I was thinking you could pass 'active' as the argument to the `get_loan_status` method when that's the one you need. Does that make sense?

elliottjf:
Not sure why this is still reporting a branch conflict.  I've merged in the latest from 'develop', and I believe it's ready for a final review.

meltheadorable:
It's reporting a conflict because it appears that it merged in an outdated version of `develop` again. Are you 100% sure that you pulled the newer version of develop from the remote before you performed the merge? If so, there must be something weird going on with your git configuration/tooling that's somehow causing you to keep merging old versions of branches.

elliottjf:
Okay.  Looks like I've gotten everything properly merged now.  (on both branches)

meltheadorable:
Looks great! Can we move the "Transient" Option & OptionSets to Legacy? Other than that, I think you're okay to merge.

PR #22, opened 2016-01-12
3742 login form

emitche:
Adds login page

- Adds forgot password page (frontend only)
- Includes work to refine how alerts are displayed across the application
- Ensures login and logout work in app


app/assets/stylesheets/admin/devise.css.scss
```diff
@@ -0,0 +1,73 @@
+@import "bootstrap_signin";
```
emitche: Instead of import, can use a direct link to stylesheet if that would be better.

app/helpers/devise_helper.rb
```diff
@@ -0,0 +1,31 @@
...
+      HTML
+
+      html.html_safe
+    end
+  end
+
+  def devise_error_messages_content?
+    if devise_error_messages! && !devise_error_messages!.empty?
+      return true
+    end
```
emitche: There has to be a better way to say `!devise_error_messages!.empty?`, but I had some errors trying to do it different ways.

app/views/devise/confirmations/new.html.erb
```diff
@@ -0,0 +1,16 @@
+<h2>Resend confirmation instructions</h2>
```
emitche: Most of the templates are generated templates from devise.

app/views/devise/passwords/new.html.erb
```diff
@@ -0,0 +1,23 @@
+<% content_for :title do "Forgot Password" end %>
```
emitche: This template has some content tweaks to make it work well with our application/ devise layout.

app/views/devise/sessions/new.html.erb
```diff
@@ -0,0 +1,38 @@
+<% content_for :title do "Login" end %>
```
emitche: This is our login form and has been tweaked/changed from the generated devise view.

app/views/devise/shared/_links.html.erb
```diff
@@ -0,0 +1,25 @@
+<%- if controller_name != 'sessions' %>
+  <%= link_to "Log in", new_session_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.registerable? && controller_name != 'registrations' %>
+  <%#= link_to "Sign up", new_registration_path(resource_name) %><br />
```
emitche: What are our plans related to signing in and users?

This view will eventually need editing once that is decided. There are also currently no translations set in these devise templates yet.

app/views/layouts/base.html.slim
```diff
@@ -14,5 +14,5 @@ html lang="en-US"
     = yield :styles
 
   body
-    = render "application/alerts"
+    = content_for?(:alerts) ? render(inline: '') : render('application/alerts')
```
emitche: I want this to render nothing, but was having trouble getting that working. That's why I have `render:(inline: '')`.

meltheadorable: Does `= render('application/alerts') if content_for?(:alerts)` work? I'm not entirely sure if slim will like it, I don't have much experience with it. If not

```ruby
- if content_for?(:alerts)
  = render('application/alerts')
```
Definitely should

emitche: It does work, but your block looks cleaner and easier to read.

I'm actually trying to get it to render nothing if alerts are defined. If alerts are previously defined, they've already loaded on another part of the page.

meltheadorable: Switching the `if` to `unless` would make it do the opposite thing, in this case, render the alerts unless they are already defined.

app/views/devise/shared/_links.html.erb
meltheadorable: I think we decided no registrations, so this little block can be removed.

emitche:
I need to adjust controllers to use authentication. Will push up an update when that is in.

meltheadorable:
@emitche re: updating the controllers, you should just need one line at the top of `app/controllers/admin/admin_controller.rb` that does `before_action :authenticate_user!`, assuming the other admin controllers are inheriting from the admin controller

emitche:
This branch is ready to be reviewed. Thanks!

app/views/devise/passwords/new.html.erb
```diff
@@ -0,0 +1,22 @@
...
+
+<% content_for :alerts do render "application/alerts" end %>
```
linesarefuzzy: For a one-liner, you can do `content_for :alerts do render "application/alerts"`.

app/views/devise/passwords/new.html.erb
```diff
@@ -0,0 +1,22 @@
...
+
+<%= form_for(resource, as: resource_name, url: password_path(resource_name), html: { method: :post, class: "form-signin" }) do |f| %>
+  <h2 class="form-signin-heading">Forgot your password?</h2>
+  <p class="centered-text">Send me reset password instructions</p>
+
+  <%= yield :alerts %>
```
linesarefuzzy: Didn't we discover that yielding the same symbol as the parent caused problems?

linesarefuzzy:
I'm getting an ArgumentError when I click Forgot Password: `Missing host to link to! Please provide the :host parameter, set default_url_options[:host], or set :only_path to true`. Does that happen for you?

linesarefuzzy:
Otherwise, looks great! Thanks!

app/views/devise/passwords/new.html.erb
emitche: :alerts is only used in devise layout and application layout. The base does not use it.

emitche:
For forgot password and email notifications, those have not been set up yet. I don't see an issue in our issue log for that. But I think someone else could handle getting that set up in the system, as someone else on team likely has experience with that part.

linesarefuzzy:
In that case, let's comment out the link for now.

linesarefuzzy:
(and don't close the login issue on redmine yet)

app/views/devise/passwords/new.html.erb
emitche: Smart. Thanks.

emitche:
Okay. I'll make the template change and keep issue open.

app/views/devise/passwords/new.html.erb
linesarefuzzy: But you have a `content_for :alerts` a few lines above in this file. Is it only when it's nested that it causes problems?

emitche: The parent never uses :alerts other than to determine if it is defined.

Alerts is defined on line 3 and is rendered in the devise template on line 9 using the yield. Inside the base template, the base template sees that `:alerts` has been defined, so it does not load `render "application\alerts"`.

I'd like to display the alerts in different spots depending on the part of the app. Alerts are displayed inside the login/forgot password box when using the devise layout. They are displayed right about the content when using the application layout.

The base layout displays alerts, but not through the `:alerts` symbol and only if `:alerts` has not been defined anywhere else.

Sorry if this is confusing. Please let me know if you have additional questions and if we need to discuss this. Thanks.

PR #25, opened 2016-01-13
3738 loan question -> custom field sets modeling & migration

elliottjf:
this also includes a proposed change to generically include the seeds.rb data as part of the DatabaseCleaner configuration


spec/rails_helper.rb
```diff
@@ -46,6 +46,9 @@
 
   config.before(:suite) do
     DatabaseCleaner.clean_with(:truncation)
+    # Seems cleaner to just put this here than track exactly which tests need the seed data.
+    # The Languages dependency at least is pretty fundamental
+    seed_data
```
meltheadorable: As written, this won't work the way you intend. `before(:suite)` only runs once, before the entire test suite, which will then be truncated for every test. I'm still really not comfortable with this though, using `seed_data` should be very rare and it's *very* useful to know which tests actually need the seed data, most shouldn't. 

In my `loans_index` branch, for instance, the call to `seed_data` for the `loans` model spec has been removed, and it also won't break anything to remove it from the `person` and `organization` model specs. It's important that we be able to keep track of where we have implicit dependencies on a specific database state, relying on it makes things brittle and seeding too much data, besides slowing down the test suite, can mask issues with the code.

meltheadorable:
Are these the same custom fields we had discussed making a many-to-many relationship with loans on Skype or are these something else?

spec/rails_helper.rb
elliottjf: It seems to be working as I expected.  And the full set of tests were running fine in my branch.  The way that I understood the 
  DatabaseCleaner.strategy = :transaction
setting is that it's not actually truncating before test, but relying on the transaction rollback.  
(I was quite puzzled at first when I was trying to debug the unit test behavior because I wasn't able to see any of the intermediate database state while pausing execution in the middle of the test.  And that is how I realized there was a transactional approach being used before I noticed the DatabaseCleaner configuration)

I personally think that the seed data should be accepted as legitimate system configuration state which can be relied up to exist within the unit tests.  It feels an unnecessary effort and added complexity to try to write all of the unit tests without its existence.

@linesarefuzzy  Do you have a perspective here?

elliottjf:
These are the custom fields needed to support the Loan Questionnaires, and I anticipate them being useful also when handling division specific custom fields to support data like 'is_recovered' on Organization.  
Once these building blocks are in place, it may provide a simpler solution for handling the project log 'achievements' use case that was raised, but it's not yet clear to me if that is really the case or not.

spec/rails_helper.rb
linesarefuzzy: See my comment on the other PR.

app/models/custom_field.rb
```diff
@@ -0,0 +1,45 @@
...
+#
+#  fk_rails_b30226ad05  (custom_field_set_id => custom_field_sets.id)
+#
+
+class CustomField < ActiveRecord::Base
+  include Translatable
+
+  belongs_to :custom_field_set
+  # note, the custom field form layout can be hierarchially nested
+  belongs_to :parent, class_name: 'CustomField'
```
meltheadorable: this looks like a potentially good use case for `closure_tree` over handling this hierarchy manually.

elliottjf: Yeah, that occurred to me also.  The exact requirements around the questionnaire feature isn't completely clear to me yet though, so I didn't want to bite off any more depth into the modelling work here until that was more clear.

elliottjf:
Any thing still pending here before it's ready to merge down to develop?

PR #29, opened 2016-01-21
3749 table filters

emitche:
- Add loans index table (#3983)
- Add external table filters to coops index and loans index
- Add styling to table filters
- Update wice grid to developer recommended version


app/assets/stylesheets/admin/content.css.scss
```diff
@@ -0,0 +1,11 @@
+.content {
```
hooverlunch: I think support for .css.scss extensions is going away. I would recommend moving everything to just .scss.

app/controllers/admin/loans_controller.rb
```diff
@@ -0,0 +1,8 @@
+class Admin::LoansController < Admin::AdminController
+  def index
+    @loans_grid = initialize_grid(
+      Loan,
+      include: [:division, :organization]
+    )
```
hooverlunch: I think this can be one line. Won't be too long.

app/views/admin/loans/index.html.slim
```diff
@@ -0,0 +1,74 @@
+- content_for :title do
+  - 'Loans'
```
hooverlunch: Should be possible to do single line blocks in slim?

Gemfile
```diff
@@ -48,7 +48,8 @@ gem 'mini_magick'
 gem 'mysql2'
 
 # Tables
-gem 'wice_grid'
+gem 'wice_grid', '3.6.0.pre4'
```
hooverlunch: Just curious, what was this specific version needed for?

app/views/admin/loans/index.html.slim
```diff
@@ -0,0 +1,74 @@
+- content_for :title do
```
hooverlunch: When I load this page I'm getting several missing assets. e.g. /images/icons/grid/expand.gif. There are three in total. Can you duplicate?

app/views/admin/loans/index.html.slim
```diff
@@ -0,0 +1,74 @@
...
+    - @empty_list = true
+    - content_tag(:div, "No records found", class: 'alert alert-info')
+
+  - g.column name: 'ID', attribute: 'id'
+
+  - g.column name: 'Division', attribute: 'name', assoc: :division,
+    custom_filter: :auto, detach_with_id: :division_filter
+
+  - g.column name: 'Co-op', attribute: 'name', assoc: :organization,
+    custom_filter: :auto, detach_with_id: :coop_filter
```
hooverlunch: For some reason I am getting a bunch of coops in the coop filter for which there are no loans defined. @linesarefuzzy any thoughts on this?

config/initializers/wice_grid_config.rb
```diff
@@ -115,40 +112,40 @@
...
-  Wice::Defaults::HELPER_STYLE = :calendar
+  # Default column filters
+  # Possible values:
+  # * <tt>:jquery_datepicker</tt> - Jquery datepicker (works for datetime, too)
+  # * <tt>:bootstrap_datepicker</tt> - Bootstrap datepicker (works for datetime, too)
+  # * <tt>:rails_date_helper</tt> - standard Rails date helper
+  # * <tt>:rails_datetime_helper</tt> - standard Rails datetime helper
+
+  Wice::Defaults::DEFAULT_FILTER_FOR_DATE     = :jquery_datepicker
+  Wice::Defaults::DEFAULT_FILTER_FOR_DATETIME = :jquery_datepicker
```
hooverlunch: Can we use the bootstrap one? The jquery one looks dated in my opinion.

hooverlunch:
I am curious why a second filter area appears only when a filter value is entered in the detached filter panel:

image

hooverlunch:
I believe the plus signs in the panel should be to the right of the selects, not below.

image

hooverlunch:
The description filter does not seem to be working. A search for 'Cummerata' yields two projects without that word in the description.

image

Then if I search for 'Flatley', which is in one of the descriptions, I get nothing:

image

hooverlunch:
The loan table and filters do not include 'loan agent' which appears on the wireframe https://moqups.com/linesarefuzzy/WmMl8DLm/p:a470ee64d

Should this be there?

hooverlunch:
The wireframe also suggests you can sort by signing date, but I don't see that here.

hooverlunch:
The created and update times probably need to be formatted a little more nicely:
image

Fuzzy had some rather "fuzzy" examples in the wireframe:
image

Perhaps he can comment on the intention here.

hooverlunch:
Signing date format also is perhaps just the Rails default. Are there standard date/time/datetime formats for the project?

hooverlunch:
Are these values being translated:

image

I assume from the lowercase letters that they are not. They probably should be.

hooverlunch:
Overall this is looking really slick.

hooverlunch:
Could we perhaps abbreviate "months" to "mo", then this would fit on one line.
image

app/views/admin/loans/index.html.slim
```diff
@@ -0,0 +1,74 @@
...
+
+  - g.column name: 'Status', attribute: 'status_value',
+    custom_filter: :auto, detach_with_id: :status_filter
+
+  - g.column name: 'Signing Date', attribute: 'signing_date'
+
+  - g.column name: 'Length', attribute: 'length_months' do |loan|
+    - loan.length_months.to_s + " months"
```
hooverlunch: This should be translated. As everywhere else throughout. Just a reminder.

app/views/admin/loans/index.html.slim
```diff
@@ -0,0 +1,74 @@
...
+
+  - g.column name: 'Description', attribute: 'name',
+    detach_with_id: :description_filter
+
+  - g.column name: 'Amount', attribute: 'amount' do |loan|
+    - number_to_currency(loan.amount)
+
+  - g.column name: 'Created', attribute: 'created_at'
+
+  - g.column name: 'Updated', attribute: 'updated_at'
```
hooverlunch: Very nice and clean.

app/views/admin/loans/index.html.slim
```diff
@@ -0,0 +1,74 @@
...
+      label.col-lg-1 Description
+      span.col-lg-7
+        = grid_filter(@loans_grid, :description_filter)
+
+    div.row.actions
+      div.col-lg-9
+      button.btn.btn-link.col-lg-1.wg-external-reset-button data-grid-name="grid" Clear
+      button.btn.btn-primary.col-lg-2.wg-external-submit-button data-grid-name="grid"
+        | Filter
+  br
```
hooverlunch: I think it's best to avoid using `<br/>` as a spacing mechanism. See http://stackoverflow.com/questions/1726073/is-it-sometimes-bad-to-use-br

app/views/admin/organizations/index.html.slim
```diff
@@ -3,19 +3,48 @@
...
-  - g.column name: 'City', attribute: 'city'
+  - g.column name: 'City', attribute: 'city', detach_with_id: :city_filter,
+    custom_filter: :auto
 
-  - g.column name: 'Country' do |organization|
+  - g.column name: 'Country', attribute: 'name', assoc: :country,
+    custom_filter: :auto, detach_with_id: :country_filter do |organization|
     - organization.country.try(:name)
 
   - g.column name: 'Created', attribute: 'created_at'
```
hooverlunch: Same comment here about date formats.

app/views/admin/organizations/index.html.slim
```diff
@@ -3,19 +3,48 @@
...
+      span.col-lg-3
+        = grid_filter(@organizations_grid, :name_filter)
+
+      label.col-lg-1 City
+      div.col-lg-3
+        = grid_filter(@organizations_grid, :city_filter)
+
+      label.col-lg-1 Country
+      div.col-lg-3
+        = grid_filter(@organizations_grid, :country_filter)
```
hooverlunch: The spacing on these filters doesn't look quite right:

image

e.g City is closer to the name select than to the city one...

I wonder if it would be easier to put the label and the filter in the same column instead of separate ones.

app/views/admin/organizations/index.html.slim
```diff
@@ -3,19 +3,48 @@
...
+    custom_filter: :auto, detach_with_id: :country_filter do |organization|
     - organization.country.try(:name)
 
   - g.column name: 'Created', attribute: 'created_at'
 
   - g.column name: 'Updated', attribute: 'updated_at'
+
+- unless @empty_list
+  header.block
+    div.row.filters
```
hooverlunch: I am noticing a lot of your `row` divs are not inside a `container`. See http://getbootstrap.com/css/#grid-intro. That is why, I think, you are getting the padding to the left of 'Name' here:

image

app/views/layouts/admin/_tables.html.slim
```diff
@@ -1,3 +1,4 @@
 - content_for :styles
   = stylesheet_link_tag "jquery-ui", :media => "all"
   = stylesheet_link_tag "vendor/wice_grid", :media => "all"
+  = stylesheet_link_tag "admin/table/table", :media => "all"
```
hooverlunch: I don't think you should need to include this explicitly. The Rails asset pipeline should pick it up.

config/initializers/wice_grid_config.rb
```diff
@@ -115,40 +112,40 @@
...
 
   # Format of the datetime displayed.
   # If you change the format, make sure to check if +DATETIME_PARSER+ can still parse this string.
-  Wice::Defaults::DATETIME_FORMAT = "%Y-%m-%d %H:%M"
+  Wice::Defaults::DATETIME_FORMAT = '%Y-%m-%d %H:%M'
 
   # Format of the date displayed.
   # If you change the format, make sure to check if +DATE_PARSER+ can still parse this string.
-  Wice::Defaults::DATE_FORMAT     =  "%Y-%m-%d"
+  Wice::Defaults::DATE_FORMAT     =  '%Y-%m-%d'
```
hooverlunch: Our style guide actually says to prefer double quotes I think.

config/initializers/wice_grid_config.rb
```diff
@@ -115,40 +112,40 @@
...
-  Wice::Defaults::DATE_FORMAT_BOOTSTRAP     =  "yyyy-mm-dd"
-
+  Wice::Defaults::DATE_FORMAT_BOOTSTRAP     =  'yyyy-mm-dd'
 
   # With Calendar helpers enabled the parameter sent is the string displayed. This lambda will be given a date string in the
   # format defined by +DATETIME_FORMAT+ and must generate a DateTime object.
   # In many cases <tt>Time.zone.parse</tt> is enough, for instance,  <tt>%Y-%m-%d</tt>. If you change the format, make sure to check this code
   # and modify it if needed.
-  Wice::Defaults::DATETIME_PARSER = lambda{|datetime_string|
+  Wice::Defaults::DATETIME_PARSER = lambda do|datetime_string|
```
hooverlunch: Nice fix. I think you want a space after the `do` though.

config/initializers/wice_grid_config.rb
```diff
@@ -156,19 +153,18 @@
...
   #  $( ".hasDatepicker" ).datepicker( "option", "yearRange", "2000:2042" );
   Wice::Defaults::DATEPICKER_YEAR_RANGE = (from = Date.current.year - 10).to_s + ':' + (from + 15).to_s
 
-
   # With Calendar helpers enabled the parameter sent is the string displayed. This lambda will be given a date string in the
   # format defined by +DATETIME+ and must generate a Date object.
   # In many cases <tt>Date.parse</tt> is enough, for instance,  <tt>%Y-%m-%d</tt>. If you change the format, make sure to check this code
   # and modify it if needed.
-  Wice::Defaults::DATE_PARSER = lambda{|date_string|
+  Wice::Defaults::DATE_PARSER = lambda do|date_string|
```
hooverlunch: Same here.

hooverlunch:
This is quite an intricate feature and it is looking great. I've requested a ton of smaller changes and refinements, but in general this is looking great.

app/assets/stylesheets/admin/content.css.scss
emitche: Yay! I would much prefer it. :)

Gemfile
emitche: I wish the developer moved this version over to 3.6.1 or something. This specific version is needed because the master branch has an older way of using associations in the table, specifically related to filters and joined tables. See https://github.com/leikind/wice_grid/wiki/What's-New-In-3.6.0#new-api-for-joined-tables.

I started using the older API and master branch, but ran into some problems with items from associations and not being able to use those contents for filtering. So instead I replaced the gem with the newest version recommended by the developer and it filtering works with items from a joined table.

config/initializers/wice_grid_config.rb
emitche: Agreed. Looks polished.

config/initializers/wice_grid_config.rb
emitche: This file is generated by wice grid.

hooverlunch: Hmm ok. It is meant for customization though so we can still fix the syntax.

app/views/layouts/application.html.slim
```diff
@@ -1,6 +1,7 @@
 - content_for :styles
   = stylesheet_link_tag "admin/bootstrap_layout", :media => "all"
   = stylesheet_link_tag "admin/menu", :media => "all"
+  = stylesheet_link_tag "admin/content", :media => "all"
```
emitche: Move to _page.slim

config/initializers/wice_grid_config.rb
emitche: I changed that.

emitche:
Looks like we definitely need to set some default time and date for the application. But it will need to be localized. Any recommendations for how to do that?

app/views/admin/loans/index.html.slim
emitche: Fixed.

emitche:
hoovrlunch: "I am curious why a second filter area appears only when a filter value is entered in the detached filter panel:"

The eye displayed on the rightmost part of the table is the show/hide filters control. When you add a filter, it then displays all the filters. That is the default wice grid functionality when using internal and external filters.

Is there a way you would want it to function ideally?

app/views/admin/loans/index.html.slim
emitche: Done.

config/initializers/wice_grid_config.rb
emitche: Made updates.

emitche:
@hooverlunch, I've removed the loans table from this feature. There is a lot more work that needs to be done on loans. So this branch is just changes to filters in the coop table/index. Let me know what you think and if additional changes are needed. Thanks.

config/routes.rb
```diff
@@ -9,6 +9,7 @@
   namespace :admin do
     resources :calendar, only: [:index]
     resources :dashboard, only: [:index]
+    resources :loans, only: [:index]
```
emitche: Simple loans index for now. A different branch addresses the loans index.

emitche:
Just noticed changes to the datetime have made datetime filters not work properly. Jan 5 2016 turns into May 1 2016. I need to fix that.

config/initializers/wice_grid_config.rb
emitche: Unfortunately, Bootstrap breaks wice grid. There is an error related to range when bootstrap is used. I replaced it back with jQuery.

Gemfile
hooverlunch: Cool, nice research!

hooverlunch: We should keep an eye on this gem and when 3.6.0 is released, we should update the Gemfile.

app/assets/stylesheets/admin/content.scss
```diff
@@ -0,0 +1,11 @@
+.content {
```
hooverlunch: Thanks for renaming all these! Hooray for simplicity!

hooverlunch: You may as well rename the ones in the /embedded dir too?

app/views/admin/loans/index.html.slim
```diff
@@ -0,0 +1,3 @@
+- content_for(:title, content='Loans')
```
hooverlunch: I don't think you need the `content=` bit here. I could be wrong though. Maybe it's used further down?

app/views/admin/organizations/index.html.slim
```diff
@@ -1,21 +1,56 @@
-- content_for :title do
-  - 'Co-ops'
+- content_for(:title, content = t("general.coops"))
```
hooverlunch: Same here.

app/views/admin/organizations/index.html.slim
```diff
@@ -1,21 +1,56 @@
...
+        = grid_filter(@organizations_grid, :country_filter)
+
+      div.actions
+        button.btn.btn-link.wg-external-reset-button data-grid-name="grid"
+          = t("general.clear")
+
+        button.btn.btn-primary.wg-external-submit-button.action data-grid-name="grid"
+          = t("general.filter")
+
+= render_grid(@organizations_grid)
```
hooverlunch: Nice work on translations.

config/locales/en.yml
```diff
@@ -1,4 +1,8 @@
 en:
+  time:
+    formats:
+      default: '%b %d, %Y %H:%M'
```
hooverlunch: Nice! I would suggest using 12-hour format instead of 24-hour though. I personally prefer 24-hour but it makes some people flip out!

hooverlunch:
Looking much better! A couple smaller things:

* It looks like the filter row at the top of the table still gets shown when you do a filter from the filter panel. It's not a great user experience since if someone doesn't know what it's for (and it's not obvious at first) then it's a little disorienting. I see from the docs that it's because it shows the filter row if any of the filters are active. Hard to change that without hacking wice grid I think. So my suggestion would be to show it all the time and perhaps let's make it a bit nicer to look at. The default styling and UI for the filter bar is not super intuitive. Can we perhaps do the following:
** The ID box doesn't need to be so wide. It makes the column much wider when the filter row is shown.
image
** The cells should be vertical-align middle or top. I think the bottom alignment looks strange.
** Perhaps a different background color for the filter bar, perhaps the same color as the upper filter panel? This might help people make the connection.
** Perhaps also make the text (and icon) color in the filter row a little fainter so it has less weight than the main table text.
** The two calendar icons on top of each other look so strange but I'm not sure what to do with them. If it were easy to just have a 'Filter...' link that shows them when you click it, I would suggest that, but it's probably not. So I think they are fine, especially given it's an admin panel.
* Great job restyling the upper filter panel!
* Just noticed this: we should be careful with that 'show all' link, especially on other tables. If this system scales to say 100,000 loans, and someone clicks that, boom, down goes the site. Does wice support offering different page sizes (e.g. 20, 50, 100) instead?

Sorry I didn't notice some of this early. There was a lot going on!

You can go ahead and merge this after dealing with the above.

Great work!

emitche:
Thanks for the comments. I will take a look and fix today and hopefully can get this merged in today.

hooverlunch:
Fantastic!

On Wed, Feb 3, 2016 at 12:16 PM, Elizabeth Mitchell <
notifications@github.com> wrote:

> Thanks for the comments. I will take a look and fix today and hopefully
> can get this merged in today.
>
> —
> <https://github.com/sassafrastech/madeline_system/pull/29#issuecomment-179355516>
> .
>



-- 
Tom Smyth

touchstonecohousing.org

emitche:
For the filters, it is possible to move all filters that are needed to external filters. This removes the table column with the eye (hide/show filters) and removes the row with the filters. I've done this, as it looks better.

The calendar filters are not ideal. I prefer Bootstrap date helper, but it seems there is a bug with that. When I have some time, I will report to the wice grid developer, as it is not working as intended.

app/views/admin/loans/index.html.slim
emitche: This was to replace `- content_for :title do t("general.coops") end`. Ruby recommended syntax `- content_for :title { t("general.coops") }` does not work here. Do you have a prefered way to display this text. My preference is `do` `end`.

emitche: Awesome. Just realized `- content_for(:title, t("general.coops"))` works just fine. Thanks.

app/assets/stylesheets/admin/content.scss
emitche: I will do on another branch, just in case.

emitche:
Yes, we can adjust number of records shown per page. I've moved it up to 50 for now.

We also can show pagination at the top of the table, if needed. That's another option that can be changed.

Gemfile
emitche: That is sensible.

app/views/admin/organizations/index.html.slim
```diff
@@ -1,21 +1,73 @@
...
+        label
+          = "Created"
+        div.filter-content.dates
+          = grid_filter(@organizations_grid, :created_filter)
+
+      div.filter
+        label
+          = "Updated"
+        div.filter-content.dates
+          = grid_filter(@organizations_grid, :updated_filter)
```
emitche: I'm not thrilled with how these look right now.

PR #30, opened 2016-01-22
3762 custom data handling

elliottjf:
handles loan response data fields and custom field data on organization (or other core models)
allows for field sets to inherited or overridden per division


elliottjf:
Do we need to upgrade the postgres version being used by Travis?
Here's the error breaking the tests:
  ActiveRecord::StatementInvalid: PG::UndefinedObject: ERROR:  type "json" does not exist

meltheadorable:
We might need to upgrade it, if so you should just need to add

```yaml
addons:
  postgresql: "9.4"
```

To `.travis.yml`

elliottjf:
Cool.  That fixed the Travis errors!

app/models/legacy/transient_option.rb
```diff
@@ -1,10 +1,14 @@
-class TransientOption
+module Legacy
```
meltheadorable: Good thinking moving these into the legacy namespace!

app/models/division.rb
```diff
@@ -50,8 +50,7 @@ def self.root_internal_name
 
   def self.root
     # create on demand if not present for convenience of blank db's and test cases
-    # @@root ||=
```
meltheadorable: I've actually been meaning to mention this, but [closure_tree already supports getting the root](https://github.com/mceachen/closure_tree#class-methods), and I'm not sure whether I understand the impetus behind switching from `find_or_create_by` to `find_by || create`. If the only issue is that you want to create it with more attributes than you want to find it by, you can just use `Division.find_or_create_by(internal_name: 'root').update_attributes(name: 'Root Division')`

spec/factories/divisions.rb
```diff
@@ -22,6 +22,9 @@
   factory :division do
     description { Faker::Lorem.sentence }
     name { Faker::Company.name }
+    internal_name nil
+    parent nil
```
meltheadorable: These don't need to be explicitly defined because they'll be `nil` if you leave them out.

spec/models/organization_spec.rb
```diff
@@ -8,4 +8,43 @@
 
   it_should_behave_like 'notable'
 
+  it 'should set and get custom values' do
```
meltheadorable: Part of the reason we define multi-model concern specs using shared example groups is so that we don't need to duplicate all this work in every model. If the worry is testing that specific features work, so that you can make sure the "plumbing" is all good for utilizing these for actual use cases, that'll be better handled in feature/integration specs once there's a UI for a user utilizing these features. I would factor out the parts of this that are mostly about whether the concerns/modules are working properly into a shared example group.

app/models/loan.rb
```diff
@@ -64,6 +70,8 @@ class Loan < ActiveRecord::Base
   attr_option_settable :status, :loan_type, :project_type, :public_level
 
 
+  attr_custom_model_linkable LOAN_CRITERIA_FIELD_SET_SYMBOL, POST_ANALYSIS_FIELD_SET_SYMBOL
```
meltheadorable: I'm not sure if I understand the reasoning behind not just using `attr_custom_model_linkable :loan_criteria, :loan_post_analysis` here.

app/models/division.rb
elliottjf: It looks like the closure_tree method will potentially either return an arbitrary root node.  I wanted a well defined root node.  

Regarding find_or_create_by() followed by an update_attribute(), won't that always perform the update operation?  I'd rather avoid the update when it already exists, which should be almost always.

spec/factories/divisions.rb
elliottjf: It felt useful to me to explicitly assign the nil, in particular for the 'parent' field, so it's clear that the field wasn't inadvertently left out.  I also noticed an explicit 'nil' in your basic_projects.rb factory, so I assumed that was an acceptable convention.

app/models/division.rb
meltheadorable: Well, the larger structural question is whether it makes sense to create the root node in the model like that. It's meaningless-to-undesirable to do in production, and only useful for the tests. This is the sort of thing I'd prefer to see handled by other means entirely.

While it's true that `closure_tree` will return an arbitrary root node, if we're using the root as a "global parent" then there will only be one (and we could potentially write a validation or DB constraint that enforces that). That's something I'd rather do than re-define the method (which looks like a `getter`) to potentially unexpectedly _create_ a record. If that's only in there for the consideration of the tests, I think we should re-think its placement in the model.

spec/factories/divisions.rb
meltheadorable: `basic_projects` was just copied over from the old system, I can't recall whether defining the nil was necessary there explicitly, for legacy reasons, but since we eventually made the decision not to carry it over to this project at all, that factory will be getting deleted entirely in the slightly-more-long-term. 

Personally, I believe that between the schema definition at the top of the file and the knowledge of whether or not your tests pass without the attribute present, it's not really necessary to further annotate which fields are being excluded by explicitly assigning nil, but it's a relatively minor point.

spec/models/organization_spec.rb
elliottjf: I considered creating a shared example group, but it seemed that there were too many context parameters which would need to be passed in and the abstractness would result in test code more difficult to follow.  In this case a set of tests specific to Organization felt more useful.  

Creating an additional shared example group could still be done, although I'm trying to balance the effort invested into more rigorous unit tests versus the pressure to keep within the project budget.

app/models/loan.rb
elliottjf: Yeah. I wanted to get feedback about best practice / style here.  The defined constants are being referenced by the unit test code, and I just realized should have also been referenced by the static_data.rb code.  There will also be UI code which needs to be specifically tied to those field set names.  I liked the idea guaranteeing that all those usages share the same symbol.  But I recognize this is a java pattern, and I was wondering if the "ruby way" is to just accept that changes to symbols can result in breakages and to rely on the unit tests to catch.

Do you prefer for me to remove the constants and just use direct symbol references?

spec/models/organization_spec.rb
meltheadorable: Good specs save us a lot of time in the long run, not just on long-term maintenance but it also gives us an early warning during development before we introduce regressions that end up going unnoticed and becoming harder to correct later once more has changed around them. 

If `CustomValueSettable` is going to be used more than just in organizations, then it should be a shared example group. Same goes for `CustomModelLinkable`. These concerns usually only have one public method that you interface with, which takes a pretty simple argument. The shared example groups shouldn't really require more arguments than that, but you will need to write the tests in a more generic way. Things like "should be able to set custom values" and "can filter by custom values" and "should inherit custom field sets" are presumably all things that should work for *every* model that includes `CustomValueSettable` not just organizations, so having a shared example group saves us time re-implementing all of the same specs for those models.

I'm definitely sensitive to the project budget concerns, and if you don't think factoring these out as a shared group would be relatively quick & straightforward I'm not going to force the issue, but we should flag these with a `TODO:` in that case and try to get to it when we can, I think it's pretty important that we can test our concerns in a generalizable way.

app/models/division.rb
elliottjf: Agreed that it's irrelevant in production (and documented as such), and conceptually it shouldn't be needed.  My personal criteria for how to approach a need is weighted more toward how to isolate the concern, avoid repetition, and simplest code, and I find it pragmatic at times to compromise some ideological principles.  

If you feel strongly that the create logic doesn't belong here, then I'll replace with a fatal exception and add the necessary code at the test spec level to guarantee the root node exists.

app/models/loan.rb
meltheadorable: The way I see it, these are essentially part of our schema, which we have to operate under the assumption will be, relatively speaking, lexically stable over most of the life of the software. The test suite should catch regressions in the unlikely event that we need to rename these, and since the constants contain the names of the symbols themselves, we're realistically going to need to rename those too if we don't want them to have misleading names for the rest of forever, which makes the addition of an extra constant into the process sort of redundant. `LOAN_CRITERIA_FIELD_SET_SYMBOL = :something_else_entirely` might save us a `find --> replace` in our editors, but I'm not sure of the benefit otherwise.

elliottjf: Fair enough.  In addition to being a convention from the java world, most of my work has been in environments where time pressures precluded the ability to maintain rigorous test cases, so I naturally look for other ways to guarantee system integrity.  I'll remove the constants.

app/models/division.rb
meltheadorable: I wouldn't replace with an exception, if we move the create logic out of the model then we can lean on the `root` method already provided by `closure_tree`, since with a single root it will have equivalent behavior. If a single root is a constraint of our system, then we should define that on a validation and make it explicit. This is a good case for inclusion in `seeds.rb` in production, and for the test suite, [the advice in this discussion from a few PRs back is sound](https://github.com/sassafrastech/madeline_system/pull/20#discussion_r49797743)

elliottjf: Okay.  I'll replace the find by internal name with the closure_tree root method, add validation logic and have faith that the root will be unique.

I still want a way to fail-fast with a useful error message if for any reason the resolved root node is missing.  Essentially an "assert".  Do you have a convention / style around "assert" style checks?

Also, the root node is currently being created in seeds.rb.  I've been meaning to ask you if you still want to keep seeds.rb and if so, how much should be in there.  For my own development, I've been using lib/legacy/static_data.rb which includes the more through set of data relevant to the migrated system.  So I'm happy to be brutal in stripping down seeds.rb to the bare minimum that you think is appropriate.

app/models/loan.rb
meltheadorable: I can understand that, I actually wasn't allowed to write tests at all at my last job, our deadlines were sometimes measured in days or weeks rather than on a scale of a month or more, but it's a way of thinking I came to believe was very very important doing my own work on my own side projects and a skill set that I've tried very hard to build up in spite of that. 

My concerns re: testing aren't purely ideological, but based on a foundation of seeing these practices result in much higher-quality and more maintainable code, and reinforced by the instability of the software produced under those environments with strong time pressures or institutional constraints that prevented good regression testing.

The ruby world has *much* better support for meaningful tests. When we start having enough UI to do them, most of the specs should end up being "feature" specs, which are integration tests that actually run a headless renderer and navigate the platform the same way an end-user does. The ability to quickly and reliably write tests that work the entire system in roughly the same way a user does makes them invaluable.

The scope of "unit tests" in rails work tends to be a little more narrow, which is part of the reason that I like to use shared example groups for functionality used in multiple models where their interface in the calling model is pretty simplistic. Mostly we just want to know "do the custom methods manually defined on this model behave the way they are supposed to?", we don't test basic rails functionality like scopes, saving records, or even validations for the most part. Finding where to make those tradeoffs is more art than science, and I understand the frustration (I was pulling my hair out the first, oh, 4 or 5 projects where I really tried to do rigorous testing like this), but I promise that I'm not asking you to make these sorts of changes for no reason :sweat_smile: .

app/models/division.rb
meltheadorable: I'm not sure *where* in the codebase it makes the most sense to do this yet, but we could `raise` an error in the event a root division cannot be found. Probably, if it's fundamental, a check like this will make it somewhere into `application_controller.rb`, for some things this is a potentially larger issue than others, so I'd like to defer turning this into a global failure condition until we've sorted through whether a single overarching global root division is strictly necessary, which is something I anticipate we'll discover when implementing the authorization system.

meltheadorable: Re: the `seeds.rb` question, we definitely want to keep the file. If it were me, the only things I would put in there are the root division, and *maybe* the countries, languages, and currencies (but without hardcoding their `id:` attributes, if at all possible.)

I think everything else is/should be more or less optional, if the system won't crash because they're gone, it probably doesn't need to be there, but there are occasional exceptions and reasons to compromise on that point.

elliottjf: The places in the code which currently reference Division.root rely on it existing.  So it feels more appropriate that the validation of existence is at the model level, not the controller level.  If the need for a root division disappears as the project evolves, then that will certainly take care of things, but it's currently depended upon, and in my mind conceptually part of the current Division contract.

If you really believe that an assert doesn't belong in the model class, I'll remove it, but it'll just result in less useful errors when debugging test code in situations where the data wasn't properly initialized.

elliottjf: Re seeds.rb.  I'll strip it down to just a generic root division without an assigned ID and nothing else.  The ID's were only important in the context of the migrated system, so that need is now handled by the static_data.rb code.  I'll let you add countries, languages or currencies if/when you see fit.

linesarefuzzy: I thought we talked way back about the root division and decided that we needed one. We need to be able to run reports on all divisions in the system. Having a root division seems like the obvious way to do that within the permission system we've designed. This root division stuff is turning out to be pretty complicated, so maybe it would have been simpler to add a super admin flag to some users. But if the code is already written that way, and refactoring would be complicated, it doesn't seem worthwhile to refactor it now. We have a budget here and we have used about 3/4 of the contingency already. 

That said, I don't understand why the code is relying so heavily on there being a root division. I thought the idea was that we would just make one, and then its members would be effective super admins. And I'd love to avoid specifying primary keys in the code. It doesn't seem like that should be necessary. 

I don't think you should use `assert` in a model. Raise an error if you need to. 

Seeds should contain things that the system will break without, right? So shouldn't that include at least one loan status (active)?

elliottjf: The main challenge with the root division is not code complexity, but simply the endless discussions around it.  The majority of the dependencies are around building the unit test and migration data.  Beyond access control, the division hierarchy is key to resolving which option sets and custom fields to use for a particular entity since these are potentially customized per division.

Anyways, the latest should reflect all of Mel's recommendations, and I've only logged a small fraction of my own that has gone into the code review discussions and updates.  

A couple of notes, the root division primary key was a quick hack when I first built out the prototype flows, and had been removed quite a while back.  That had been replaced with a special 'internal_name', but I've now removed that and the code assumes there is a unique root node without a parent.  The current live code does not verify that Division.root returns a value.  This results in obscure errors in the unit test execution when the root division is not properly initialized.  I have a version of the Division.root logic implemented, but commented out, which was very useful in getting the unit tests working again once the logic to autocreate the root division was removed.  I'm still partial to re-enabling the version of Division.root which throws a helpful exception if the root division is missing.

PR #31, opened 2016-01-25
3738 loan response migration

elliottjf:
also provides for more flexible custom model associations, which allows questionnaire responses to be shared between loans

PR #35, opened 2016-01-26
3738 loan response migration

meltheadorable:
old PR #31


elliottjf:
Okay.  The latest from `develop` is now merged both in this branch and the base code review branch.  Unfortunately this makes this code review even bigger now because of the various updates needed to reconcile the translatable interface changes.

app/models/concerns/custom_field_addable.rb
```diff
@@ -0,0 +1,163 @@
...
+  def custom_field(field_identifier, required: true)
+    field_set = resolve_custom_field_set(required: required)
+    field_set.field(field_identifier, required: required)  if field_set
+  end
+
+  def custom_field?(field_identifier)
+    self.class.custom_field?(field_identifier, model: self)
+  end
+
+  def method_missing(method_sym, *arguments, &block)
```
meltheadorable: Would it be possible to document these metaprogramming methods? There's quite a lot of complexity here.

app/models/concerns/custom_field_addable.rb
```diff
@@ -0,0 +1,163 @@
...
+  def respond_to_missing?(method_sym, include_private = false)
+    attribute_name, action = match_dynamic_method(method_sym)
+    if action
+      true
+    else
+      super
+    end
+  end
+
+  def match_dynamic_method(method_sym)
```
meltheadorable: See above re: documentation

app/models/concerns/translatable.rb
```diff
@@ -2,7 +2,7 @@ module Translatable
   extend ActiveSupport::Concern
 
   included do
-    has_many :translations, as: :translatable
+    has_many :translations, as: :translatable, autosave: true
```
meltheadorable: Could you explain the rationale behind making these autosave? Part of the reason I originally chose not to when refactoring translatable was because with the ability to assign these as if they were normal ActiveRecord attributes, they would exhibit the same saving behavior as the rest of the object and other types of associations (i.e. only when using `create` or `update_attributes`, rather than `build`, `assign_attributes` or `new`). Ordinarily, you wouldn't want the line `@instance.label = 'My new label'`, for instance, to persist immediately.

app/models/loan.rb
```diff
@@ -66,8 +67,12 @@ class Loan < ActiveRecord::Base
   # beware, the methods generated by this include will fail without the corresponding OptionSet records existing in the database
   attr_option_settable :status, :loan_type, :project_type, :public_level
 
-  # results in linked CustomModel instances with the associated CustomFieldSet schema
-  attr_custom_model_linkable :loan_criteria, :loan_post_analysis
+  # provides acess to linked CustomModel instances with the associated CustomFieldSet schema.
+  # note these instances will be 'owned' by the loan's organization and potentially shared by more than one loan
```
meltheadorable: I thought we had discussed duplicating this data rather than hiding that ownership relationship between the custom instances and organizations? Am I misremembering? The use of `has_one_` here strongly implies that the loan, not the organization, is the owning instance for the data.

meltheadorable:
The migration logic itself is, as usual, very elegant. I just had a few questions about the change to make translations autosave, ownership of custom instances, and would like to see some documentation around the new metaprogramming as I'm somewhat apprehensive about its clarity and maintainability.

app/models/concerns/translatable.rb
elliottjf: With the autosave enabled and the other change to make sure the Translation instance which is included in the association list is updated when the field setter is used, things now behave as I believe is desired:  The change remains transient until the parent record is saved, and then the dependent change is persisted automatically with the parent.  
Without both changes, a translatable field was only getting saved with the initial record creations, but changes to those fields were never getting saved.

app/models/loan.rb
elliottjf: I believe that we are still waiting for confirmation from Brendan if he wants the questionnaire data duplicated, and the initial feedback was to just ignore the cross-links.

app/models/concerns/custom_field_addable.rb
elliottjf: more detailed documentation now provided: https://github.com/sassafrastech/madeline_system/commit/0151102ca25ae9d91dd1c3c73d24d99f2036e83a

PR #40, opened 2016-01-28
3998: Moves layout files around to accommodate embeddable templates

meltheadorable:


meltheadorable:
Merging this as it's a relatively minor renaming

PR #41, opened 2016-01-28
3998: Wordpress Template Fixes

meltheadorable:
* Adds sample wordpress layout file for specs
* Refactors wordpress library logic to read from fixture in dev/test
* Namespaces wordpress-related routing logic
* Gives relative paths in environment configs for template fetch
* Factors wordpress update logic into `WordpressEmbeddable` concern
* Creates static pages controller for WP test page

### DIFF
--- a/app/controllers/concerns/wordpress_embeddable.rb
+++ b/app/controllers/concerns/wordpress_embeddable.rb
@@ -1,0 +1,37 @@
+module WordpressEmbeddable
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :update_template
+    helper_method :get_division_from_url
+  end
+
+  # def get_division_from_url
+  #   # @get_division_from_url ||= Rails.configuration.x.wordpress_template[:division_urls].select { |key, val|
+  #   #   request.url.match key
+  #   # }.values.first || default_division
+  # end
+
+  def default_division
+    :us
+  end
+
+  def get_division_from_url
+    division_urls = Rails.configuration.x.wordpress_template[:division_urls]
+    matching = division_urls.select { |expression, _| request.url.match expression }
+    division = matching.values.first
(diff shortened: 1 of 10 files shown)


app/controllers/concerns/wordpress_embeddable.rb
```diff
@@ -0,0 +1,37 @@
...
+  def update_template
+    template_path = "layouts/embedded/wordpress-#{get_division_from_url}"
+    return if template_exists?(template_path)
+    WordpressTemplate.update(
+      division: get_division_from_url,
+      base_uri: [request.protocol, request.host_with_port].join
+    )
+  end
+
+end
```
linesarefuzzy: I like what you did here.

app/controllers/static_pages_controller.rb
```diff
@@ -0,0 +1,7 @@
+class StaticPagesController < ApplicationController
+  include WordpressEmbeddable
+
+  def test
+    render inline: "<% provide :title, 'Test' %><h1>Test Successful</h1>", layout: 'application'
```
linesarefuzzy: Does `layout: application` need to be specified?

meltheadorable: For inline layouts, it does, otherwise it renders the HTML with no layout at all. If I created an actual file in the views folder with the contents, it would be unnecessary, but this really only exists so that we have a simple page that will render within an embedded wordpress template without other dependencies (like all the legacy loan stuff being working).

lib/wordpress_template.rb
```diff
@@ -1,14 +1,26 @@
 module WordpressTemplate
```
linesarefuzzy: I like what you did here too.

linesarefuzzy:
Looks great.

PR #43, opened 2016-01-30
3752:  Removes `Language` model from translatable

meltheadorable:
* Updates translatable concern to utilize locales instead of language model
* Updates related models to work correctly with new translatable
* Re-writes translatable example group specs to ensure functioning
* Deletes unneeded legacy models
* Updates schema for new translatable changes
* Conforms to rails expectations for method naming to ensure functions like `create` and `update_attributes` work correctly
* Changes behavior of translatable assignment methods to assign but not immediately save
* Updates several broken factories


app/models/legacy/loan.rb
```diff
@@ -112,7 +112,7 @@ def public_level_option_value
 
 
 
-  MIGRATION_STATUS_OPTIONS = TransientOptionSet.new(
+  MIGRATION_STATUS_OPTIONS = Legacy::TransientOptionSet.new(
```
elliottjf: These are also outstanding changes on another open PR.  And I've been needing to temporarily update in my latest branch to get the latest db migrate commands to function.  Really hoping that we can start getting branches merged down to develop and avoid duplicating already completed efforts.

elliottjf:
This all looks great to me.  What is still needed before it can be merged down to develop?

app/models/language.rb
```diff
@@ -2,44 +2,12 @@
...
+#  created_at :datetime
 #  id         :integer          not null, primary key
+#  locale     :string
 #  name       :string
-#  code       :string
-#  created_at :datetime
 #  updated_at :datetime
 #
 
 class Language < ActiveRecord::Base
```
hooverlunch: Is this model still used? I thought it was going away...

db/schema.rb
```diff
@@ -56,12 +56,12 @@
 
   create_table "custom_fields", force: :cascade do |t|
     t.datetime "created_at", null: false
-    t.integer "custom_field_set_id"
-    t.string "data_type"
-    t.string "internal_name"
-    t.string "label"
-    t.integer "parent_id"
-    t.integer "position"
+    t.integer  "custom_field_set_id"
```
hooverlunch: I hate seeing these microscopic changes in the schema file. There is a gem that fixes this, called fix-db-schema-conficts or something.

spec/support/modules/translatable.rb
```diff
@@ -1,63 +1,70 @@
...
+shared_examples_for "translatable" do |attributes|
   let(:translatable_model) { create(described_class.to_s.underscore) }
-  context 'with translations' do
-    before do
-      column_names.each do |column_name|
-        translatable_model.send("set_#{column_name}".to_sym, Faker::Lorem.sentence)
-      end
-    end
 
+  attributes.each do |attribute|
```
hooverlunch: Interesting. I have never tried defining contexts in a block before. Why not!?

app/models/language.rb
meltheadorable: Last word I had on this was that we were going to bypass this model for the vast majority of the translation code, but that we were still going to need it to resolve locales to language names for the purposes of, for instance, crafting dropdowns for locale selection and for some things on the admin side for assigning multiple translations for a field at once with friendly language names. As far as I could tell, rails provides no other way of getting the language name for a locale.

hooverlunch:
These changes look very nice. Seems like a big improvement. I don't pretend to understand the full history here but the code looks fine. I am assuming that the test coverage is good here so that the refactor is well covered by them. This is also not functionality that is prone to hard-to-detect bugs. If the translation system is failing we will know about it quickly I reckon! Anyway this looks ready to merge from my standpoint!

db/schema.rb
elliottjf: I think the recent gemfile version updates somehow triggered all of these deltas.

meltheadorable: We're using `fix-db-schema-conflicts`, I have no idea why it doesn't seem to be doing its job here.

app/models/language.rb
hooverlunch: What I do in ELMO is just have an entry in the translation file called 'locale_name'. In en.yml it's 'English', in fr.yml it's 'Français', etc. Then you can grab them all by iterating over the locales you have defined and doing `t('locale_name', locale: l)`

meltheadorable: That's a good option, I'd considered it but wasn't sure how well it would actually work or whether there was a possibility there was other metadata we'd want/need to store about a language, so I'd decided to leave the model. Can switch that over though and delete the model entirely. Always happy to detonate a little more code.

hooverlunch: :D

On Wed, Feb 3, 2016 at 12:20 PM, Melody <notifications@github.com> wrote:

> In app/models/language.rb
> <https://github.com/sassafrastech/madeline_system/pull/43#discussion_r51754545>
> :
>
> >  #  name       :string
> > -#  code       :string
> > -#  created_at :datetime
> >  #  updated_at :datetime
> >  #
> >
> >  class Language < ActiveRecord::Base
>
> That's a good option, I'd considered it but wasn't sure how well it would
> actually work or whether there was a possibility there was other metadata
> we'd want/need to store about a language, so I'd decided to leave the
> model. Can switch that over though and delete the model entirely. Always
> happy to detonate a little more code.
>
> —
> <https://github.com/sassafrastech/madeline_system/pull/43/files#r51754545>
> .
>



-- 
Tom Smyth

touchstonecohousing.org

PR #44, opened 2016-02-02
3738 media migration

elliottjf:


db/migrate/20160201223840_add_uploader_to_media.rb
```diff
@@ -0,0 +1,5 @@
+class AddUploaderToMedia < ActiveRecord::Migration
+  def change
+    add_reference :media, :uploader, references: :people
```
linesarefuzzy: Add foreign key?

linesarefuzzy:
Looks good. I assume you've tested it with some actual media and it works? Go ahead and resolve conflicts and merge.

db/migrate/20160201223840_add_uploader_to_media.rb
elliottjf: added!

elliottjf:
yep. tested with actual media.  note that the current media uploader implementation is hardwired to 'uploads'.  on my system, i just mapped around that with symlinks, but we'll likely want to revisit before proceeding with a production configuration.  (i didn't want to mess with mel's original media work as part of this branch)

PR #45, opened 2016-02-05
4060: Add show loan page with styled tabs

emitche:
Add simple show loan page tabs


app/assets/stylesheets/admin/loans/_colors.css.scss
```diff
@@ -0,0 +1,9 @@
+$green1: rgb(115,200,100);
```
emitche: These colors will be used later in timeline and calendar.

PR #53, opened 2016-02-24
3738 - fix translation migration to use locale instead of language_id

elliottjf:
simple fix to translations migration to use a corresponding locale instead of the language_id


emitche:
This looks okay to me, but someone else who does more migrations and back end work should take a look. Thanks.

elliottjf:
this was mostly redundant with an existing fix

PR #54, opened 2016-03-01
3769 timeline show logs

emitche:
- Add display of project logs within project steps on the loan timeline.
- Add ability to delete project logs.


app/views/admin/logs/_step_logs.html.slim
```diff
@@ -0,0 +1,27 @@
...
+        - else
+          span.no-logs
+            = t("log.no_logs")
+      span.icon-box.add-log.condensed.text-success.last-col
+        i.fa.fa-plus-circle
+
+      - if step.logs_count > 0
+        section.logs
+          header.row
+            span.col-lg-8.first-col.section-heading Logs
```
emitche: Use translated text for line 18 and 23.

app/controllers/admin/project_logs_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Admin::ProjectLogsController < Admin::AdminController
+  def destroy
+    @log = ProjectLog.find(params[:id])
+    @step = ProjectStep.find(@log.project_step_id)
+    @loan = Loan.find(@step.project_id)
+
+    if @log.destroy
+      redirect_to admin_loan_path(@loan), notice: I18n.t(:notice_deleted)
```
emitche: This takes us back to the default view for a loan page, which shows the details tab. However, when a project log is deleted, we are on the timeline tab under a specific project step. Not sure if we want this redirect to reflect a desired page flow. And if so, not sure how to go about doing that.

meltheadorable: I think if you do `redirect_to admin_loan_path(@loan, anchor: 'timeline')...` it will redirect you to the timeline tab, but I'm not sure what the easiest way is to make sure that the correct step's log view is re-opened.

meltheadorable:
The code looks good! The green highlight color being used as the background for completed logs though makes the text really hard to read. Maybe it would be good to have it match the notice color for successful deletion? (#dff0d8)

It would be nice if we could get it to auto-open the log list for the project step as well as switching to the timeline tab, but I'm not sure the best way to do it right now and I think we can merge it without that for now. I don't need to see it again after you make the little tweaks mentioned.

app/controllers/admin/project_logs_controller.rb
emitche: Jumping down to the correct step will take some more planning. I tested out `anchor: 'timeline'`, which works, except it takes user to the very first project step, which is the top of `#timeline`.  I'll add jumping down to the correct step to the backlog as a tweak and think about that.

emitche:
Border and background colors on project steps are currently stubbed. There is another feature that will specific border and background colors based on how late or on time a step is.

I went ahead and used the Bootstrap palette for now.

Thanks!

app/controllers/admin/project_logs_controller.rb
meltheadorable: Yeah, that behavior might actually be browser-dependent. At least for me, on chrome beta channel for mac, it redirects me back to however far down I had been scrolled when I hit delete, but the project logs are hidden again instead of open. In any case, making a separate issue as a tweak is probably a good plan, finding a solution that's compatible with the tabs might be tricky.

PR #57, opened 2016-03-07
4206 tab colors

emitche:
Fix bug related to tabs not displaying with gray background as expected when active.


app/views/admin/loans/_tab.slim
```diff
@@ -1,4 +1,3 @@
-li role="presentation"
-  - classes = [*classes].join if defined? classes
```
emitche: I don't know who added this in, but I changed something in parent template that makes this obsolete.

PR #58, opened 2016-03-10
3766 timeline batch actions

emitche:
- Adds front end of batch action on the timeline of project steps within a loan


app/views/admin/loans/timeline/batch/_adjust_dates_modal.html.slim
```diff
@@ -0,0 +1,31 @@
+div.modal.fade id="adjust-dates-modal" tabindex="-1"
+  aria-labelledby="adjust-dates-label"
+  div.modal-dialog
+    div.modal-content
+      div.modal-header
+        button type="button" class="close" data-dismiss="modal"
+          aria-label="<%= t(:cancel) %>"
```
emitche: Fix syntax.

config/locales/en.yml
```diff
@@ -63,6 +63,7 @@ en:
   cancel: "Cancel"
   city: "City"
   clear: "Clear"
+  close_button: "&times;"
```
emitche: Remove and replace back in template itself.

emitche:
Given the changes to #59, I will need to make sure those changes are included in this branch.

emitche:
The changes made to ensure multi-select and batch actions do not interfere with other forms on the page is already merged into this branch.

emitche:
Convert JS to Backbone.

emitche:
Both the select steps and batch actions JS files from the timeline have been converted to Backbone CoffeeScript for consistency.

I did notice that it may be sensible to disable batch actions when no items are checked. I'm not sure how we'd want to implement that feature visually. But we can discuss this.

app/assets/javascripts/backbone/views/timeline_batch_actions_view.coffee
```diff
@@ -0,0 +1,27 @@
+class MS.Views.TimelineBatchActionsView extends Backbone.View
+
+  el: 'body'
+
+  events: (params) ->
+    "confirm:complete .batch-actions .batch-action": 'adjust_form'
+    "click #adjust-dates-confirm": 'hide_adjust_dates_modal'
+    "shown.bs.modal #adjust-dates": 'show_adjust_dates_modal'
```
hooverlunch: I believe our styleguide says to prefer single quotes in JS. Can you please keep that in mind going forward but ok to not fix these for now.

app/assets/javascripts/backbone/views/timeline_select_steps_view.coffee
```diff
@@ -0,0 +1,76 @@
...
+    'change .select-step': 'remember_checked'
+    'click #choose-all': 'control_all'
+    'click #check-all-ctrl': 'check_all'
+    'click #uncheck-all-ctrl': 'uncheck_all'
+    'click #check-completed-ctrl': 'check_completed'
+    'click #check-incomplete-ctrl': 'check_incomplete'
+ 
+  check_all: (e) ->
+    control = e.currentTarget
+    $inputs = $(control).closest(".timeline").find('.select-step')
```
hooverlunch: I like the $ for jquery objects convention...

app/assets/javascripts/backbone/views/timeline_select_steps_view.coffee
```diff
@@ -0,0 +1,76 @@
...
+    this.check_items($inputs)
```
hooverlunch: You should substitute `@` for `this.` in CoffeeScript. It compiles to the same thing and is shorter.

app/assets/javascripts/backbone/views/timeline_select_steps_view.coffee
```diff
@@ -0,0 +1,76 @@
...
+
+    $master_checkbox = $(control).closest(".btn-group").find("#choose-all")
```
hooverlunch: When selecting something by ID you don't need to scope it like this. There should only be one of them on the whole page, so @$('#choose-all') should be sufficient. If there are more than one, should change to class instead of id.

app/assets/javascripts/backbone/views/timeline_select_steps_view.coffee
```diff
@@ -0,0 +1,76 @@
...
+    else
+      this.uncheck_items($inputs)
+
+    this.remember_checked(e)
+
+  remember_checked: (e) ->
+    ids = [];
+
+    $('.select-step').each (index) ->
+      if (this.checked == true)
```
hooverlunch: It is very rare that you need to do `== true`. In this case, `this.checked` by itself is equivalent.

app/assets/javascripts/backbone/views/timeline_select_steps_view.coffee
```diff
@@ -0,0 +1,76 @@
...
+
+    if (control.checked == true)
+      this.check_items($inputs)
+    else
+      this.uncheck_items($inputs)
+
+    this.remember_checked(e)
+
+  remember_checked: (e) ->
+    ids = [];
```
hooverlunch: No need for `;` in coffeescript.

app/assets/javascripts/backbone/views/timeline_select_steps_view.coffee
```diff
@@ -0,0 +1,76 @@
...
+
+    $('.select-step').each (index) ->
+      if (this.checked == true)
+        ids.push($(this).data('id'))
```
hooverlunch: I assume by `$(this)` you mean to get a jquery object for the element controlled by this view? If so, then you should use `@$el`. See http://backbonejs.org/#View-$el

app/assets/stylesheets/admin/loans/timeline.scss
```diff
@@ -118,4 +118,15 @@
...
     vertical-align: middle;
   }
+
+  input[type="number"] {
+    max-width: 5em;
+    text-align: center;
+  }
+}
+
+#adjust-dates-modal {
```
hooverlunch: I'm curious why this block is not scoped within `.timeline` or similar. As it is, it is a global style declaration living at the bottom of a page-specific style file, which can lead to unintended consequences. Let's discuss.

app/views/admin/loans/show.html.slim
```diff
@@ -9,9 +9,14 @@
 
 - content_for :scripts
   = javascript_include_tag "admin/forms/forms"
-  = javascript_include_tag "admin/timeline/select_all"
   = javascript_include_tag "admin/expand"
   = javascript_include_tag "admin/calendar/calendar"
+
```
hooverlunch: Let's talk about per-page JS.

app/assets/javascripts/backbone/views/timeline_batch_actions_view.coffee
```diff
@@ -0,0 +1,27 @@
...
+  
+  adjust_form: (e) ->
+    item = e.currentTarget
+    method_key = $(item).attr("data-method-key")
+    action_key = $(item).attr("data-action-key")
+
+    $form = $(item).closest("form")
+
+    $form.find("#set-method").attr("value", method_key)
```
hooverlunch: For things like this I recommend using e.g. `find('input[name=_method]')`. The input already has a unique attribute so it's better to select it using that instead of adding a new ID attrib to it. Try to minimize the number of attributes you're adding just so you can grab something from jQuery.

app/views/admin/loans/timeline/batch/_adjust_dates_modal.html.slim
```diff
@@ -0,0 +1,28 @@
+div.modal.fade id="adjust-dates-modal" tabindex="-1" aria-labelledby="adjust-dates-label"
+  div.modal-dialog
+    div.modal-content
+      div.modal-header
+        button type="button" class="close" data-dismiss="modal" aria-label== t(:cancel)
+            span aria-hidden="true" &times;
```
hooverlunch: Extra indent?

config/locales/en.yml
```diff
@@ -150,6 +162,7 @@ en:
     logs: "Logs"
     no_steps: "No timeline steps to display."
     project_step: "project step"
+    project_steps: "project steps"
```
hooverlunch: The Rails way to do plurals like this is to do
```
project_step:
  one: "project step"
  other: "project steps"
```

Then when you translate you do e.g. `t("project_step", count: 2)`

hooverlunch:
Great work so far! Let's chat about a few of these things, and also can you show me how this looks in the browser (my env is not yet setup)...

app/assets/stylesheets/admin/loans/timeline.scss
emitche: This is scoped inside .timeline-header.

app/views/admin/loans/show.html.slim
emitche: Now that we have moved over to Backbone, many JS files will need to be converted to Backbone. And yes, let's talk about per-page JS.

emitche:
I merged this by mistake before meeting with you, @hooverlunch , thinking it was the 3768 branch. However, this branch is marked revise and merge, so that should be fine. I did make the requested changes. Anything else that needs adjustment, we can discuss.

PR #59, opened 2016-03-16
3767 multiselect steps

emitche:
Refactor multi-select form
 
Refactor selecting multiple steps so that multi-select and batch actions will not interfere with other forms on the page.


app/assets/javascripts/admin/timeline/select_all.coffee
```diff
@@ -1,34 +1,51 @@
...
 
   $('#check-incomplete-ctrl').click (e) ->
     uncheck_all(e)
     check_incomplete(e)
+    remember_checked(e);
+
+  $('.select-step').change (e) ->
+    remember_checked(e);
+
+  remember_checked = (e) ->
```
emitche: Would like a different method name, but can't think of one.

PR #60, opened 2016-03-25
3768 duplicate step

emitche:
Add duplicate step modal to timeline

- Includes front-end only of duplicate step feature


app/models/project_step.rb
```diff
@@ -123,4 +123,46 @@ def scheduled_bg
...
+  end
+
+  def scheduled_weekday_key
+    self.nth_weekday.to_s + "_" + self.scheduled_weekday.downcase
+  end
+
+  def nth_weekday
+    day = scheduled_day.to_i
+
+    if (day <= 8)
```
emitche: < 8 (days 1-7)

app/views/admin/project_steps/_duplicate_step_form.html.slim
```diff
@@ -0,0 +1,11 @@
...
+
+  div.form-group.row
+    input.repeat-once type="radio" name="repeat_duration"value="repeat_once" checked="checked"
+      span.bold = t("project_step.once")
+
+  div.form-group.row
+    input.custom-repeat type="radio" name="repeat_duration" value="custom_repeat"
+      span.bold = t("project_step.repeat")
+
+  = render "admin/project_steps/repeat_options", step: step, f: f
```
meltheadorable: I really like the way you've been choosing to organize these partials, keeps all of them pretty easily digestible and straightforward.

app/models/project_step.rb
```diff
@@ -123,4 +123,46 @@ def scheduled_bg
...
+    elsif (15 <= day) && (day < 22)
+      3
+    elsif (22 <= day) && (day < 29)
+      4
+    else
+      5
+    end
+  end
+
+  def nth_day
```
meltheadorable: There's a built-in method in rails that actually handles this called `ordinalize`:

```ruby
date = DateTime.now
# => Mon, 04 Apr 2016 19:26:47 -0400
date.day.ordinalize
# => "4th"
```

app/models/project_step.rb
```diff
@@ -123,4 +123,46 @@ def scheduled_bg
...
+
+  def scheduled_weekday
+    Date::DAYNAMES[self.scheduled_date.wday]
+  end
+
+  def scheduled_weekday_key
+    self.nth_weekday.to_s + "_" + self.scheduled_weekday.downcase
+  end
+
+  def nth_weekday
```
meltheadorable: I'm struggling to think of a better name for this, but I found myself a little confused by what it was supposed to be doing judging based just on the name. I wonder whether we should try to come up with something a little more descriptive.

meltheadorable:
The overall approach here looks great, though it's tough to judge whether stuff like the form values will end up working out until the backend piece is also in place.

emitche:
Convert JS to Backbone.

meltheadorable:
The changes here look really good! I think you're all set.

PR #65, opened 2016-04-20
3768 - duplicate step back-end

elliottjf:


app/controllers/admin/project_steps_controller.rb
```diff
@@ -27,6 +27,41 @@ def update
     render partial: "/admin/project_steps/project_step", locals: {step: @step, mode: valid ? :show : :edit}
   end
 
+
+  def duplicate_step
+    puts "duplicate_step - params: #{params}"
```
hooverlunch: This should be removed.

app/models/concerns/translatable.rb
```diff
@@ -131,6 +131,17 @@ def translate(*terms)
     end
   end
 
+  # clones all translations associated with current instance into target translatable instance as transient values
+  def clone_translations(destination)
+    translations.each do |translation|
+      destination.translations.build(
+          translatable_attribute: translation.translatable_attribute,
+          locale: translation.locale,
+          text: translation.text
```
hooverlunch: Remove extra indent.

app/models/legacy/project_event.rb
```diff
@@ -49,7 +49,7 @@ def self.migrate_all
       # on the fly to handle the orphaned logs
       max = self.connection.execute("select max(id) from ProjectEvents").first.first
       puts "setting projects_step_id_seq to: #{max+1000}"
-      ::ProjectStep.recalibrate_sequence(gap: 1000)
+      ::ProjectStep.recalibrate_sequence(id: max+1000)
```
hooverlunch: Space around `+` please.

app/models/project_step.rb
```diff
@@ -167,12 +171,18 @@ def scheduled_bg
     end
   end
 
+  # should probably rename these methods if they encapsulates special handling needed for the duplicate step feature
```
hooverlunch: We should discuss this. I'm not quite getting you...

app/models/project_step.rb
```diff
@@ -196,6 +206,67 @@ def week_of_scheduled_date
     end
   end
 
+  # todo: confirm if we want to hardcode a limit for end_date bound repeats. and if so, what should that be?
+  MAX_OCCURRENCES = 100
```
hooverlunch: I would say 60. 100 seems like a lot at one time. I worry about performance degrading. Please put constants at top of file.

app/models/project_step.rb
```diff
@@ -196,6 +206,67 @@ def week_of_scheduled_date
...
+
+  def duplicate_series(frequency, time_unit, month_repeat_on, num_of_occurrences, end_date)
+    if time_unit == 'days' || time_unit == 'weeks'
```
hooverlunch: I thought time_unit was a symbol, which means this will always be false.

app/models/project_step.rb
```diff
@@ -196,6 +206,67 @@ def week_of_scheduled_date
...
+    end
+
+    results = []
+    remaining = num_of_occurrences || MAX_OCCURRENCES
+
+    allow_error = true
+    last_date = duplicate_step_base_date
+    while remaining > 0 do
+      next_date = apply_time_interval(last_date, frequency, time_unit, month_repeat_on)
+      break  if end_date && next_date > end_date
```
hooverlunch: Remove extra space after `break` please.

app/models/project_step.rb
```diff
@@ -196,6 +206,67 @@ def week_of_scheduled_date
...
+    if time_unit == 'days' || time_unit == 'weeks'
+      interval = frequency.send(time_unit)
+    end
+
+    results = []
+    remaining = num_of_occurrences || MAX_OCCURRENCES
+
+    allow_error = true
+    last_date = duplicate_step_base_date
+    while remaining > 0 do
```
hooverlunch: Ruby way to do this would be `(num_of_occurrences || MAX_OCCURRENCES).times do`. Then you can remove the decrement line at the bottom.

app/models/project_step.rb
```diff
@@ -196,6 +206,67 @@ def week_of_scheduled_date
...
+      reference_date = date.beginning_of_month + interval - 1.month
+      Chronic.parse("#{month_repeat_on} of next month", now: reference_date)
+    end
+  end
+
+  def create_duplicate(date = nil, should_persist: true, allow_error: true)
+    begin
+      date ||= scheduled_date
+      new_step = ProjectStep.new(
+          project: project,
```
hooverlunch: Remove extra indent please.

app/models/project_step.rb
```diff
@@ -196,6 +206,67 @@ def week_of_scheduled_date
...
+          agent: agent,
+          step_type_value: step_type_value,
+          scheduled_date: date,
+          original_date: nil,
+          completed_date: nil,
+          is_finalized: false,
+      )
+      clone_translations(new_step)  # this will create transient copies of all of the source translatable attributes
+      new_step.save  if should_persist
```
hooverlunch: Please remove these extra spaces throughout.

app/models/project_step.rb
```diff
@@ -196,6 +206,67 @@ def week_of_scheduled_date
...
+      new_step = ProjectStep.new(
+          project: project,
+          agent: agent,
+          step_type_value: step_type_value,
+          scheduled_date: date,
+          original_date: nil,
+          completed_date: nil,
+          is_finalized: false,
+      )
+      clone_translations(new_step)  # this will create transient copies of all of the source translatable attributes
```
hooverlunch: This is nice.

app/models/project_step.rb
```diff
@@ -196,6 +206,67 @@ def week_of_scheduled_date
...
+      new_step.save  if should_persist
+      new_step
+      # note, would likely want to also copy custom fields at the point in time which we expect those to be used on ProjectSteps
```
hooverlunch: Please break this comment into two lines. Line limit is 100 +/- 10

app/models/project_step.rb
```diff
@@ -196,6 +206,67 @@ def week_of_scheduled_date
...
+    rescue StandardError => e
+      Rails.logger.error("create_duplicate error: #{e}")
+      raise e  if allow_error
+      {error: e.inspect}  #todo: confirm if the front-end logic desires this data or if it should be simply omitted
```
hooverlunch: Space inside `{}` please

app/models/project_step.rb
```diff
@@ -196,6 +206,67 @@ def week_of_scheduled_date
...
+          scheduled_date: date,
+          original_date: nil,
+          completed_date: nil,
+          is_finalized: false,
+      )
+      clone_translations(new_step)  # this will create transient copies of all of the source translatable attributes
+      new_step.save  if should_persist
+      new_step
+      # note, would likely want to also copy custom fields at the point in time which we expect those to be used on ProjectSteps
+    rescue StandardError => e
```
hooverlunch: `StandardError =>` is not needed.

app/controllers/admin/project_steps_controller.rb
```diff
@@ -27,6 +27,41 @@ def update
...
+      result = step.duplicate_series(frequency, time_unit, month_repeat_on, num_of_occurrences, end_date)
+
+    else
+
+      new_step = step.create_duplicate(should_persist: false)
+      result = [new_step]
+
+    end
+
+    render json: result
```
hooverlunch: Should be returning a 4xx code if the process errors.

config/routes.rb
```diff
@@ -11,7 +11,17 @@
     resources :dashboard, only: [:index]
     resources :loans
     resources :organizations
-    resources :project_steps
+    resources :project_steps do
+      collection do
+        delete '', to: 'project_steps#batch_destroy'
+        patch 'adjust_dates', to: 'project_steps#adjust_dates'
+        patch 'finalize', to: 'project_steps#finalize'
```
hooverlunch: You can write the above as follows:

```
delete :batch_destroy
patch :adjust_dates
patch :finalize
```

config/routes.rb
```diff
@@ -11,7 +11,17 @@
...
+      end
+      member do
+        # get 'duplicate_step', to: 'project_steps#show_duplicate_step'
+        post 'duplicate_step', to: 'project_steps#duplicate_step'
```
hooverlunch: The `_step` here is redundant. Please rename the route to just `duplicate`.

app/policies/project_step_policy.rb
```diff
@@ -1,2 +1,7 @@
 class ProjectStepPolicy < ApplicationPolicy
+
+  def duplicate_step?
```
hooverlunch: Rename to `duplicate?`

app/views/admin/project_steps/_duplicate_step_modal.html.slim
```diff
@@ -9,21 +9,27 @@ div.modal.fade.duplicate-step id="duplicate-step-modal-#{step.id}" tabindex="-1"
           = t("project_step.duplicate_step")
 
       div.modal-body
-        = form_tag do |f|
+        = form_tag("/admin/project_steps/#{step.id}/duplicate_step", method: 'post') do |f|
```
hooverlunch: Please use route helpers here. So `duplicate_admin_project_step_path(step)` (I think that should work).

app/views/admin/project_steps/_duplicate_step_modal.html.slim
```diff
@@ -9,21 +9,27 @@ div.modal.fade.duplicate-step id="duplicate-step-modal-#{step.id}" tabindex="-1"
...
 
           div.form-group.row
-            input.repeat-once type="radio" name="repeat_duration"value="repeat_once" checked="checked"
-              span.bold = t("project_step.once")
+            = label_tag :repeat_once
```
hooverlunch: I think a label tag with no attribs will work just fine.

hooverlunch:
Thank you for this. In general it's looking solid. A few changes requested above. Let's discuss the method naming thing when you're online today.

app/models/project_step.rb
elliottjf: Yeah, this was actually irrelevant code which should have been removed after some refactoring.

app/models/project_step.rb
elliottjf: Right now you'll get the default fatal exception response if there is an error on the first record.  Is that sufficient for you AJAX response, or do you want a custom JSON coded error response?  
And I've updated to strip out any partials failures (if first record succeeds, but some of the remaining records fail)

hooverlunch: Yes 500 is fine. We are not actually /expecting/ any errors, right? Like
we're not doing this to catch validation errors or some such?

On Thu, Apr 21, 2016 at 12:51 PM, Joseph Elliott <notifications@github.com>
wrote:

> In app/models/project_step.rb
> <https://github.com/sassafrastech/madeline_system/pull/65#discussion_r60616472>
> :
>
> > +          project: project,
> > +          agent: agent,
> > +          step_type_value: step_type_value,
> > +          scheduled_date: date,
> > +          original_date: nil,
> > +          completed_date: nil,
> > +          is_finalized: false,
> > +      )
> > +      clone_translations(new_step)  # this will create transient copies of all of the source translatable attributes
> > +      new_step.save  if should_persist
> > +      new_step
> > +      # note, would likely want to also copy custom fields at the point in time which we expect those to be used on ProjectSteps
> > +    rescue StandardError => e
> > +      Rails.logger.error("create_duplicate error: #{e}")
> > +      raise e  if allow_error
> > +      {error: e.inspect}  #todo: confirm if the front-end logic desires this data or if it should be simply omitted
>
> Right now you'll get the default fatal exception response if there is an
> error on the first record. Is that sufficient for you AJAX response, or do
> you want a custom JSON coded error response?
>
> And I've updated to strip out any partials failures (if first record
> succeeds, but some of the remaining records fail)
>
> —
> <https://github.com/sassafrastech/madeline_system/pull/65/files/7c9699b0fe12dfcb29808bb4c6331889998993ea#r60616472>
>



-- 
Tom Smyth

touchstonecohousing.org

app/controllers/admin/project_steps_controller.rb
```diff
@@ -27,6 +27,12 @@ def update
     render partial: "/admin/project_steps/project_step", locals: {step: @step, mode: valid ? :show : :edit}
   end
 
+  def duplicate
+    step = ProjectStep.find(params[:id])
+    authorize step
+    render json: step.duplication.perform(params[:duplication])
```
hooverlunch: Wow so much cleaner!!!

hooverlunch:
Really well done on the refactor. Thanks for working with me on this.

elliottjf:
Thanks.  Yes, much cleaner this way!  Is this ready to merge?

PR #68, opened 2016-04-22
3743 division select

elliottjf:
ready for review


app/models/user.rb
```diff
@@ -41,4 +41,25 @@ class User < ActiveRecord::Base
...
+
+  def accessible_divisions
+    # Todo: Confirm desired sort order
+    accessible_division_ids.map { |id| Division.find_safe(id) }.compact
+  end
+
+  def default_division
+    id = roles.where(resource_type: :Division, name: [:admin]).pluck(:resource_id).first
+    Division.find(id)
+  end
```
hooverlunch: There is a lot of code duplication in these three methods. Please refactor to make tighter and more efficient.

app/controllers/application_controller.rb
```diff
@@ -3,11 +3,47 @@ class ApplicationController < ActionController::Base
...
+  def selected_division_id
+    session[:selected_division_id]
+  end
+
+  def set_selected_division_id(id)
+    id = nil if id.blank?
+    session[:selected_division_id] = id
+  end
+
+  def division_select_options
```
hooverlunch: This should be in application_helper, not the controller.

app/controllers/application_controller.rb
```diff
@@ -3,11 +3,47 @@ class ApplicationController < ActionController::Base
...
+
+  # Represents the current division filter applied to index views.
+  def selected_division
+    id = selected_division_id
+    if id
+      Division.find_safe(id)
+    else
+      nil
+    end
+  end
```
hooverlunch: This should be condensed to `(id = selected_division_id) ? Division.find_safe(id) : nil`.

app/controllers/application_controller.rb
```diff
@@ -3,11 +3,47 @@ class ApplicationController < ActionController::Base
...
+  helper_method :division_select_options
+
+  # Represents the division to use when creating new entities.
   def current_division
-    # TODO
-    Division.root
+    selected_division || default_division
+  end
+
+  # Division to uses for new entities if a division is not specifically selected
```
hooverlunch: Fix typo please.

app/controllers/application_controller.rb
```diff
@@ -3,11 +3,47 @@ class ApplicationController < ActionController::Base
...
+      Division.find_safe(id)
+    else
+      nil
+    end
+  end
+
+  def division_index_filter
+    selected = selected_division
+    ids = selected ? selected.self_and_descendant_ids : current_user.accessible_division_ids
+    {division: ids}
```
hooverlunch: Shouldn't this be `{ division_id: ids }` if it's a where condition? I'm surprised this even works as is...

app/controllers/application_controller.rb
```diff
@@ -3,11 +3,47 @@ class ApplicationController < ActionController::Base
...
+  def selected_division
+    id = selected_division_id
+    if id
+      Division.find_safe(id)
+    else
+      nil
+    end
+  end
+
+  def division_index_filter
```
hooverlunch: Please add a comment for this method as it's not super clear what it's doing (especially the descendant_ids vs accessible_ids bit).

app/controllers/application_controller.rb
```diff
@@ -3,11 +3,47 @@ class ApplicationController < ActionController::Base
...
+  end
+
+  def selected_division_id
+    session[:selected_division_id]
+  end
+
+  def set_selected_division_id(id)
+    id = nil if id.blank?
+    session[:selected_division_id] = id
+  end
```
hooverlunch: Please move all these division related methods into a controller concern (DivisionSelectable?)

app/controllers/application_controller.rb
```diff
@@ -3,11 +3,47 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
-  helper_method :current_division
 
+  helper_method :selected_division_id
+  helper_method :division_select_options
```
hooverlunch: This can go away since it should really just be in a helper.

app/models/user.rb
```diff
@@ -41,4 +41,25 @@ class User < ActiveRecord::Base
   def name
     profile.try(:name)
   end
+
+  def accessible_division_ids
+    # Todo: Confirm what other roles types to include here.
```
hooverlunch: I don't know the answer to this. @linesarefuzzy?

app/views/layouts/admin/_page.html.slim
```diff
@@ -1,6 +1,13 @@
 - content_for :styles
   = stylesheet_link_tag "admin/content", :media => "all"
 
+- content_for :scripts
+
+  javascript:
+    $(function () {
```
hooverlunch: No space after `function` please.

app/views/layouts/admin/_page.html.slim
```diff
@@ -1,6 +1,13 @@
 - content_for :styles
   = stylesheet_link_tag "admin/content", :media => "all"
 
+- content_for :scripts
+
+  javascript:
```
hooverlunch: Please put this javascript block at the bottom of the template. This is our general convention (I know there are exceptions in this project).

app/views/layouts/admin/_page.html.slim
```diff
@@ -23,8 +30,10 @@ div
 
           .division.centered-box.nav.navbar-nav.col-sm-6
             span> Division:
-            select.form-control
-              option All
+            / Todo: Sort out proper css display handling here
```
hooverlunch: Not sure what is meant here, but this should be fixed and removed.

app/assets/javascripts/backbone/views/page_view.coffee
```diff
@@ -0,0 +1,12 @@
+class MS.Views.PageView extends Backbone.View
+
+  el: 'body'
```
hooverlunch: This should be just the form. Best to keep view scopes as narrow as possible.

app/assets/javascripts/backbone/views/page_view.coffee
```diff
@@ -0,0 +1,12 @@
...
+
+  events: (params) ->
+    'change select[name="division_id"]': 'selectDivision'
```
hooverlunch: If el is just the form, then this becomes just `'change select': 'selectDivision'`

app/assets/javascripts/backbone/views/page_view.coffee
```diff
@@ -0,0 +1,12 @@
...
+
+  selectDivision: (e) ->
+    $item = $(e.currentTarget)
+    $("input[name=redisplay_url]").val(window.location)
```
hooverlunch: Isn't this bound to lead to 404s errors, if, e.g., the user is on a specific loan page and then changes divisions?

app/controllers/admin/divisions_controller.rb
```diff
@@ -0,0 +1,14 @@
+class Admin::DivisionsController < ApplicationController
+  before_action :authenticate_user!
+  after_action :verify_authorized
+
+  def select
+    redisplay_url = params[:redisplay_url] || root_path
+    division_id = params[:division_id]
+    set_selected_division_id(division_id)
+    division = Division.find_safe(division_id)
+    authorize division || current_division
```
hooverlunch: I'm confused here. division_id gets saved, but then division may be null if find_safe fails, in which case a _different_ division is authorized against? If find_safe fails, shouldn't we not be storing the errant division_id?

hooverlunch:
Thanks Joseph. Good progress here. I have some changes and a few questions. Please reply inline. Thanks.

app/views/layouts/admin/_page.html.slim
elliottjf: Sure, and yeah, I was just copying what I saw in this project..  Should the 'content_for :styles' block also go at the bottom?

app/controllers/admin/divisions_controller.rb
elliottjf: The previous code was always just authorizing against Division.root.  I was trying to mirror that behavior but with with an accessible division.  Would it be better to just bypass the need for an 'authorize' call in this case?  I wasn't sure if that was preferred or off hand how to do that.

app/views/layouts/admin/_page.html.slim
elliottjf: This was a question that I raised to Slack yesterday.  I tried for a while, but I don't have the CSS skills to solve this.  Fuzzy pinged Elizabeth about it.

app/views/layouts/admin/_page.html.slim
hooverlunch: Actually no, that should go away.

app/views/layouts/admin/_page.html.slim
hooverlunch: Ok Fuzzy and I will deal with it.

app/models/user.rb
hooverlunch: Actually, thinking more about this, this code really belongs in the DivisionPolicy as a scope. We should talk more about this. This feature is not slated for the release today though so we can pick it up next week.

app/assets/javascripts/backbone/views/page_view.coffee
elliottjf: Currently the selected division controls the filtering applied to the index view, but it doesn't change any access control logic, so it's not going to trigger any 404 errors.  Do we want some different behavior around this?

app/views/layouts/admin/_page.html.slim
```diff
@@ -20,8 +20,7 @@ div
 
           .division.centered-box.nav.navbar-nav.col-sm-6
             span> Division:
-            / Todo: Sort out proper css display handling here
-            = form_tag(admin_divisions_select_path, id: 'select_division_form', class: 'form-control') do
+            = form_tag(admin_divisions_select_path, id: 'select_division_form') do
               = select_tag(:division_id, options_for_select(division_select_options, selected_division_id), class: 'form-control')
```
emitche: .form-control adds some Bootstrap styling to form items (like select and input). I made some changes to the CSS so the form displays inline as expected.

spec/support/modules/base_policy.rb
```diff
@@ -9,7 +9,9 @@
   let!(:child_division) { create(:division, parent: division) }
 
   context 'being a member of a parent division' do
-    let(:user) { create(:user, :member, division: parent_division) }
+    let(:user) {
+      create(:user, :member, division: parent_division)
+    }
```
hooverlunch: This can/should be one line.

app/policies/division_policy.rb
```diff
@@ -18,4 +18,30 @@ def update?
...
+    end
+
+    # This merges in child divisions of the divisions for which a user has been specifically
+    # granted access.
+    def accessible_ids
+      all_ids = base_accessible_ids.map do |id|
+        division = Division.find_safe(id)
+        division.self_and_descendants.pluck(:id) if division
+      end
+      all_ids.flatten.uniq.compact
```
hooverlunch: I think you need a `compact` after `flatten` to remove any `nil`s in case divisions aren't found.

app/policies/division_policy.rb
```diff
@@ -18,4 +18,30 @@ def update?
...
+    end
+
+    # List of division hierarchy nodes for which user has been granted access.
+    def base_accessible_ids
+      user.roles.where(resource_type: :Division, name: [:member, :admin]).pluck(:resource_id).uniq
+    end
+  end
```
hooverlunch: This looks soooo much better!

app/policies/division_owned_scope.rb
```diff
@@ -0,0 +1,7 @@
+class DivisionOwnedScope < ApplicationPolicy::Scope
+
+  def resolve
+    scope.where(division_id: user.accessible_division_ids)
```
hooverlunch: I'm not sure that you need to reach out of the Policy system and back into the user model here. Why not just construct the DivisionPolicy in this class directly instead of in User?

app/views/layouts/admin/_page.html.slim
```diff
@@ -48,3 +46,9 @@ div
             h1.page-title = yield :title
             .with-title = yield :with_title
           = yield
+
+- content_for :scripts
```
hooverlunch: Shouldn't need this.

app/policies/division_policy.rb
```diff
@@ -18,4 +18,30 @@ def update?
   def destroy?
     division_admin(division: @record)
   end
+
+  def select?
+    show?
+  end
+
+  class Scope < Scope
```
hooverlunch: We really should be testing this as so much depends on it. There are some good docs online for testing scopes in Pundit. Inside UserPolicySpec I would put two describe blocks, one for permissions, and one for scope. Create a few different divisions (child, parent, etc.) and users with different roles (admin, member, none) and ensure that the scope resolves to the right divisions. Shouldn't take long.

hooverlunch:
Looking much better, thanks! A few small changes and it should be good!

app/policies/division_policy.rb
elliottjf: Does it make a difference if the `compact` is before or after the `uniq`?

hooverlunch: No

sent via wireless telephone machine.

app/policies/division_policy.rb
elliottjf: Okay.  As I was verifying I discovered that Mel had actually already implemented quite a bit of unit testing, including `show` verification which was already dependent on the scope logic.  I'll go ahead and add some additional tests specific to scope.

hooverlunch: ...I'm not seeing how the `show` method is dependent on this new code you've written, but I could be missing something...

app/controllers/application_controller.rb
```diff
@@ -3,12 +3,6 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
-  helper_method :current_division
-
-  def current_division
```
meltheadorable: `current_division`, or at least a version of it, needs to be in the `ApplicationController`, the public side driving the wordpress stuff is dependent on it.

hooverlunch: Ah, good catch.

elliottjf: Ah.  Thanks!  I hadn't realized that.  I'm curious what will drive the selection here.  Will each division have their own public site?  It's not clear to me if this should be driven by the same logic and or helper method or made distinct.  Thoughts?

app/policies/division_policy.rb
elliottjf: ApplicationPolicy.show? verifies that the scope can resolve the given record id.
And I've added a 'Scope' block to the division_policy_spec, and I've factored out a division_owned_scope set of shared tests.  
I'm guessing you may have some feedback on how I could better factor the rspec code, so I'll give you a chance to review before I merge down to develop.

spec/policies/division_policy_spec.rb
```diff
@@ -44,4 +44,49 @@
...
+      division_scope(user, id: division_id)
+    end
+
+    def division_scope(user, conditions = nil)
+      DivisionPolicy::Scope.new(user, Division.where(conditions)).resolve
+    end
+  end
+
 end
+
```
hooverlunch: Very nice tests. Exactly what I was hoping for.

spec/support/modules/division_owned_scope.rb
```diff
@@ -0,0 +1,64 @@
...
+    def record_class(record_type)
+      record_type.to_s.camelize.constantize
+    end
+
+    def scope_class(record_type)
+      "#{record_type.to_s.camelize}Policy::Scope".constantize
+    end
+  end
+
+end
```
hooverlunch: Lovely.

hooverlunch:
Very nice work. I'm glad we took the time to do this properly. Thanks.

PR #69, opened 2016-04-22
4435 - project step unfinalization handling

elliottjf:


app/views/admin/project_steps/_edit_project_step.html.slim
```diff
@@ -31,7 +31,13 @@ div.panel-body
       .form-group.date.optional.project_step_completed_date
         = label_tag :is_finalized, "Finalized", class: "date optional col-sm-3 checkbox-label"
         .col-sm-9
-          = f.check_box :is_finalized, class: "step_form_right_9 finalized"
+          - if step.is_finalized_locked?
+            = step.is_finalized ? t("reply_yes") : t("reply_no")
+            / Todo: Confirm what display handling is desired here, and what asset to use if
+            / to be represented as an image
+            =< "(#{t('locked')})"
```
elliottjf: @linesarefuzzy Can you confirm what display handling you want here?

app/models/project_step.rb
```diff
@@ -135,6 +138,27 @@ def handle_original_date_logic
     end
   end
 
+  def handle_finalized_at
```
hooverlunch: These methods should be private (as generally should any that don't need to be called from outside the class).

app/models/project_step.rb
```diff
@@ -135,6 +138,27 @@ def handle_original_date_logic
...
+    if is_finalized && ! finalized_at
```
hooverlunch: No spc after ! please.

app/models/project_step.rb
```diff
@@ -135,6 +138,27 @@ def handle_original_date_logic
...
+      self.finalized_at = Time.now
+    elsif ! is_finalized && finalized_at
+      self.finalized_at = nil
+    end
+  end
+
+  # Validates that a step may not be unfinalized more than 24 hours since it was previously marked
+  # as finalized.  Note, should generally be avoided by front-end logic, but guards against edge
+  # cases.
```
hooverlunch: Nice comment!

app/models/project_step.rb
```diff
@@ -135,6 +138,27 @@ def handle_original_date_logic
...
+  def unfinalize_allowed
+    if is_finalized_changed? && ! is_finalized && is_finalized_locked?
+      errors.add(:is_finalized, I18n.t("project_step.unfinalize_disallowed"))
+    end
+  end
```
hooverlunch: A quick unit test for this validation would probably be a good idea. I don't want to get out of the habit of writing unit tests for this project. No need to test every supporting method, but this one would be good. Just testing the positive and negative cases.

app/models/project_step.rb
```diff
@@ -135,6 +138,27 @@ def handle_original_date_logic
     end
   end
 
+  def handle_finalized_at
+    if is_finalized && ! finalized_at
+      self.finalized_at = Time.now
+    elsif ! is_finalized && finalized_at
+      self.finalized_at = nil
+    end
+  end
```
hooverlunch: Also a test checking that this works as expected on save would be a good idea I think.

hooverlunch:
Looks great. Just a few changes and it's good for merging.

app/models/project_step.rb
elliottjf: Yeah.  I've added tests for the finalize handling as well as the automatic original_date updates.  (and I got tripped up yet one more time with the confusing `original_date` getter which masks an empty db value)

hooverlunch: Fuzzy said he's ok with refactoring that so go for it. I'd recommend a new
method like `best_known_date` or something that does the check and returns
the most up to date value.

On Fri, Apr 29, 2016 at 9:43 AM, Joseph Elliott <notifications@github.com>
wrote:

> In app/models/project_step.rb
> <https://github.com/sassafrastech/madeline_system/pull/69#discussion_r61579270>
> :
>
> > +  def handle_finalized_at
> > +    if is_finalized && ! finalized_at
> > +      self.finalized_at = Time.now
> > +    elsif ! is_finalized && finalized_at
> > +      self.finalized_at = nil
> > +    end
> > +  end
> > +
> > +  # Validates that a step may not be unfinalized more than 24 hours since it was previously marked
> > +  # as finalized.  Note, should generally be avoided by front-end logic, but guards against edge
> > +  # cases.
> > +  def unfinalize_allowed
> > +    if is_finalized_changed? && ! is_finalized && is_finalized_locked?
> > +      errors.add(:is_finalized, I18n.t("project_step.unfinalize_disallowed"))
> > +    end
> > +  end
>
> Yeah. I've added tests for the finalize handling as well as the automatic
> original_date updates. (and I got tripped up yet one more time with the
> confusing original_date getter which masks an empty db value)
>
> —
> <https://github.com/sassafrastech/madeline_system/pull/69/files/384c30beb7e505b87ce1fa30d40b434b91106b53#r61579270>
>



-- 
Tom Smyth

touchstonecohousing.org

PR #72, opened 2016-04-26
[WIP] 3772: Add loan tab view

emitche:


app/assets/javascripts/backbone/views/loan_tab_view.coffee
```diff
@@ -0,0 +1,13 @@
+class MS.Views.LoanTabView extends Backbone.View
+
+  events: (params) ->
```
hooverlunch: events should be a hash, not a function. So remove `(params) ->` from here and your event handler should start working.

app/assets/javascripts/backbone/views/loan_tab_view.coffee
```diff
@@ -0,0 +1,13 @@
...
+    'click .calendar-tab': 'openCalendar'
+
+  openCalendar: () ->
```
hooverlunch: You can remove `()`

app/views/admin/loans/show.html.slim
```diff
@@ -36,3 +36,8 @@ section.loan.details class=(@loan.valid? ? 'show-view' : 'edit-view')
 
   div.loan-content
     = render "admin/loans/tabs/tabs"
+
+javascript:
+  $(function(){
+    new MS.Views.LoanTabView();
```
hooverlunch: You need to set `el` or Backbone won't work. So `new MS.Views.LoanTabView({ el: 'something' });`

emitche:
Thanks for the help!

PR #75, opened 2016-04-29
4311: Adjust date format in steps and logs

emitche:
Adjust timeline to use localized dates


app/views/admin/logs/_log.html.slim
```diff
@@ -2,7 +2,7 @@
   .panel-heading
     .row
       .col-sm-10
-        b => log.date
+        b => log.date.try(:strftime, t('time.formats.date_only'))
```
hooverlunch: I believe the Rails way to do this is to use the `l` helper:

http://guides.rubyonrails.org/i18n.html#adding-date-time-formats

Instead of `try` you will need to use a conditional operator, like `log.date ? l(log.date) : ""`

hooverlunch:
Looks good. See note above.

Are we sure there aren't other places where dates need to be localized? What about in the coops and loans index tables?

I note that the `l` helper is already being used in one place, loans#index, line 45: `l(loan.SigningDate, format: :long) : ''` but it's not using one of the formats in the en.yml. Perhaps refactor that and do something similar in coop index?

hooverlunch:
Marking this revise and merge. If you are unsure about any of this, let me know.

emitche:
@hooverlunch Do you recommend using try in general for things that are not being translated? I prefer the `item ? item : ''` syntax, but it is a bit repetitive.

app/views/admin/loans/index.html.slim
```diff
@@ -24,7 +24,7 @@
 
   - g.column name: t("loan.sign_date"), attribute: 'signing_date',
     detach_with_id: :signing_date_filter do |loan|
-    - loan.signing_date.try(:strftime, t('time.formats.date_only'))
+    - loan.signing_date ? l(loan.signing_date) : ''
```
emitche: Can also do ` l(x) if x` here, but either fine.

app/views/admin/loans/content/_form.html.slim
```diff
@@ -14,15 +14,15 @@
   = f.input :organization_id, label: t("coop")
     .view-element
       = link_to admin_organization_path(@loan.organization_id) do
-        - @loan.organization.try(:name)
+        - @loan.organization.name ? @loan.organization.name : ''
```
emitche: Keep as try.

emitche:
Fixing an rspec test. Once the build runs successfully, will be merging this branch into develop.

PR #76, opened 2016-05-03
3777 manage divisions

elliottjf:
This isn't ready yet for a final review, but I think it's close.  Most important aspect that I'd like feedback on is the index policy handling.  It seems that the standard convention with Pundit is to pass class objects, not instances as the record object for `index` permission checks.  Tidying up the policy logic and controllers was easy enough, but there are deeper implications with current unit tests, so I've only partially addressed that so far.

One other issue that I wasn't able to resolve yet is getting a parent division filter to work on the index page, but I wanted to confirm if that was even desired before spending time on that.  (Or maybe there is something simple and obvious to fix)


hooverlunch:
Hopefully @meltheadorable can comment on the former issue. I would say leave out the latter as it is not in the spec I think and can be added later if desired.

app/policies/division_policy.rb
```diff
@@ -1,22 +1,34 @@
...
     division_member_or_admin(division: @record)
   end
 
   def create?
-    division_admin(division: @record)
+    # Need to let record with missing parent pass through the policy check so a validation
+    # message can be presented to the user.
+    # Todo: Confirm if there a better approach here?
+    return true unless @record.parent
+    division_admin(division: @record.parent)
```
hooverlunch: I would put this logic in the controller using skip_authorization or so. Putting it here risks side effects if this policy is used somewhere else.

app/policies/division_policy.rb
```diff
@@ -1,22 +1,34 @@
...
   def update?
-    division_admin(division: @record)
+    division_admin(division: @record) &&
+      !@record.root?  # Root division is considered read-only.
   end
 
+  # Note, for now we disallow deletion of divisions which have any organizations, loans, people,
+  # or child divisions.  Can change later if needed to allow the ability to delete with all
+  # dependencies.
+  # Todo: Confirm if destroy permission should be restricted to admins of the parent division.
```
hooverlunch: This sounds fine for now. Again we can tweak later if user needs dictate.

app/views/admin/divisions/show.html.slim
```diff
@@ -0,0 +1,30 @@
...
+      - if policy(@division).update?
+        - if @division.valid?
+          a.edit-action.view-element
+            i.fa.fa-pencil.fa-large>
+            = t(".edit")
+          a.show-action.form-element
+            = t("cancel_edit")
+        - else
+          a href=admin_division_path(@division) = t(:cancel)
+        / TODO: Confirm and cancel should have defaults
```
hooverlunch: Not sure what you mean here.

config/locales/en/divisions.en.yml
```diff
@@ -0,0 +1,13 @@
+en:
+  # Note, 'New Division' needs to be shared between the 'index' and 'new' actions, so can't use
+  # ".new" shortcut, and 'admin.divisions.new' seems too long.
+  # Todo, Confirm how exactly to name these kinds of shared keys.
```
hooverlunch: I would say `divisions.new_division` and `divisions.parent` are fine, just perhaps put a comment above the `admin` key saying that it's there because rails requires it for using the shortcut, so below are template-specific translations.

app/views/admin/divisions/show.html.slim
elliottjf: Neither do I.  This 'TODO' was copied from organizations/show.html.slim

app/controllers/admin/divisions_controller.rb
```diff
@@ -11,8 +12,84 @@ def select
...
+      per_page: 50
+    )
+    @parent_filter_choices = current_user.accessible_divisions.map{ |d| [d.name, d.id] }
+  end
+
+  # show view includes edit
+  def show
+    @division = Division.find(params[:id])
+    authorize @division
+    # Todo: Confirm best way to factor these view helper data structures.
```
elliottjf: Here's the context to our chat

app/controllers/admin/divisions_controller.rb
```diff
@@ -11,11 +12,101 @@ def select
...
+  def index
+    authorize Division
+    @divisions_grid = initialize_grid(
+      policy_scope(Division),
+      conditions: index_filter,
+      order: 'name',
+      per_page: 50
+    )
+    @parent_filter_choices = current_user.accessible_divisions.map{ |d| [d.name, d.id] }
+  end
```
hooverlunch: Please add the csv export boilerplate which is now in e.g. loans_controller and loans/index view. It's just a few lines here and a few in the template (grep for csv in the template to find the special stuff there).

app/controllers/admin/divisions_controller.rb
```diff
@@ -11,11 +12,101 @@ def select
...
+  def prep_form_vars
+    @currency_choices = Currency.all
+    @parent_choices = parent_choices(@division)
+  end
+
+  # List of other divisions which the current user has access to and are allowed to be assigned
+  # as a parent to this division.
+  def parent_choices(division)
+    (current_user.accessible_divisions - division.self_and_descendants) | [division.parent]
+  end
```
hooverlunch: Everything looks great here.

app/controllers/admin/organizations_controller.rb
```diff
@@ -1,6 +1,9 @@
 class Admin::OrganizationsController < Admin::AdminController
   def index
-    authorize Organization.new(division: current_division)
+    # Todo: Confirm that it makes more sense to authorize an index view against the class instead
+    # a transient instance
+    # authorize Organization.new(division: current_division)
+    authorize Organization
```
hooverlunch: I thought we settled this?

app/models/division.rb
```diff
@@ -26,27 +26,32 @@
...
 class Division < ActiveRecord::Base
   include CustomFieldAddable  # supports 'default_locales' persistence
-  has_closure_tree
+  has_closure_tree dependent: :restrict_with_exception
   resourcify
   alias_attribute :super_division, :parent
 
+  has_many :loans, dependent: :restrict_with_exception
+  has_many :people, dependent: :restrict_with_exception
+  has_many :organizations, dependent: :restrict_with_exception
```
hooverlunch: Nice use of restrict_with_exception

app/models/user.rb
```diff
@@ -69,4 +69,17 @@ def owning_division_id
     profile.try(:division_id)
   end
 
+  def has_some_access?
+    division_scope.base_accessible_ids.present?
+  end
+
+  # Require a user to have access to at least some division in order to login.
+  # Note, this avoids needing to worry about a nil current_division in the controller logic.
+  def active_for_authentication?
```
hooverlunch: This seems like a good idea.

app/policies/application_policy.rb
```diff
@@ -7,7 +7,9 @@ def initialize(user, record)
   end
 
   def index?
-    division_member_or_admin
+    # Rely on scope filtering and assume that users are allowed to view an index unless specifically
+    # disallowed.
+    true
```
hooverlunch: Nice comment.

app/policies/application_policy.rb
```diff
@@ -74,4 +76,10 @@ def check_self_and_ancestors_for_role(role:, division: @record.division)
 
     false
   end
+
+  # Require admin role on at least one division to allow access to index view for divisions or users
+  def any_division_admin?
+    user.roles.where(name: 'admin', resource_type: 'Division').present?
```
hooverlunch: `.any?` is technically more efficient here since it will run a COUNT query instead of loading all the role objects and then checking for `empty?`. (`present?` == `!(nil? || empty?)`)

app/views/admin/divisions/_form.html.slim
```diff
@@ -0,0 +1,41 @@
...
+    },
+    wrapper: :horizontal_form_condensed,
+    defaults: { \
+      input_html: {class: 'form-element form-control'},
+    },
+  ) do |f|
+
+  = f.error_notification
+
+  = f.input :name, label: t("name")
```
hooverlunch: You don't need to put these here explicitly. Just make sure `en.activerecord.attributes.division.name` is defined and it will happen automagically. Please refactor here and in organization.

spec/policies/division_policy_spec.rb
```diff
@@ -10,27 +10,41 @@
...
+    end
+    permit_actions [:show, :edit, :update]
+    forbid_actions [:create, :destroy]
   end
 
   context 'being a member of a parent division' do
     let(:user) { create(:user, :member, division: parent_division) }
 
-    permit_actions [:index, :show]
+    permit_actions [:show]
```
hooverlunch: Looks like the index permission is no longer being tested here?

spec/policies/division_policy_spec.rb
```diff
@@ -10,27 +10,41 @@
   context 'being a member of a division' do
     let(:user) { create(:user, :member, division: division) }
 
-    permit_actions [:index, :show]
+    it 'can not index' do
+      expect(described_class.new(user, Division).index?).to be_falsey
+    end
```
hooverlunch: Have you tried just putting this in `forbid_actions`? It should still work I think. Same below.

spec/policies/division_policy_spec.rb
```diff
@@ -10,27 +10,41 @@
...
-    permit_actions [:index, :show]
+    permit_actions [:show]
     forbid_actions [:create, :edit, :update, :destroy]
   end
 
   context 'being an admin of a parent division' do
     let(:user) { create(:user, :admin, division: parent_division) }
 
-    permit_all
+    permit_actions [:show, :create, :edit, :update, ]
```
hooverlunch: Please remove trailing comma.

spec/policies/division_policy_spec.rb
```diff
@@ -10,27 +10,41 @@
...
+    it 'can not delete division with a child' do
+      should forbid_action :destroy
+    end
+    it 'can delete division without children' do
+      expect(DivisionPolicy.new(user, child_division).destroy?).to be_truthy
+    end
```
hooverlunch: Looks like it makes sense to special case these.

spec/policies/division_policy_spec.rb
```diff
@@ -45,6 +59,8 @@
     forbid_all
   end
 
+  # Todo: Confirm business rules and implement tests for Division index permissions.
```
hooverlunch: Can remove this now?

spec/support/modules/base_policy.rb
```diff
@@ -51,16 +57,18 @@
...
 end
 
 def forbid_all
-  forbid_actions [:index, :create, :show, :edit, :update, :destroy]
+  forbid_actions [:create, :show, :edit, :update, :destroy]
 end
 
 def permit_all_but_destroy
-  permit_actions [:index, :create, :show, :edit, :update]
+  permit_actions [:create, :show, :edit, :update]
```
hooverlunch: Not sure why you're removing `:index` from here. Let's discuss.

hooverlunch:
Clicking the Root division link here throws an error:

image

hooverlunch:
I can't seem to sort only by parent division. If I click it, the name sort stays as well. This may predate this issue but can you look into it:

image

hooverlunch:
This looks fantastic. Small things above. We should briefly discuss what's going on in the base_policy file. But generally this is great work. Tagging it revise and merge. But let's discuss the base policy thing before you do so.

spec/policies/division_policy_spec.rb
elliottjf: The factored up `permit_actions` and `forbit_actions` tests are all testing against the shared `subject` which is built around an model instance.  The `index` action tests need to use a policy instance around the class.

hooverlunch: Actually I tested this myself and it worked. Can you just try it and see?

spec/policies/division_policy_spec.rb
elliottjf: I'm only testing the index permission for the two 'division' contexts.  Since the parent/child relations doesn't impact.  Would you want duplicate tests included for the other permutations anyways?

hooverlunch: This is fine.

On Fri, May 6, 2016 at 11:48 AM, Joseph Elliott <notifications@github.com>
wrote:

> In spec/policies/division_policy_spec.rb
> <https://github.com/sassafrastech/madeline_system/pull/76#discussion_r62348458>
> :
>
> >    end
> >
> >    context 'being a member of a parent division' do
> >      let(:user) { create(:user, :member, division: parent_division) }
> >
> > -    permit_actions [:index, :show]
> > +    permit_actions [:show]
>
> I'm only testing the index permission for the two 'division' contexts.
> Since the parent/child relations doesn't impact. Would you want duplicate
> tests included for the other permutations anyways?
>
> —
> <https://github.com/sassafrastech/madeline_system/pull/76/files/b996bf238caf158080fcee1cfa09858482b64ce3#r62348458>
>



-- 
Tom Smyth

touchstonecohousing.org

PR #77, opened 2016-05-03
3755 export table

emitche:
Allow export of loans and organizations index to CSV


app/views/admin/loans/_grid_content.html.slim
```diff
@@ -0,0 +1,47 @@
+/ - g.blank_slate do
```
emitche: Commenting this out is unnecessary.

app/views/admin/loans/_loans_grid.html.slim
```diff
@@ -0,0 +1,52 @@
+= define_grid(@loans_grid, show_filters: :always,
+    html: {class: 'rowlink table-hover'}) do |g|
+
+    / = render "grid_content", g: g
+
+    / - g.blank_slate do
```
emitche: Same as above.

emitche:
Visually, the index of organizations and loans may need some tweaks in terms of the filters and buttons. But CSV export functionally works in this feature.

app/controllers/admin/loans_controller.rb
```diff
@@ -10,8 +10,18 @@ def index
       order: 'loans.signing_date',
       order_direction: 'desc',
       custom_order: { 'loans.signing_date' => 'loans.signing_date IS NULL, loans.signing_date' },
-      per_page: 50
+      per_page: 50,
+      name: 'loans',
+      enable_export_to_csv: true,
+      csv_file_name: 'loans'
```
hooverlunch: I don't think this is needed unless it differs from `name`?

app/controllers/admin/loans_controller.rb
```diff
@@ -10,8 +10,18 @@ def index
...
     )
+
+    @csv_mode = true
+
+    export_grid_if_requested('loans' => 'loans_grid') do
```
hooverlunch: I don't think you need the argument here. I think loans -> loans_grid is the default. Let me know if I'm wrong. In general the Rails philosophy is convention over configuration so good to stick to that if possible.

hooverlunch:
Looks like filter bar no longer has gray bg. Was this on purpose?

hooverlunch:
Interesting issue when you try to open one of these in Excel:

image

It's because the first two chars of file are 'ID':

https://support.microsoft.com/en-us/kb/215591

hooverlunch:
Great work!!!

emitche:
It appears some stylesheets have been removed along the way. The index table's header with filters looks much better with a background.

PR #78, opened 2016-05-06
3774 click day new step

emitche:
Allow a project step to be created from a loan calendar day


app/assets/javascripts/backbone/views/calendar_step_modal_view.coffee
```diff
@@ -1,22 +1,32 @@
...
+      @showStep()
+    else
+      @loanId = params.loanId
+      @showNewStep()
 
   events:
-    'click .loan-calendar .cal-step': 'showStepModal'
-    'click a.action-delete': 'deleteStep'
+    'click a.action-delete': 'hideModal'
+    'click #new_project_step .cancel': 'hideModal'
```
hooverlunch: Let's stick to dashes for field IDs please.

app/assets/javascripts/backbone/views/calendar_step_modal_view.coffee
```diff
@@ -1,22 +1,32 @@
...
+
+  showStep: ->
+    $.get '/admin/project_steps/' + @id, context: @context, (html) =>
```
hooverlunch: Please use string interpolation in cases like these: `"/admin/project_steps/#{@id}"`

app/assets/javascripts/backbone/views/calendar_step_modal_view.coffee
```diff
@@ -1,22 +1,32 @@
...
-    $.get '/admin/project_steps/' + id, context: 'calendar', (html) =>
+  showNewStep: ->
+    $.get '/admin/project_steps/new', context: @context, loan_id: @loanId, (html) =>
       @replaceContent(html)
 
   replaceContent: (html) ->
-    @$('#calendar-step-modal').find('.modal-content').html(html)
-    @$('#calendar-step-modal').modal({show: true})
+    @$el.find('.modal-content').html(html)
+    @$el.modal({show: true})
```
hooverlunch: I believe this initializes the modal _and then_ shows it. Do you really want to initialize it each time this method is called? Should the initializer (just `.modal()`) go in `initialize` above, and just call `modal('show')` here?

app/assets/javascripts/backbone/views/calendar_view.js.coffee
```diff
@@ -4,27 +4,39 @@ class MS.Views.CalendarView extends Backbone.View
...
+      loading: @loading.bind(this)
       events: params.calendar_events_url
       customButtons:
         legend:
           text: 'Legend'
       header:
         left: 'prev,next today'
         center: 'title'
-        right: 'month,agendaWeek legend'
+        right: 'month,basicWeek legend'
```
hooverlunch: Nice fix. Does this fix the agendaWeek issue everywhere?

app/assets/javascripts/backbone/views/calendar_view.js.coffee
```diff
@@ -4,27 +4,39 @@ class MS.Views.CalendarView extends Backbone.View
...
     @renderLegend()
 
-  refresh: (e) ->
-    @$calendar.fullCalendar('refetchEvents')
+  events:
+    'click .loan-calendar .cal-step': 'showStepModal'
+
+  dayClick: ->
+    if (@$el.find('.loan-calendar'))
+      loanId = @$el.find('.loan-calendar').attr('data-loan-id').replace(/loan-/, '')
```
hooverlunch: Instead of `attr('data-loan-id')` you can do `data('loan-id')`. Always preferable to do that because if you set data attrib with `data()` it is only retrieveable with `data()`. `attr()` won't work.

app/assets/javascripts/backbone/views/calendar_view.js.coffee
```diff
@@ -37,8 +49,11 @@ class MS.Views.CalendarView extends Backbone.View
...
-  eventRender: (calEvent, element) ->
-    element.find('.fc-title').html(calEvent.title)
+  refresh: (e) ->
+    @$calendar.fullCalendar('refetchEvents')
 
-  loading: (isLoading) ->
-    MS.loadingIndicator[if isLoading then 'show' else 'hide']()
+  showStepModal: (e) ->
+    calStep = e.currentTarget
+    id = @$(calStep).attr('data-step-id').replace(/project_step-/, '')
```
hooverlunch: Same thing here.

app/assets/javascripts/backbone/views/project_step_view.coffee
```diff
@@ -9,6 +9,7 @@ class MS.Views.ProjectStepView extends Backbone.View
     @initTypeSelect()
     @persisted = params.persisted
     @duplicate = params.duplicate
+    @context = @$el.attr('data-context')
```
hooverlunch: And here.

app/assets/stylesheets/admin/calendar/_calendar.scss
```diff
@@ -90,7 +90,7 @@
   }
 
   .loan-calendar {
-    .cal-step {
+    .cal-step, td {
       &:hover {
```
hooverlunch: I don't think you need `&:hover` here. `cursor` means on hover by default, doesn't it?

app/views/admin/calendar/_calendar.slim
```diff
@@ -2,6 +2,6 @@ section.calendar
   h1.sr-only = t(:calendar)
   = render "admin/calendar/legend"
       
-  div#calendar.loan-calendar
+  div#calendar.loan-calendar data-loan-id="loan-#{loan.id}"
```
hooverlunch: Why do you need `loan-` as a prefix to the loan ID. Seems redundant? Can you remove, and then you wouldn't need `replace` in the JS?

app/assets/javascripts/backbone/views/calendar_step_modal_view.coffee
hooverlunch: Nevermind, I see what you're doing here. Cool!

app/assets/javascripts/backbone/views/calendar_view.js.coffee
```diff
@@ -4,27 +4,39 @@ class MS.Views.CalendarView extends Backbone.View
...
 
     @renderLegend()
 
-  refresh: (e) ->
-    @$calendar.fullCalendar('refetchEvents')
+  events:
+    'click .loan-calendar .cal-step': 'showStepModal'
+
+  dayClick: ->
+    if (@$el.find('.loan-calendar'))
```
hooverlunch: Outer parens not needed in Coffee

hooverlunch:
Looks like we are showing the newly created step in the modal after create. I think it would be better to just hide the modal so the user can watch the step appear in the calendar. That is confirmation enough I think.

hooverlunch:
The scheduled date should be pre-populated in the modal to whatever date was clicked.

image

Let's discuss how to achieve this. Working with dates can be tricky.

hooverlunch:
Ok I think that's it! This was very well done! Very impressive! And in a good amount of time too!

app/assets/javascripts/backbone/views/calendar_view.js.coffee
emitche: Yes. This is for all calendars.

app/views/admin/calendar/_calendar.slim
emitche: I would prefer not to have. Was part of other work. Will remove.

app/assets/javascripts/backbone/views/calendar_step_modal_view.coffee
emitche: I'm not sure how the modal id is being added/populated. This is the id on the modal. I did not see a partial for this modal.

emitche: Nevermind. `.cancel` and/or other selectors are plenty and that id does not need to be used in the JS.

app/assets/javascripts/backbone/views/calendar_step_modal_view.coffee
emitche: `('show')` is shorthand. I changes it to that, as I think that is best.

hooverlunch: Are you sure? Can you link to where it says that in the docs? I am
confused...

On Mon, May 16, 2016 at 11:58 AM, Elizabeth Mitchell <
notifications@github.com> wrote:

> In app/assets/javascripts/backbone/views/calendar_step_modal_view.coffee
> <https://github.com/sassafrastech/madeline_system/pull/78#discussion_r63377119>
> :
>
> >        @replaceContent(html)
> >
> >    replaceContent: (html) ->
> > -    @$('#calendar-step-modal').find('.modal-content').html(html)
> > -    @$('#calendar-step-modal').modal({show: true})
> > +    @$el.find('.modal-content').html(html)
> > +    @$el.modal({show: true})
>
> ('show') is shorthand. I changes it to that, as I think that is best.
>
> —
> <https://github.com/sassafrastech/madeline_system/pull/78/files/3624899d75a6267d9e09080d65279d58710f6b38#r63377119>
>



-- 
Tom Smyth

touchstonecohousing.org

emitche:
@hooverlunch: I went ahead and incorporated your suggestions, as they were good ones. Thanks so much.

PR #83, opened 2016-05-18
4513 modal titles

emitche:
Refine modal titles in project step and log


hooverlunch:
Looks great, good catch on using the one/other stuff!

PR #87, opened 2016-05-25
new loan links and loan relation field access control

elliottjf:
@hooverlunch  Two things that I wanted to get input on.  I implemented read-only/edit behavior as we discussed when editing a loan with references to people or orgs for which the current user does not have access.  I did not see a clean way to factor out the pattern though, perhaps you have an idea.

For the links to a new loan and all loans from the co-op detail page, I've provisionally used inline styles to get the layout approximately along the lines of the mockup.  I wanted to confirm if the layout approach (i.e. span with a float: right) is reasonable and how specifically these styles should be factored into the .scss files.


app/views/admin/loans/content/_form.html.slim
```diff
@@ -10,18 +10,29 @@
...
-    = f.input :division
-      .view-element
-        - if @loan.division
+  = f.input :division
+    .view-element
+       - if @loan.division && division_policy(@loan.division).show?
           = link_to @loan.division.name, admin_division_path(@loan.division_id)
-      = f.input_field :division_id, collection: @division_choices, include_blank: false
+       - else
+        = @loan.division.try(:name)
```
hooverlunch: Indent is not quite right above. I think if/else are one space too far in. Please take a look.

app/views/admin/loans/content/_form.html.slim
```diff
@@ -10,18 +10,29 @@
...
+    .form-element
+      - if !@loan.division || division_policy(@loan.division).show?
+        = f.input_field :division_id, collection: @division_choices, include_blank: false
+      - else
+        = @loan.division.try(:name)
```
hooverlunch: Don't need `try` here b/c you're checking for presence above.

app/views/admin/loans/content/_form.html.slim
```diff
@@ -10,18 +10,29 @@
...
-    = f.input_field :organization_id, collection: @organizations, include_blank: true
+      - if @loan.organization && organization_policy(@loan.organization).show?
+        = link_to @loan.organization.name, admin_organization_path(@loan.organization_id)
+      - else
+        = @loan.organization.try(:name)
+    .form-element
+      - if !@loan.organization || organization_policy(@loan.organization).show?
+        = f.input_field :organization_id, collection: @organization_choices, include_blank: true
+      - else
+        = @loan.organization.try(:name)
```
hooverlunch: Same here.

app/views/admin/loans/content/_form.html.slim
```diff
@@ -48,44 +59,68 @@
...
+    .view-element
+      - if @loan.representative && person_policy(@loan.representative).show?
+        = link_to @loan.representative.name, admin_person_path(@loan.representative)
+      - else
+        = @loan.representative.try(:name)
+    .form-element
+      - if !@loan.representative || person_policy(@loan.representative).show?
+        = f.input_field :representative_id, collection: @representative_choices, include_blank: true
+      - else
+        = @loan.representative.try(:name)
```
hooverlunch: You asked about abstracting this out... I would use a helper method called e.g. `authorized_form_field`. You only need 3 parameters: the object, the field name (e.g. representative_id), and the choice collection. The policy class can be guessed from the object class, and the link path can be obtained with `url_for(object)`. Make sense?

app/views/admin/organizations/_loans.html.slim
```diff
@@ -1,4 +1,11 @@
-h1 = t('organization.active_loans')
+/ JE Todo: Confirm how these styles should be factored
+div style="margin-bottom: 15px"
+  h1 style="display: inline"
+    = t('organization.active_loans')
+  span style="margin-left: 10px"
+    = link_to t(".view_all_loans"), admin_loans_path({'loans[f][organizations.name][]' => @org.name})
+  span style="float: right"
+    = link_to t("loan.shared.new"), new_admin_loan_path({organization_id: @org.id}), class: 'btn btn-default new-button'
```
hooverlunch: Don't need curly braces if the hash is the only or last param to the method. Please fix above.

hooverlunch: Going to get input from Liz in a min on how to handle styles. Please stay tuned.

hooverlunch: I would make a style inside the .block style in general called e.g. .header-with-links. Then you can put your margin in there, and then descendant h1's can be made inline-block, any link can have a margin-left (no need for a wrapping span, links should also be made inline-block), and any button can be floated right. That should get rid of all the inline stuff.

hooverlunch:
Nice work! Ready to merge after the above changes. Thank you.

app/views/admin/loans/content/_form.html.slim
elliottjf: I was thinking to not make the display code dependent upon that assumption in case the 'may_edit?' logic changes in the future.  I've removed the try, but added comments warning of the dependency as the with the refactoring out to helper method and partial as further separated the logic and dependency.

app/views/admin/loans/content/_form.html.slim
elliottjf: Okay.  the `url_for` method is handy.  I'd forgot that was already provided. I'm also passing along simple form `f` object, so my helper method has 4 params.  And I've broken a partial which the helper method renders along with several calculated local params.  Give that we have 4 param, do you want me to use named parameters?  If we do, it will make all of these method calls two line instead of a single line.

hooverlunch: Probably a good idea.

PR #90, opened 2016-05-31
4574 move export to csv button

emitche:
Move export to CSV below index table

### DIFF
--- a/app/views/admin/divisions/index.html.slim
+++ b/app/views/admin/divisions/index.html.slim
@@ -29,5 +29,5 @@
 = render_grid(@divisions_grid)
 
 .text-right
-    button.btn.btn-default.wg-external-csv-export-button data-grid-name="divisions"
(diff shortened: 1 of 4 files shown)


emitche:
@rgalusca: I moved the buttons. They are styled as links but act like buttons. The visual difference is that when they are clicked or are in focus, they have a border until you interact with another element on the page. If this is a problem, let me know and I will make into a link. Links would require custom code, but buttons behave as expected out the box.

app/views/admin/divisions/index.html.slim
```diff
@@ -24,9 +24,10 @@
...
 
-        button.btn.btn-default.wg-external-csv-export-button data-grid-name="divisions"
-          = t("export_to_csv")
-
       .clearfix
 
 = render_grid(@divisions_grid)
+
+.text-right
+    button.btn.btn-link.wg-external-csv-export-button data-grid-name="divisions"
```
hooverlunch: Looks like this is double indented.

PR #91, opened 2016-05-31
4515: Export visible columns in divisions index to CSV

emitche:
Ensure CSV tables reflect visible columns in index tables


hooverlunch:
The idea was to export /all/ columns, not just visible columns, to the CSV export only. So we will need to add a bunch more here. Just look in the schema.rb file to see what columns need to be added.

I don't need to see the code again, it should be simple, but please do test each export with some real data (@rgalusca has a DB dump if you need one) after the changes are made to make sure they still work. Then this can be merged.

emitche:
Alright. I will check in with @rgalusca about this and get data. Thanks a bunch.

PR #92, opened 2016-05-31
4578 bug loans index no records

emitche:
Show message when filters return no records in grid

### DIFF
--- a/app/views/admin/divisions/_divisions_grid.html.slim
+++ b/app/views/admin/divisions/_divisions_grid.html.slim
@@ -3,7 +3,7 @@
 
   - g.blank_slate do
     - @empty_list = true
-    - content_tag(:div, t("no_records"), class: 'alert alert-info')
+    - content_tag(:div, t('common.no_records'), class: 'alert alert-info')
 
   - g.column name: t("id"), attribute: 'id', filter: false, in_csv: false do |division|
     - link_to(division.id, admin_division_path(division.id))
--- a/app/views/admin/divisions/index.html.slim
+++ b/app/views/admin/divisions/index.html.slim
@@ -6,6 +6,7 @@
 = render 'layouts/admin/tables'
 
 = render "divisions_grid"
+- @no_records = @divisions_grid.current_page_records.length < 1
 
 - unless @empty_list
   header.block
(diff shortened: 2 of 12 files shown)


app/views/admin/divisions/_divisions_grid.html.slim
```diff
@@ -3,7 +3,7 @@
 
   - g.blank_slate do
     - @empty_list = true
```
emitche: Empty list only happens when on page load there are no records returned from the initial filters (usually all records found).

app/views/admin/divisions/index.html.slim
```diff
@@ -6,6 +6,7 @@
 = render 'layouts/admin/tables'
 
 = render "divisions_grid"
+- @no_records = @divisions_grid.current_page_records.length < 1
```
emitche: No records means that no records match a specific set of filters.

Wice grid does not have a built-in way to display to the user in a more UX-friendly fashion that no records match the filters used. So this variable helps with that.

Per the Wice grid documentation, the method current_page_records *must* be used after the divisions grid. It could be included later in the page, but cannot be included in the controller because `define_grid(@divisions_grid, ...)` is defined in the view.

app/views/admin/divisions/index.html.slim
```diff
@@ -29,4 +30,6 @@
 
       .clearfix
 
-= render_grid(@divisions_grid)
+.index-grid class="#{'no-records' if @no_records}"
+  = render_grid(@divisions_grid)
```
emitche: Clearing out the filters does not work unless the grid is loaded on the page. Therefore, the grid is hidden on the page but in the DOM.

config/locales/en/en.yml
```diff
@@ -219,6 +218,8 @@ en:
     description: "Description"
     division: "Division"
     divisions: "Divisions"
+    no_records: "No records found."
+    no_records_match: "No records match. Adjust the filters or use Clear to reset filters."
```
emitche: This message can be altered as needed.

app/views/admin/divisions/index.html.slim
```diff
@@ -29,4 +30,6 @@
...
+  = content_tag(:div, t('common.no_records_match'), class: 'alert alert-info no-records-match')
```
hooverlunch: This seems like a fine approach but can we move it into a helper method to avoid duplication?

hooverlunch:
Looks great. Ready to go after the refactor we discussed.

PR #93, opened 2016-06-03
4574: Hide export to CSV when no records

emitche:
- Move export to CSV to grid partial
- Remove empty space around button due to hidden note


app/helpers/application_helper.rb
```diff
@@ -47,10 +47,6 @@ def person_policy(record)
 
   def render_index_grid(grid)
     no_records = grid.current_page_records.length < 1
-    content_tag(:div, class: "index-grid #{'no-records' if no_records}") do
```
emitche: I really like using this way of loading content, but more content needed to be added and a partial felt more appropriate in this context.

PR #94, opened 2016-06-08
4609: Empty date click opens new step on loan calendar only

emitche:
Fixes loading indicator showing when global calendar's empty dates are clicked


emitche:
As this is a tiny change, I went ahead and merged it in.

PR #96, opened 2016-06-10
4518: Fix shift dates modal being opened during wrong actions

emitche:
Fix shift dates modal displaying when delete and duplicate step modals are displayed

- Move JS for shift dates into step view model
- Change class that opened modal when wrong action called


app/views/admin/project_steps/_shift_subsequent_confirm_modal.html.slim
```diff
@@ -1,4 +1,4 @@
-div.modal.fade.duplicate-step [id="step-shift-subsequent-confirm-modal-#{step.id}" tabindex="-1"
```
emitche: This is the underlying reason for the bug.

emitche:
The underlying reason for this bug appears to be a class on the modal that applied to another action. However, I reorganized the logic for shift dates as I was expecting to find it inside the view model and not inside the shift dates modal partial.

hooverlunch:
Looks good. The reorganization is fine. I think this stuff might be subsumed by my work on the drag event issue, but fine to merge for now.

PR #101, opened 2016-06-10
4605 bug fix double log button

emitche:
- Remove add log button at bottom of logs list
- Ensure last deleted log returns view to no logs view

### DIFF
--- a/app/assets/javascripts/backbone/views/project_step_view.coffee
+++ b/app/assets/javascripts/backbone/views/project_step_view.coffee
@@ -93,18 +93,13 @@ class MS.Views.ProjectStepView extends Backbone.View
       @modalView.showNew(@$(link).data('parent-step-id'))
 
   deleteLog: (e, response) ->
-    $.post @$(e.target).attr('href'), {_method: 'DELETE'}
-
-    # Number of logs currently visible includes one just deleted from database
-    logsList = @$(e.target).closest('.logs-list')
-    numOfLogs = @$(logsList).children().length
-
-    if numOfLogs < 2
-      logContainer = @$(logsList).closest('.step-logs')
-      @$(logContainer).removeClass('expanded')
-      @$(logContainer).find('.view-logs-ctrl').html('<span class="no-logs">No Logs</span>')
-
-    @$(e.target).closest('.log').remove()
+    self = this
+    $.post(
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/backbone/views/project_step_view.coffee
```diff
@@ -91,5 +91,15 @@ class MS.Views.ProjectStepView extends Backbone.View
 
   deleteLog: (e, response) ->
     $.post @$(e.target).attr('href'), {_method: 'DELETE'}
+
+    # Number of logs currently visible includes one just deleted from database
+    logsList = @$(e.target).closest('.logs-list')
+    numOfLogs = @$(logsList).children().length
```
hooverlunch: Please combine the two above lines into one. It's ok to chain a bunch of jquery methods in a row (it's designed for this), as long as readability doesn't suffer.

`@$(e.target).closest('.logs-list').children().length` is quite clear.

app/assets/javascripts/backbone/views/project_step_view.coffee
```diff
@@ -91,5 +91,15 @@ class MS.Views.ProjectStepView extends Backbone.View
...
+
+    if numOfLogs < 2
+      logContainer = @$(logsList).closest('.step-logs')
+      @$(logContainer).removeClass('expanded')
+      @$(logContainer).find('.view-logs-ctrl').html('<span class="no-logs">No Logs</span>')
```
hooverlunch: Better to show/hide the existing than to insert one like this. The HTML you're inserting is not translated, for one. This is why it's best to keep HTML out of JS if at all possible.

emitche:
Best to rerender entire logs partial instead.

emitche:
The way this is currently set up, if the additional add log button was conditional on number of logs present, it would be easy to implement through a variable in the controller.

hooverlunch:
GREAT WORK. Much better!

I made one commit simplifying the coffeescript. Notice that you don't need to list out all JS method params if you aren't using them. You can cut off the unused ones on the end.

emitche:
Smart. Thanks!

hooverlunch:
Going to merge this now as I don't want conflict in other issue I'm working on :)

PR #104, opened 2016-06-15
4515 csv columns

emitche:
Add two missing columns into CSV export


emitche:
Addresses two missing columns. An additional column missing needs feedback first before being addressed. This either can be reviewed and merged in or can wait for feedback.

PR #105, opened 2016-06-15
4644: Bug fix: Ensure new steps can be added

emitche:
- Applies to timeline and calendar


app/views/admin/project_steps/_project_step.html.slim
```diff
@@ -21,13 +21,14 @@ div.panel.step [
     = render "admin/project_steps/shift_subsequent_confirm_modal", step: step,
       days_shifted: days_shifted, subsequent_count: subsequent_count
 
+  - stepId = step.id ? step.id : 0
```
emitche: @hooverlunch: How would you approach this differently? The project step view model expects a stepId. Is there some temp step id or something else that can be used here?

hooverlunch:
I fixed this as part of an issue I'm working on, so going to close.

PR #109, opened 2016-06-29
[WIP] 3759 criteria print

emitche:
- Beginning work on integrating loan criteria into loan print memo


emitche:
This work will end up in its parent branch, 3759_print_loan_memo.

PR #110, opened 2016-06-30
3864 manage loan questions

hooverlunch:


app/assets/javascripts/backbone/views/auto_loading_indicator_view.coffee
```diff
@@ -0,0 +1,10 @@
+class MS.Views.AutoLoadingIndicatorView extends Backbone.View
+  # Instantiate this class in a view to make the loading indicator show and hide
+  # automatically with every ajax request
+
+  initialize: ->
+    $(document)
+      .ajaxStart ->
+        MS.loadingIndicator.show()
+      .ajaxStop ->
+        MS.loadingIndicator.hide()
```
hooverlunch: Nice

app/assets/javascripts/backbone/views/filter_switch_view.coffee
```diff
@@ -0,0 +1,36 @@
...
+    $('.filter-switch .btn').removeClass('active')
+    $(".filter-switch input[value='#{selected}']").closest('.btn').addClass('active')
+    @doFilter(selected)
+
+  doFilter: (selected) ->
+    if selected == 'all'
+      $('.filterable').show()
+    else
+      $('.filterable').hide()
+      $(".filterable.#{selected}").show()
```
hooverlunch: Yeah I could have used this.. should refactor my stuff later

app/assets/javascripts/backbone/views/loan_questions_view.coffee
```diff
@@ -0,0 +1,127 @@
+class MS.Views.LoanQuestionsView extends Backbone.View
```
hooverlunch: Should probably call this ManageLoanQuestionsView since there is also LoanQuestionnairesView

app/assets/javascripts/backbone/views/log_modal_view.coffee
```diff
@@ -23,7 +23,7 @@ class MS.Views.LogModalView extends Backbone.View
 
   replaceContent: (html) ->
     @$el.html(html)
-    new MS.Views.TranslationsView(el: @$('[data-content-translatable="log"]'))
+    new MS.Views.TranslationsView()
```
hooverlunch: Not sure if this is a good idea since by hard-coding the scope into TranslationsView, you are (I think) making it impossible to have two translatable field areas on the same page, which seems like it could be a possibility.

app/assets/stylesheets/admin/_bootstrap_layout.scss
```diff
@@ -34,7 +34,8 @@ body {
 
 .navbar-fixed-top .navbar-collapse,
 .navbar-fixed-bottom .navbar-collapse {
-  max-height: none;
+  // This was making things get cut off on Chrome's emulator without being able to scroll
+  // max-height: none;
```
hooverlunch: Nice comment

app/assets/stylesheets/admin/loan_questions.scss
```diff
@@ -0,0 +1,45 @@
+.loan-questions {
```
hooverlunch: Can we call all your stuff 'manage-loan-questions'. I worry about people getting confused between this and mine which is called 'questionnaire(s)'.

app/controllers/admin/divisions_controller.rb
```diff
@@ -1,8 +1,4 @@
 class Admin::DivisionsController < Admin::AdminController
-
-  before_action :authenticate_user!
-  after_action :verify_authorized
-
```
hooverlunch: Nice catch to remove this.

app/assets/stylesheets/admin/loan_questions.scss
hooverlunch: Erm let's discuss this, complicated...

app/controllers/admin/loan_questions_controller.rb
```diff
@@ -0,0 +1,78 @@
...
+    target = CustomField.find(params[:target])
+    method = case params[:relation]
+      when 'before' then :prepend_sibling
+      when 'after' then :append_sibling
+      when 'inside' then :prepend_child
+    end
+
+    target.send(method, @loan_question)
+    head :no_content
+  rescue
```
hooverlunch: Are we expecting any errors here? If not I'd rather let them happen and result in 500s so that we can notice and fix them instead of handling them gracefully and presenting what is probably a cryptic error message like "No method + for nil" to the user. So I would just drop the rescue block here and in destroy.

app/helpers/translations_helper.rb
```diff
@@ -27,3 +27,34 @@ def translate_boolean(bool)
...
+          <div class="language-block" data-locale="#{l}">
+            <a class="remove-language" href="#">#{I18n.t('common.remove')}</a>
+        }.html_safe
+        out += input :"locale_#{l}", label: false,
+          collection: locale_options,
+          input_html: {class: 'locale'}, include_blank: false
+
+        out += template.capture { yield l }
+
+        out += "</div>".html_safe
```
hooverlunch: This is a big no-no I think because there can be arbitrary content inside the input tag, so you shouldn't be using html_safe. You shouldn't need it if you use `content_tag` and such. Let's discuss.

app/helpers/translations_helper.rb
```diff
@@ -27,3 +27,34 @@ def translate_boolean(bool)
     t(bool ? "reply_yes" : "reply_no")
   end
 end
+
+module SimpleForm
+  class FormBuilder
+    include TranslationsHelper
+
+    def translatable_inputs(&block)
```
hooverlunch: Great to add this!

app/helpers/translations_helper.rb
```diff
@@ -27,3 +27,34 @@ def translate_boolean(bool)
     t(bool ? "reply_yes" : "reply_no")
   end
 end
+
+module SimpleForm
```
hooverlunch: Put in own file?

app/serializers/custom_field_serializer.rb
```diff
@@ -0,0 +1,23 @@
+class CustomFieldSerializer < ActiveModel::Serializer
```
hooverlunch: Nice.

app/views/admin/media/_translatable.html.slim
```diff
@@ -1,14 +1,5 @@
...
-        = f.input :"caption_#{l}",
-          input_html: {data: {translatable: 'common.summary'}},
-          placeholder: I18n.t('common.caption', locale: l),
-          label: I18n.t('common.caption', locale: l)
-  a.add-language href='#' = t('common.add_language')
+= f.translatable_inputs do |l|
+  = f.input :"caption_#{l}",
+    input_html: {data: {translatable: 'common.summary'}},
+    placeholder: I18n.t('common.caption', locale: l),
+    label: I18n.t('common.caption', locale: l)
```
hooverlunch: mmm, clean!

config/locales/en/loan_questions.en.yml
```diff
@@ -0,0 +1,38 @@
...
+      custom_field:
+        data_type:
+          string: String
+          text: Text
+          number: Number
+          range: Range
+          group: Group
+          boolean: Boolean
+          translatable: Translatable
+          list: List
```
hooverlunch: Very nice, clean.

app/controllers/admin/loan_questions_controller.rb
linesarefuzzy: I don't understand your logic. If we let an error happen, since this is an ajax request, it will look to the user like it worked, until they reload the page. Or maybe it will hang with the loading indicator. This way they know it failed, they have a clue as to why, and they can copy and paste the error message when they tell us about it. Why would it be better to let the error happen?

hooverlunch: The ajax request would still receive a 500 response and the page should then say 'server error'. If it's not a known error, that's all we should be reporting to the user. They can then tell us it errored, and we can look into it. When was the last time you saw an error page on Google that included technical stuff like `nil` this or that? They just say 'server error'. I feel pretty strongly about this one.

hooverlunch: Note there is a global error modal for this in there somewhere I believe. You could add that to your global loading indicator code.

PR #112, opened 2016-07-01
4748 language button

emitche:


app/helpers/translations_helper.rb
```diff
@@ -40,7 +40,7 @@ def translatable_inputs(&block)
       }.html_safe
       object.used_locales_or_current_locale.each do |l|
         out += %Q{
-          <div class="language-block" data-locale="#{l}">
+          <div class="language-block row" data-locale="#{l}">
```
emitche: This feel more like a temp fix for this issue. Something is wrong with the language blocks.

PR #113, opened 2016-07-14
4736 reformat questionnaire

linesarefuzzy:


app/controllers/admin/custom_value_sets_controller.rb
```diff
@@ -2,21 +2,21 @@ class Admin::CustomValueSetsController < Admin::AdminController
   def create
     @record = CustomValueSet.new(record_params)
     authorize @record
-    @record.save
+    @record.save!
```
linesarefuzzy: There being no error handling here, I thought it better to raise an exception rather than return a success message.

hooverlunch:
Looks like there is a CI failure. Can you please fix?

app/controllers/admin/custom_value_sets_controller.rb
hooverlunch: Yes! Always a good idea.

app/assets/javascripts/backbone/views/loan_questionnaires_view.coffee
```diff
@@ -5,37 +5,34 @@ class MS.Views.LoanQuestionnairesView extends Backbone.View
...
   refreshContent: ->
     MS.loadingIndicator.show()
-    @$('.questionnaires-content').empty()
-    $.get "/admin/loans/#{@loanId}/questionnaires", (html) =>
+    @$('.questionnaires-content').load "/admin/loans/#{@loanId}/questionnaires", =>
       MS.loadingIndicator.hide()
-      @$('.questionnaires-content').html(html)
-      @initTopButtons()
+      @initializeTree()
+      @filterSwitchView.filterInit()
```
hooverlunch: Thanks for changing to use your View.

app/assets/javascripts/backbone/views/loan_questionnaires_view.coffee
```diff
@@ -5,37 +5,34 @@ class MS.Views.LoanQuestionnairesView extends Backbone.View
...
+      selectable: false
+      useContextMenu: false
+      onCreateLi: (node, $li) =>
+        $li.attr('data-id', node.id)
+            .find('.jqtree-element')
+            .append(@$(".question[data-id=#{node.id}] > .explanation"))
+            .append(@$(".question[data-id=#{node.id}] > .answer-wrapper"))
+
+    tree.each ->
+      $(this).tree 'loadData', $(this).data('data')
```
hooverlunch: I think some quick comments on the above `onCreateLi` and `tree.each` are in order. It is not clear what is going on.

app/assets/stylesheets/admin/tree.scss
```diff
@@ -0,0 +1,59 @@
...
+  display: block;
+
+  .explanation, .answer, .form-element {
+    margin: 10px 0 0 0;
+  }
+
+  .form-element {
+    display: block;
+  }
+}
```
hooverlunch: I feel like these styles don't really belong in tree.scss. Can they be in their own files (one for loan-questions and one for questionnaires)?

app/models/custom_field.rb
```diff
@@ -47,6 +47,13 @@ class CustomField < ActiveRecord::Base
 
   DATA_TYPES = %i(string text number range group boolean)
 
+  def self.loan_questions(field_set = nil)
+    field_set &&= "loan_#{field_set}"
+    field_set ||= ['loan_criteria', 'loan_post_analysis']
```
hooverlunch: I admit I am stumped here. If the data type of field_set is string, then this seems like a way-too-cutsie way to do things. If it's not, then I'm even more stumped. I would consider refactoring for clarity's sake.

hooverlunch:
This looks great codewise. Some good refactoring and good use of abstraction. No major issues that I can see, just a few small tidies above. Great work!

app/models/custom_field.rb
linesarefuzzy: Yeah, it's a little cutesie. `field_set` is a string, either 'criteria' or 'post_analysis', or nil. If it's given, it needs to be prepended, and if it's not, it needs to be set to both. What if I just add this as a comment?

linesarefuzzy:
Thanks!

PR #115, opened 2016-07-19
4737-  question groups with loan types

elliottjf:
I think this is basically functional.  I still need to implement some unit tests


app/models/custom_field.rb
```diff
@@ -56,6 +69,24 @@ def self.loan_questions(field_set = nil)
...
+  #   and optional for all others
+  # - If override is true and no records are present, all are optional
+  # - If override is false, inherit from parent
+  # Note, loan type association records are ignored for questions without the 'override_assocations'
+  # flag assigned.
+  def required_for?(loan)
+    if override_associations
+      options.include?(loan.loan_type_option)
+    else
+      parent && parent.required_for(loan)
```
linesarefuzzy: Missing `?` here?

app/models/custom_field.rb
```diff
@@ -2,17 +2,19 @@
...
-#  required             :boolean          default(FALSE), not null
-#  updated_at           :datetime         not null
+#  created_at            :datetime         not null
+#  custom_field_set_id   :integer
+#  data_type             :string
+#  has_embeddable_media  :boolean          default(FALSE), not null
+#  id                    :integer          not null, primary key
+#  internal_name         :string
+#  migration_position    :integer
+#  overridden_id         :integer
```
linesarefuzzy: Remove from db?

app/models/custom_field.rb
```diff
@@ -23,11 +25,22 @@
...
 class CustomField < ActiveRecord::Base
   include Translatable
 
   belongs_to :custom_field_set
 
+  # Used for Questions(CustomField) to LoanTypes(Options) associations which imply a required
+  # question for a given loan type.
+  #has_and_belongs_to_many :options
+  has_many :custom_field_requirements, dependent: :destroy
+  has_many :options, through: :custom_field_requirements
```
linesarefuzzy: Maybe we should call this `loan_types`

elliottjf: How about adding as an alias?
`alias_method :loan_types, :options`

linesarefuzzy:
I reviewed this and am approving it since Tom is out right now and other work is waiting on it. Great work Joseph!

PR #116, opened 2016-07-21
4825: Fix Sass error

emitche:
- A sass-rails bug is not loading in Sass variables as expected.
- This is a work around.


emitche:
This sass-rails bug is likely the reason this is not working: https://github.com/rails/sass-rails/issues/377. Other people are having similar problems.

PR #118, opened 2016-07-22
4744 spreadsheets frontend

emitche:
Add styling to add and edit spreadsheets form


emitche:
These are basic styles for this feature.

PR #122, opened 2016-08-10
4942 attached spreadsheets styling

emitche:
- Style attach spreadsheet form
- Ensure attached spreadsheet has auto adjusted height


db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
```
emitche: I am not sure if this should be included, but it was its own commit, just in case.

emitche:
Closing. This branch should be evaluated against develop.

PR #123, opened 2016-08-11
4843 questions per loan type migration

elliottjf:
Note, that I've rolled back the inclusion of the 'due diligence' question groups into the 'criteria' question set.  Things should be functional now with the 'amount' field now included in the per loan type question associations, but we'll still need a discussion around how to handle the due diligence stuff.


linesarefuzzy:
Looks great. Thanks.

PR #125, opened 2016-08-12
4757 - loan question set schema and migration updates

elliottjf:


app/models/legacy/loan_question.rb
```diff
@@ -7,6 +7,25 @@ class LoanQuestion < ActiveRecord::Base
...
+
+      # filter out questions which seemed to be ignored by old system.
+      connection.execute("UPDATE LoanQuestions set Active = 0 where NewOrder = 0")
+      connection.execute("UPDATE LoanQuestions set Active = 0 where NewGroup is null and Type != 'Grupo'")
+      # ignore some random cruft from lastest prod db
+      connection.execute("UPDATE LoanQuestions set Active = 0 where Type = ''")
+
+      # create some test data for the 'inactive' state
+      connection.execute("update LoanQuestions set NewGroup = Grupo where active = 2 and id > 100")
+    end
```
linesarefuzzy: The first two statements here are all I did on the production db. These other changes seem like they might mess up the old system. Can they be made in the migration?

linesarefuzzy:
Other than that one comment, this looks fine. Thanks!

app/models/legacy/loan_question.rb
elliottjf: I was working on a clone of the db, so I wasn't thinking from the context playing nicely with the existing production system.  I was actually expecting those additional changes to be performed manually on production.  I'll just comment them out from what gets committed.  They're handy though to be able to temporarily uncomment to cleanup recloned production data.

PR #129, opened 2016-08-31
WIP: 4993 charts loan memo

emitche:
- Add pie charts for breakeven financial model


emitche:
Waiting on some feedback from @rgalusca regarding design before proceeding, but remaining is visual/contextual tweaks.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -0,0 +1,166 @@
...
+      name = product.name
+      total = product.total
+      rows.push({"c":[{"v": name, "f":null},{"v": total, "f":null}]})
+
+    chartData = {"cols": columns, "rows": rows}
+    chartData = new google.visualization.DataTable(chartData);
+    chart = new google.visualization.PieChart(document.getElementById('breakeven-revenue-chart'));
+    chart.draw(chartData, options);
+
+  breakevenTotalCostsChart: () ->
```
emitche: I did not have time to change this chart over to the desired chart that shows total costs only rather than the fixed vs. product breakdown. The colors have been changed to match the hue of the default Google Charts colors. Those two base colors (blue and red) match the Google Charts hue.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -0,0 +1,166 @@
...
+    columns = [
+      {"id":"","label":I18n.t('loan.breakeven.fixed_costs', count: 1),"pattern":"","type":"string"},
+      {"id":"","label":I18n.t('loan.breakeven.amount'),"pattern":"","type":"number"}
+    ]
+
+    rows = []
+    for key,cost of @breakevenFixedCosts
+      name = cost.name
+      total = cost.amount
+      rows.push({"c":[{"v": name, "f":null},{"v": total, "f":null}]})
```
emitche: I was not able to convert the displayed amount, which is f, to use $ because of time constraints.

app/views/admin/loans/content/_breakeven_table.html.slim
```diff
@@ -54,7 +54,7 @@
 
       / Fixed Costs
       tr.group-title
-        td colspan="4" = t("loan.breakeven.fixed_costs")
+        td colspan="4" = t("loan.breakeven.fixed_costs", count: 2)
```
linesarefuzzy: What is this `count: 2` here for?

app/views/admin/loans/content/_breakeven_charts.html.slim
```diff
@@ -0,0 +1,24 @@
+.summary-chart.breakeven-chart.row
+  h3.col-sm-12 = t("loan.breakeven.charts_title")
+  .chart.col-lg-6.col-sm-12
```
linesarefuzzy: Thanks for designing for mobile! I'm pretty sure that `col-sm-12` is unnecessary though, as blocks will default to full width if unspecified.

app/views/admin/loans/content/_breakeven_table.html.slim
linesarefuzzy: Oh, right. Never mind.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -0,0 +1,166 @@
...
+    columns = [
+      {"id":"","label":I18n.t('loan.breakeven.product'),"pattern":"","type":"string"},
+      {"id":"","label":I18n.t('loan.breakeven.profit'),"pattern":"","type":"number"}
+    ]
+
+    rows = []
+    for key,product of @breakevenProductProfit()
+      name = key
+      total = product.profit
+      rows.push({"c":[{"v": name, "f":null},{"v": total, "f":null}]})
```
linesarefuzzy: Does google charts require using these one-letter keys? They're a little opaque.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -0,0 +1,166 @@
...
+    @breakevenProductionCosts = @breakevenData["cogs"]
+    @breakevenRevenue = @breakevenData["revenue"]
+    @defaultChartOptions = {
+      width: "100%"
+    }
+
+    @loadCharts()
+
+  breakevenFixedCostsChart: () ->
+    chartData = {}
```
linesarefuzzy: This line seems unnecessary

app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -0,0 +1,166 @@
...
+    chart = new google.visualization.PieChart(document.getElementById('breakeven-total-costs-chart'))
+    chart.draw(chartData, options);
+
+  loadCharts: () ->
+    google.charts.load('current', {'packages':['corechart']});
+    google.charts.setOnLoadCallback @breakevenRevenueChart.bind @
+    google.charts.setOnLoadCallback @breakevenProductionCostsChart.bind @
+    google.charts.setOnLoadCallback @breakevenFixedCostsChart.bind @
+    google.charts.setOnLoadCallback @breakevenProductProfitChart.bind @
+    google.charts.setOnLoadCallback @breakevenTotalCostsChart.bind @
```
linesarefuzzy: Why the `bind @`'s here? Do you know about coffeescript's fat arrow `=>` ? But I would think neither should be necessary for top level functions like these.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -0,0 +1,166 @@
...
+    @breakevenFixedCosts = @breakevenData["fixed_costs"]
+    @breakevenProductionCosts = @breakevenData["cogs"]
+    @breakevenRevenue = @breakevenData["revenue"]
+    @defaultChartOptions = {
+      width: "100%"
+    }
+
+    @loadCharts()
+
+  breakevenFixedCostsChart: () ->
```
linesarefuzzy: No need for empty parens.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -0,0 +1,166 @@
...
+    chartData = new google.visualization.DataTable(chartData);
+    chart = new google.visualization.PieChart(document.getElementById('breakeven-fixed-cost-chart'));
+    chart.draw(chartData, options);
+
+  breakevenProductProfitChart: () ->
+    chartData = {}
+    columns = [
+      {"id":"","label":I18n.t('loan.breakeven.product'),"pattern":"","type":"string"},
+      {"id":"","label":I18n.t('loan.breakeven.profit'),"pattern":"","type":"number"}
+    ]
```
linesarefuzzy: Looking at the [google charts documentation](https://google-developers.appspot.com/chart/interactive/docs/gallery/piechart), it looks like the data format can be simplified significantly. Is there a reason you chose to specify all these keys?

app/assets/javascripts/backbone/views/loan_charts_view.coffee
linesarefuzzy: Couldn't this just be written as: `rows.push [name, total]`?

linesarefuzzy:
Overall, this looks really good. You are getting into some pretty complicated algorithms here and doing great. I'm sorry you put so much effort into the sub-breakdowns in the total costs chart and then we ended up deciding not to use it. It is impressive work anyway. 

I may make some of these changes myself today, or we may have to put some of them off until after the release if you're off for the rest of the day. If so, it's not a problem. Take care of yourself. Thank you for all your hard work!

app/views/admin/loans/content/_breakeven_charts.html.slim
```diff
@@ -0,0 +1,24 @@
...
+    h4 = t("loan.breakeven.profit_by_product")
+    #breakeven-product-profit
+  .chart.col-lg-6.col-sm-12
+    h4 = t("loan.breakeven.fixed_costs", count: 2)
+    #breakeven-fixed-cost-chart
+  .chart.col-lg-6.col-sm-12
+    h4 = t('loan.breakeven.total_cost_breakdown')
+    #breakeven-total-costs-chart
+
+<script type="text/javascript" src="https://www.gstatic.com/charts/loader.js"></script>
```
linesarefuzzy: Shouldn't this be part of the manifest file?

linesarefuzzy: I mean, no, it shouldn't. But, I get errors when it's here. I moved it to the head in base.html.slim for now, but we need to find a better place for it.

emitche:
Liz to make changes in this branch and then merge into develop.

app/views/admin/loans/content/_breakeven_charts.html.slim
emitche: Where do you recommend it should be placed? Thanks for moving it. That was supposed to be temporary and I wish I would have added a note there to make that clear.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
emitche: Simple chart is now in.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
emitche: Ideally these would be `google.charts.setOnLoadCallback @breakevenTotalCostsChart.bind @`, but the data that was set to @[variable] is not reachable once the function is called here. Using the fat arrow here also does not resolve that. Any suggestions?

linesarefuzzy: Weird. I don't get it. But I don't think it matters much. It works.

app/views/admin/loans/content/_breakeven_charts.html.slim
linesarefuzzy: No problem. I looked into it a bit and I think it should be inserted into the head by this partial, so it's only called when needed. That might require a bit of rearranging of `yield`/`content_for`s. Also I think it's a little better to use Rails' `javascript_include_tag`.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
emitche: f is the formatted value. But I found an easier way to format the numbers, so this data structure should be able to be simplified. Will look into that.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
emitche: Done.

emitche:
@linesarefuzzy: I made all the changes except that related to the location of the JS file. We can discuss.

I also suspect that the reason the charts are loading strangely in the view loan questionnaire page but working fine in the print view is due to load order of the JS files. The questionnaire view is doing some not-so-ordinary things in how items are loaded into the DOM. But the reason for things loading strangely could be due to something else. I will let Roxana know about that one so a bug can be filed. I could not quickly address that issue.

PR #130, opened 2016-09-01
Explicitly give location of color variables

emitche:
- Is a work-around for a Sass inside Rails bug

Developer see the colors working properly unless branches are
changed or server is restarted. When that happens, the color
variables cannot be found without explicitly stating location.


emitche:
New styling to changes to develop break the development environment because of a Sass bug that we are using a work-around to address. See comment above.

PR #131, opened 2016-09-01
4989 table questions backend

linesarefuzzy:


linesarefuzzy:
There will be some minor conflicts between this branch and Issue #5014, which I will resolve.

PR #132, opened 2016-09-02
4989 table questions backend

hooverlunch:


app/models/breakeven_table_question.rb
```diff
@@ -0,0 +1,41 @@
+class BreakevenTableQuestion
```
hooverlunch: Hooray for POROs!

app/models/breakeven_table_question.rb
```diff
@@ -0,0 +1,41 @@
...
+      }
+      report[:cogs] << {
+        name: product['name'],
+        quantity: product['quantity'],
+        amount: product['cost'],
+        total: product['cost'] * product['quantity'],
+      }
+    end
+    report[:total_revenue] = report[:revenue].map { |i| i[:total] }.sum
+    report[:total_cogs] = report[:cogs].map { |i| i[:total] }.sum
```
hooverlunch: I believe you can just call sum with a block. Not a big deal though. Just for future ref.

app/models/breakeven_table_question.rb
```diff
@@ -0,0 +1,41 @@
...
+    report[:gross_margin] = report[:total_revenue] - report[:total_cogs]
+    report[:fixed_costs] = @breakeven_data['fixed_costs'].map(&:symbolize_keys)
+    report[:total_fixed_costs] = report[:fixed_costs].map { |i| i[:amount] }.sum
+    report[:net_margin] = report[:gross_margin] - report[:total_fixed_costs]
+
+    report
+  end
```
hooverlunch: Beautiful!

app/models/custom_field.rb
```diff
@@ -148,6 +149,7 @@ def value_types
     result
   end
 
+  # TODO: Not used anywhere? Remove?
```
hooverlunch: Yes nuke it now!

hooverlunch: Yes nuke it now!

app/models/loan.rb
```diff
@@ -69,6 +69,7 @@ class Loan < ActiveRecord::Base
   # beware, the methods generated by this include will fail without the corresponding OptionSet records existing in the database
   attr_option_settable :status, :loan_type, :project_type, :public_level
 
+  # TODO: I think this may be over-abstraction ~Fuzzy
```
hooverlunch: Fair enough. I'm fine with changing it.

app/views/admin/loans/content/_linked_document.html.slim
```diff
@@ -0,0 +1,43 @@
+.attached-document
```
hooverlunch: Good choice to move this.

app/views/admin/loans/content/_table_question.html.slim
```diff
@@ -0,0 +1,2 @@
+/ Displaying report is a separate issue
+= report.inspect
```
hooverlunch: Nicely handled.

app/models/breakeven_table_question.rb
```diff
@@ -0,0 +1,41 @@
+class BreakevenTableQuestion
+  def initialize(breakeven_data)
+    @breakeven_data = JSON.parse(breakeven_data) if breakeven_data.present?
+  end
+
+  def report
```
hooverlunch: Can you please put a comment on this method stating its purpose and include the super nice stub that you made for Liz as an example of the output. Shame to see that go away!

app/models/breakeven_table_question.rb
```diff
@@ -0,0 +1,41 @@
...
+        amount: product['cost'],
+        total: product['cost'] * product['quantity'],
+      }
+    end
+    report[:total_revenue] = report[:revenue].map { |i| i[:total] }.sum
+    report[:total_cogs] = report[:cogs].map { |i| i[:total] }.sum
+    report[:gross_margin] = report[:total_revenue] - report[:total_cogs]
+    report[:fixed_costs] = @breakeven_data['fixed_costs'].map(&:symbolize_keys)
+    report[:total_fixed_costs] = report[:fixed_costs].map { |i| i[:amount] }.sum
+    report[:net_margin] = report[:gross_margin] - report[:total_fixed_costs]
```
hooverlunch: These calculations (there are a lot of them) all seem to assume no nil values. I think this is a problem because we don't currently have any validation on the questionnaire. So you will get errors if you try to multiply something by nil, for instance.

I think it might be better to check each revenue and cogs line and if price/quantity/cost is nil then simply ignore them. Thoughts?

app/models/loan_response.rb
```diff
@@ -69,8 +75,12 @@ def has_boolean?
     field_attributes.include?(:boolean)
   end
 
+  def has_breakeven_table?
+    field_attributes.include?(:breakeven)
+  end
+
   def blank?
-    text.blank? && number.blank? && rating.blank? && boolean.blank? && url.blank?
+    text.blank? && number.blank? && rating.blank? && boolean.blank? && url.blank? && breakeven_data.blank?
```
hooverlunch: Are you sure `.blank?` is the right test here? We will need to make sure to normalize breakeven_data so it doesn't end up like [{x: [], y: []},[],[]] or something, which is not `blank?`.

hooverlunch:
Looks fantastic! Normally I'd say we should have some unit tests for the PORO but given the time constraints and the relatively low risk nature of this (just display logic basically) I think we can let it. go.

app/models/breakeven_table_question.rb
linesarefuzzy: O cool.

app/models/breakeven_table_question.rb
linesarefuzzy: Yes, I suppose so

app/models/breakeven_table_question.rb
linesarefuzzy: Good catch.

linesarefuzzy: What about just setting nil values to 0? 
`%w(quantity price cost).each { |key| product[key] ||= 0 }`

hooverlunch: Works for me.

PR #135, opened 2016-09-09
4993 charts loan memo

emitche:
- Address pull request changes from #129


app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -8,159 +8,172 @@ class MS.Views.LoanChartsView extends Backbone.View
     @breakevenProductionCosts = @breakevenData["cogs"]
     @breakevenRevenue = @breakevenData["revenue"]
     @defaultChartOptions = {
-      width: "100%"
+      width: '100%',
+      backgroundColor: 'none',
+      chartArea: {
```
emitche: This may not be needed.

PR #140, opened 2016-09-14
5098 charts styling

emitche:
- Fix styling tweaks and bugs related to breakeven charts

### DIFF
--- a/app/assets/javascripts/backbone/views/loan_charts_view.coffee
+++ b/app/assets/javascripts/backbone/views/loan_charts_view.coffee
@@ -8,6 +8,14 @@ class MS.Views.LoanChartsView extends Backbone.View
     @breakevenProductionCosts = @breakevenData["cogs"]
     @breakevenRevenue = @breakevenData["revenue"]
     @defaultChartOptions = {
+      height: 200,
+      width: "100%",
+      chartArea: {
+        left: 0,
+        top: 10,
+        width: '100%',
+        height: '90%'
+      },
       backgroundColor: 'none'
     }
     @loadCharts()
--- a/app/assets/javascripts/backbone/views/loan_charts_view.coffee
+++ b/app/assets/javascripts/backbone/views/loan_charts_view.coffee
@@ -13,7 +13,6 @@ class MS.Views.LoanChartsView extends Backbone.View
       chartArea: {
         left: 0,
         top: 10,
-        right: 0,
         width: '100%',
         height: '90%'
       },
--- a/app/assets/javascripts/backbone/views/loan_charts_view.coffee
+++ b/app/assets/javascripts/backbone/views/loan_charts_view.coffee
@@ -9,7 +9,7 @@ class MS.Views.LoanChartsView extends Backbone.View
     @breakevenRevenue = @breakevenData["revenue"]
     @defaultChartOptions = {
       height: 200,
-      width: "100%",
+      width: '100%',
       chartArea: {
         left: 0,
         top: 10,
--- a/app/assets/javascripts/backbone/views/loan_charts_view.coffee
+++ b/app/assets/javascripts/backbone/views/loan_charts_view.coffee
@@ -8,7 +8,6 @@ class MS.Views.LoanChartsView extends Backbone.View
     @breakevenProductionCosts = @breakevenData["cogs"]
     @breakevenRevenue = @breakevenData["revenue"]
     @defaultChartOptions = {
-      width: '100%',
       backgroundColor: 'none'
     }
     @loadCharts()
--- a/app/assets/javascripts/backbone/views/loan_charts_view.coffee
+++ b/app/assets/javascripts/backbone/views/loan_charts_view.coffee
@@ -13,6 +13,7 @@ class MS.Views.LoanChartsView extends Backbone.View
       chartArea: {
         left: 0,
         top: 10,
+        right: 0,
         width: '100%',
         height: '90%'
       },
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -52,18 +58,27 @@ class MS.Views.LoanChartsView extends Backbone.View
     chart.draw(data, options)
 
   breakevenProductProfitChart: ->
-    data = new google.visualization.DataTable();
+    data = new google.visualization.DataTable()
     data.addColumn 'string', I18n.t('loan.breakeven.product')
     data.addColumn 'number', I18n.t('loan.breakeven.profit')
+    options = @defaultChartOptions
 
+    negativeNums = false
```
emitche: There is a possibility that this chart may be negative if cost to produce is more than revenue for products. I've displayed a bar chart if that happens. The default Google Chart error message in red has impact on charts surrounding it if values are negative. Google Charts cannot display negative values in a pie chart.

app/assets/stylesheets/admin/loans/_charts.scss
```diff
@@ -1,3 +1,23 @@
-.chart {
-  text-align: center;
+.breakeven-chart {
+  margin: 2em .5em 1em;
+  page-break-before: always;
+
+  .section-title, .chart-title {
+    display: block !important;
```
emitche: I prefer not to use importants, but the tree that the questions are loaded into has some strange styling. I can add more specificity to these elements, if preferred.

app/views/admin/loans/content/_breakeven_charts.html.slim
```diff
@@ -1,20 +1,20 @@
...
-    h4 = t("loan.breakeven.fixed_costs", count: 2)
-    #breakeven-fixed-cost-chart
-  .chart.col-lg-6
-    h4 = t('loan.breakeven.total_cost_breakdown')
-    #breakeven-costs-chart
+.summary-chart.breakeven-chart
+  h3.section-title = t("loan.breakeven.charts_title")
+  .chart
+    h4.chart-title = t("loan.breakeven.revenue_by_product")
+    #breakeven-revenue-chart.chart-object
```
emitche: If there is more than one breakeven table question, the chart only loads in the first spot, rather than the spot corresponding to the questions that have been answered. I've started work on fixing this, but that was not listed in the things that needed to be changed. Is this a priority?

app/views/layouts/base.html.slim
```diff
@@ -13,6 +13,8 @@ html lang="en-US"
     
     / TODO: Find a better location for this (only used in loan criteria charts)
     = javascript_include_tag "https://www.gstatic.com/charts/loader.js"
+    javascript:
+      google.charts.load('current', {'packages':['corechart']});
```
emitche: Google Charts expects that it will be loaded only once a particular page and will return an error otherwise (at least up to the current version of Google Charts). I moved this out of the backbone view in case there are multiple breakeven question types used in a questionnaire.

emitche: One of the requested fixes in this branch is: "Figure out styling of percentages as some numbers are uncentered in some cases and cut off, in other cases."

The tree that the questions are loaded into is interfering with the usual rendering of Google Charts. A chart can be loaded fine outside of our questionnaire tree and the percentages hide and show based on how large the pie slice is, as expected. Inside the tree, the visual representation of the chart is skewed and the interactions with the chart are disabled or not working. In print view, one can click on a pie slice and see more information about the slice (percent and amount). Though I'm not sure we want that functionality inside the print view, it does make sense inside the loan view where one can edit and view the questionnaire. But clicking the chart does not work inside that context.

emitche: Inside the page where you can edit the loan questionnaire, all issues related to displaying the chart mentioned in the above comment are addressed,

app/assets/javascripts/backbone/views/loan_charts_view.coffee
emitche: I am not a fan of how the bar chart is visually represented on the page, so I may remove this logic and just let Google Charts display the message it displays when an item is negative.

app/assets/stylesheets/admin/loans/_charts.scss
emitche: Elizabeth to fix. Don't use .section-title and .chart-title. Try to use CSS instead to make this scope specific enough that h3 and h4 can be used.

app/views/admin/loans/content/_breakeven_charts.html.slim
```diff
@@ -1,20 +1,20 @@
...
+    #breakeven-production-cost-chart.chart-object
+  .chart
+    h4.chart-title = t("loan.breakeven.profit_by_product")
+    #breakeven-product-profit.chart-object
+  .chart
+    h4.chart-title = t("loan.breakeven.fixed_costs", count: 2)
+    #breakeven-fixed-cost-chart.chart-object
+  .chart
+    h4.chart-title = t('loan.breakeven.total_cost_breakdown')
+    #breakeven-costs-chart.chart-object
```
hooverlunch: Maybe quickly get rid of these semi-redundant classes as discussed.

app/views/admin/loans/content/_breakeven_charts.html.slim
hooverlunch: I agree, not a likely scenario, but perhaps put a comment in the code somewhere noting that this is a known issue.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
emitche: The bar chart breaks the legend of the other two charts and has been removed. If we need to address negative values (which would only occur if it costs more to produce an item than it what the item brings in with its revenue), a separate issue to address this can be filed.

emitche:
I addressed all of @hooverlunch's requests. But GitHub's `Changes requested` feature is not reflecting that, unfortunately.

PR #133, opened 2016-09-02
5014 editable tables

emitche:
Add products and fixed cost tables to loan questionnaire

PR #143, opened 2016-09-15
5014 editable tables

emitche:
- Adjustments to edit breakeven table requested in #133


emitche:
Merged because #133 was set to revise-and-merge and these are requested changes from that pull request.

PR #147, opened 2016-09-23
5022 steps table

emitche:
- Display project steps in timeline table
- Display project groups in timeline table for Redmine issue 5157
- Add fake data for testing


app/views/admin/loans/tabs/_tabs.slim
```diff
@@ -1,7 +1,8 @@
 ul class="nav nav-tabs nav-tabs-responsive hidden-print" role="tablist"
   = render "admin/loans/tabs/tab", identifier: "details", label: t("loan.details"), tab_class: "active", link_class: "edit-tab"
   = render "admin/loans/tabs/tab", identifier: "questions", label: t("loan.questions"), tab_class: "", link_class: "questions-tab"
-  = render "admin/loans/tabs/tab", identifier: "timeline", label: t("loan.timeline"), tab_class: "", link_class: "timeline-tab"
+  = render "admin/loans/tabs/tab", identifier: "timeline-table", label: "Timeline - Table", tab_class: "", link_class: "timeline-tab"
```
emitche: These tab labels are not yet translated, but this is an intermediate feature. Do we need to translate these tabs now?

app/views/admin/loans/timeline/table/_group.html.slim
```diff
@@ -0,0 +1,13 @@
+- wrapper_tag = {tag: is_first ? 'div' : 'tr'}
+- column_span = entry.descendants_only_steps? ? max_depth - entry.depth : 1
+
+*wrapper_tag
+  / The root group is not actually displayed
+  - unless entry.root?
+    td rowspan="#{entry.descendant_step_count}" colspan="#{column_span}" data-item="group"
```
emitche: These data-items have been added to prepare for Redmine issue 5023 and remainder of 5024 (styling of table).

emitche:
I noticed that a manually added step using Timeline List does not display in Timeline Table. Is that the expected behavior?

app/views/admin/loans/timeline/table/_timeline.html.slim
```diff
@@ -0,0 +1,11 @@
+.timeline-table
+  - if @loan.root_timeline_entry
+    - if @loan.root_timeline_entry.is_a?(ProjectGroup)
+      table.table
+        = render "admin/loans/timeline/table/group", entry: @loan.root_timeline_entry, is_first: true,
+          max_depth: @loan.root_timeline_entry.max_descendant_depth
+    / If user has created a step through the timeline list, the timeline table is incompatible
```
emitche: This is a temporary fix. If a loan has project steps, an empty root group should exist. If a new step is created, the step currently becomes the root, which is not what we want. Please see earlier note in pull request.

app/models/project_group.rb
```diff
@@ -37,6 +37,26 @@ class DestroyWithChildrenError < StandardError; end
...
+  end
+
+  # Determine if the group's children are all steps or a mix of steps and groups.
+  def descendants_only_steps?
+    children.each do |c|
+      if c.is_a?(ProjectGroup)
+        return false
+      end
+    end
+    return true
```
hooverlunch: Nice addition! This would do the same thing: `children.all? { |c| c.is_a?(ProjectStep) }`

app/views/admin/loans/tabs/_tabs.slim
hooverlunch: I think it's fine to leave them since they'll just be deleted later.

app/views/admin/loans/timeline/table/_step.html.slim
```diff
@@ -0,0 +1,7 @@
+- wrapper_tag = {tag: is_first ? 'div' : 'tr'}
+
+*wrapper_tag
+  / Add spacer for step who has a group as a sibling
```
hooverlunch: Nice comment!

app/views/admin/loans/timeline/table/_step.html.slim
```diff
@@ -0,0 +1,7 @@
...
+  - if !siblings_are_all_steps
```
hooverlunch: Our styleguide says to always use `unless` instead of `if !`.

app/views/admin/loans/timeline/table/_step.html.slim
```diff
@@ -0,0 +1,7 @@
...
+      td colspan="#{max_depth - entry.depth}" data-item="spacer"
```
hooverlunch: I think this is one indent too many?

hooverlunch: BTW it's fine and better to use classes instead of `data-...` if you need to identify something only for styling.

When I say to prefer to not use classes, I just mean don't use them if you could reasonably select the element some other way (for instance if it's the only `img` tag inside an outer div).

Does that make sense?

app/views/admin/loans/timeline/table/_timeline.html.slim
hooverlunch: Oh interesting. Can you make a note in the next issue in this series for us to fix that. Should be an easy fix, just need to remember to do it. This is fine for now. Good catch!

app/views/admin/loans/timeline/table/_timeline.html.slim
```diff
@@ -0,0 +1,11 @@
...
+    - elsif @loan.root_timeline_entry.is_a?(ProjectStep)
+      #no-steps-msg.alert.alert-warning This loan is not yet compatible with the timeline table.
+  - else
+    #no-steps-msg.alert.alert-warning = t("project_step.no_steps")
```
hooverlunch: Warning is perhaps a bit heavy handed here. I'm not sure if an alert is appropriate actually. I'd just put this as plain text for now. Most sites will have a fairly lightweight message in the zero objects case. Just something so that the user doesn't think the page is still loading or something.

app/views/admin/loans/timeline/table/_timeline.html.slim
emitche: Will do.

app/models/project_group.rb
emitche: Much more succinct!

app/views/admin/loans/timeline/table/_step.html.slim
emitche: Yes. I will replace with classes.

app/views/admin/loans/timeline/table/_timeline.html.slim
emitche: Okay. I used the same alert container as the timeline list when there are no steps to display.

app/models/project_group.rb
X0nic: I usually try to avoid checking the class type. Maybe instead of `is_a?` add two methods. Something like:

```ruby
# timeline_entry.rb
def is_step?
  false
end

# project_step
def is_step?
  true
end
```

It might be overkill, but it would make future refactors easier.

PR #148, opened 2016-09-26
5116: Allow steps that are not finalized to be moved in calendar

emitche:
Both finalized and not finalized steps can have a date change


emitche:
We need to confirm first if non-finalized step date changes from the calendar will also open up the date change modal or simply be moved and saved. Once I have more info on that, I will update this pull request.

emitche:
Change in client specs means this will become a feature and not a bug fix. This has been moved to an upcoming sprint.

PR #149, opened 2016-09-27
5117 bug fix blank log forms save

emitche:
- Require a summary be present to submit log form
- Add error message if a summary is not present
- Ensure empty logs are not saved to the database


app/assets/javascripts/backbone/views/log_modal_view.coffee
```diff
@@ -24,14 +23,27 @@ class MS.Views.LogModalView extends Backbone.View
...
+    $form = @$('form')
+    formIsEmpty = true
+
+    # Check to make sure summary is completed for at least one language
+    # Only submit form if summary is present
+    $form.find("[data-translatable='common.summary']").each ->
+      if ($.trim($(this).val()) != '')
+        $form.submit()
+        @submitted = true
+        formIsEmpty = false
```
hooverlunch: If there are multiple summaries then the form will get submitted multiple times.

You could take the `submit()` outside the loop and check if the @submitted flag is true when you get outside the loop.

Also you don't need both the `formIsEmpty` and `@submitted` flags, right? They are just opposites?

app/views/admin/project_logs/_form.slim
```diff
@@ -4,21 +4,23 @@
...
             = check_box_tag(:step_completed_on_date)
             = t('log.step_completed')
 
-      div.col-lg-6.col-md-6
+      .col-lg-6.col-md-6
         = f.association :agent, include_blank: true, collection: @people
         = f.input :progress_metric_value, as: :select, collection: @progress_metrics
 
-      div.col-lg-6.col-md-6.media-container
+      .col-lg-6.col-md-6.media-container
```
hooverlunch: Remove redundant column classes

app/assets/javascripts/backbone/views/log_modal_view.coffee
emitche: Great way to clean this up. Thanks for that. I will use @submitted instead.

app/views/admin/project_logs/_form.slim
```diff
@@ -4,22 +4,22 @@
...
+          = check_box_tag(:step_completed_on_date)
+          = t('log.step_completed')
 
-      div.col-lg-6.col-md-6
-        = f.association :agent, include_blank: true, collection: @people
-        = f.input :progress_metric_value, as: :select, collection: @progress_metrics
+    = f.association :agent, include_blank: true, collection: @people
+    = f.input :progress_metric_value, as: :select, collection: @progress_metrics
 
-      div.col-lg-6.col-md-6.media-container
```
emitche: Turns out the column and row classes were not needed at all. :)

emitche:
@hooverlunch: Please make sure to give log_modal_view.coffee another look over so I know that's what you wanted there, if you have time. Thanks.

PR #156, opened 2016-10-05
4994 add step group in table

X0nic:


app/controllers/admin/project_groups_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    render_modal_partial
+  end
+
+  def create
+    @entry = ProjectGroup.new(project_group_params)
+    authorize @entry
+
+    @loan = @entry.project
+    @entry.parent = @loan.root_timeline_entry
```
X0nic: Would it make sense to assign `root_timeline_entry` when parent is nil?

app/controllers/admin/project_groups_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    else
+      render_modal_partial(status: 422)
+    end
+  end
+
+  private
+
+  def render_modal_partial(status: 200)
+    link_params = params.slice(:loan_id)
+    @submit_url = @entry.new_record? ? admin_project_groups_path(link_params) : edit_admin_project_groups_path(link_params)
```
X0nic: I didn't add edit support yet, but I left this is for it.

app/views/admin/loans/timeline/table/_timeline_table.html.slim
```diff
@@ -0,0 +1,15 @@
+- max_depth = @loan.root_timeline_entry.max_descendant_depth
+
+= render "admin/loans/timeline/table/timeline_header", loan: @loan
```
X0nic: I think the toughest part of this change for me was finding all of the little partials. Some seem to be under loans, and some not. These views could probably get simplifed by removing the hardcoded paths, and allowing rails to find the correct partial. http://api.rubyonrails.org/classes/ActionView/PartialRenderer.html

I think there would be two major benefits to that. First would be that being forced to follow the rails conventions makes it easier for new peeps to find things. The second would make it a little easier to customize a view based on the classes loaded. IE no need to do `is_a?(ProjectGroup)`.

Ideally we would also remove the locals and ivars, but that is a separate discussion. ;)

rgalusca:
Tested the functionality and there are a few errors. The group is saved as an empty step and not group; on adding a second and third group, they are saved on the same table row as the first one, under the start and end date columns. Also, the add group button should be the dark red color of primary action buttons.

X0nic:
I was able to re-create the issue. From my testing, it appears to be a rendering bug. The groups are in fact added to the DB and displayed. It looks like that if a Group happens to be the first item added to the table, it will render the group name under step instead of shifting it to the left. I agree, this will need to be fixed. Possibly logged in a new issue?

X0nic:
Oh as for the button colour. Maybe I read this wrong? `It should be the default button style, not the primary button style` I read that as use `btn-default` not `btn-primary` for the button style. I can change that back?

X0nic:
`On submit, display success message.` Because this is a remote form, the rails flash messages are not going to be displayed. Is there a method in the site yet to display flash messages for remote forms? Is that something I should spend some time building?

PR #158, opened 2016-10-07
5023 timeline table steps styling

emitche:
- Style steps in timeline table
- Add opacity to project step colors


emitche:
image

This is the color difference with steps. This is an example of a project without groups with several completed steps (early, on time, late, very late, super late).

emitche:
Spacers were changed to white, per @rgalusca's request in this branch. Spacer styling is part of Madeline 5158.

image

app/models/project_step.rb
```diff
@@ -221,34 +221,38 @@ def old_start_date_statement
   end
 
   # Generate a CSS color based on the status and lateness of the step
-  def color
+  def color(opacity)
```
hooverlunch: I suggest opacity should have a default value of 1 here. Then you only need pass it if you want something that is not opaque.

app/models/project_step.rb
```diff
@@ -359,7 +363,7 @@ def handle_scheduled_start_date
   end
 
   # start and finish are each CSS color strings in hsl format
-  def color_between(start, finish, fraction = 0.5)
+  def color_between(start, finish, fraction = 0.5, opacity)
```
hooverlunch: Opacity default here as well?

app/views/admin/loans/timeline/_step.slim
```diff
@@ -1,15 +1,41 @@
...
 *wrapper_tag
   / Add spacer for step who has a group as a sibling
   - unless siblings_are_all_steps
     td.spacer colspan="#{max_group_depth - entry.depth + 1}"
-  td = entry.summary
-  td = entry.display_start_date ? l(entry.display_start_date, format: :short) : ''
-  td = entry.display_end_date ? l(entry.display_end_date, format: :short) : ''
-  td
+
+  td class="#{classes}"
```
hooverlunch: I think you can do just `class=classes`, without the quotes or `#{}`. Give it a try?

hooverlunch: Also, can you put the class on the tr/div and then do `tr.whatever td` in your stylesheet? This is what I usually do for tables and it saves a lot of duplication.

app/views/admin/loans/timeline/_step.slim
```diff
@@ -1,15 +1,41 @@
...
+    input.select-step type="checkbox" class=(entry.completed? ? "completed-item" : "incomplete-item") data-id=entry.id
```
hooverlunch: Could you reuse the `completed` variable here?

emitche: I'm not sure why we have this here unless it is here for when we select all complete/incomplete items. But that would not be the best way to go about selecting all completed, given that completed class will be in the parent element. Just something to consider.

emitche:
Changes made.

PR #160, opened 2016-10-12
5005 add edit step group

X0nic:
Adds the ability to edit groups, as well as create child groups


app/controllers/admin/project_groups_controller.rb
```diff
@@ -14,7 +19,13 @@ def create
     authorize @entry
 
     @loan = @entry.project
-    @entry.parent = @loan.root_timeline_entry
+
+    parent_id = project_group_params[:parent_id]
+    if parent_id && parent_id.present?
```
hooverlunch: Don't think you need both of these expressions. .present? checks for nil? or empty?

app/controllers/admin/project_groups_controller.rb
```diff
@@ -14,7 +19,13 @@ def create
...
+      @entry.parent = ProjectGroup.find(parent_id)
+    else
+      @entry.parent = @loan.root_timeline_entry
+    end
```
hooverlunch: This should be just `@entry.parent = parent_id.present? ? ProjectGroup.find(parent_id) : @loan.root_timeline_entry` I think.

app/controllers/admin/project_groups_controller.rb
```diff
@@ -23,14 +34,32 @@ def create
...
+    if @entry.update(project_group_params)
+      render nothing: true
+    else
+      render_modal_partial(status: 422)
+    end
+  end
+
+  def edit
+    render_modal_partial
+  end
```
hooverlunch: We usually try to follow the Rails scaffold method order. So ISNECUD.

config/locales/es/es.yml
```diff
@@ -389,6 +389,7 @@ es:
 
   project_group:
     add: "Translate Me"
+    edit: "Translate Me"
```
hooverlunch: Can you please add a happy path feature spec, one for adding child and one for editing? I think we need to start doing these as part of frontend issues.

I suggest adding `  gem "capybara-screenshot", "~> 1.0"` to the gemfile. It helps a lot for debugging.

We will also need a web driver. Do you use poltergeist? If so please add it or suggest an alternative. Also can you please add PhantomJS or equivalent to the Readme.md as a dependency for dev envs.

Thank you!

app/assets/javascripts/backbone/views/timeline_table_view.coffee
```diff
@@ -20,9 +22,18 @@ class MS.Views.TimelineTableView extends Backbone.View
 
   newGroup: (e) ->
     e.preventDefault()
-    @modal.show()
+    @modal.new()
+
+  newChildGroup: (e) ->
+    e.preventDefault()
```
X0nic: We see `e.preventDefault()` in each of these methods, but I don't know of a way to clean that up without making some questionable choices / unexpected behaviour.

PR #162, opened 2016-10-13
4997 add new step in tableview

emitche:
- Add step through timeline table


emitche:
Ran into some bugs related to this. See Redmine for more details.

PR #167, opened 2016-10-26
5030 rampup breakeven table

X0nic:
image

Note, as of last night I was unable to get the latest develop to work for me. I was using the latest DB dump from Google Drive. I think we need a data fix there, or there is a bug. I can't resolve the merge error until that is resolved.


X0nic:
Also not great if there are many periods.

project_with_new_era__llc_-_madeline

X0nic:
This is an initial draft of the rampup table. It will obviously need some more work, but I wouldn't mind gathering some feedback before I make the tweaks.

project_with_new_era__llc_-_madeline

X0nic:
It does look better, but there is probably some room for improvement.

project_with_new_era__llc_-_madeline

X0nic:
project_with_new_era__llc_-_madeline

app/assets/stylesheets/admin/loans/_rampup_table.scss
```diff
@@ -0,0 +1,46 @@
+@import '../colors';
+
+.rampup-table {
```
X0nic: This is similar to .summary-table, but not quite the same...

app/models/breakeven_table_question.rb
```diff
@@ -133,10 +133,10 @@ def calculate_line_item_totals(total_key)
       next unless product.present? && product[:name].present?
       {
         name: product[:name],
-        quantity: product.fetch(:quantity, 0),
-        amount: product.fetch(total_key, 0),
+        quantity: product.fetch(:quantity, 0).to_f,
```
X0nic: I fixed a couple more issues with `breakeven_table_question` and added some specs to cover it.

app/views/admin/loans/questionnaires/breakeven_rampup/_main.html.slim
```diff
@@ -0,0 +1,32 @@
+.rampup-table.breakeven-table
+  table.table
+    thead
+      tr.table-title
+        th colspan="4" = t("loan.breakeven_rampup.title")
+
+    tbody
+      = render "admin/loans/questionnaires/breakeven_rampup/rampup_table",
```
X0nic: Seemed like a lot of repetition, so broke out some partials

app/views/admin/loans/questionnaires/breakeven_rampup/_rampup_table.html.slim
```diff
@@ -0,0 +1,26 @@
+tr.group-content
+  td
+    td.text-center colspan=periods
+      = units
```
X0nic: Not sure if this is a great way to do this or not? I could have done this instead with two rows, but less padding?

app/views/admin/loans/questionnaires/breakeven_rampup/_rampup_table.html.slim
```diff
@@ -0,0 +1,26 @@
...
+  td.group-title = title
+  - (1..periods).each do |period|
+    td.text-right = period
+
+- rampup.each do |item|
+  tr.group-content
+    td.group-row-headings =item[:name]
+    - item[:rampup].each do |rampup|
+      td.text-right
+        = number_to_human(rampup[:quantity], precision: 0)
```
X0nic: Same thing here

app/views/admin/loans/questionnaires/_questionnaire_question.html.slim
```diff
@@ -28,6 +28,7 @@
       .view-element
         = render "admin/loans/questionnaires/breakeven_table_report", report: value.breakeven_report
         = render "admin/loans/questionnaires/breakeven_charts", report: value.breakeven_report
+        = render "admin/loans/questionnaires/breakeven_rampup/main", report: value.breakeven_report if value.breakeven_report[:periods] > 1
```
X0nic: I could add a `has_breakeven_rampup?` here?

app/views/admin/loans/questionnaires/breakeven_rampup/_rampup_table.html.slim
linesarefuzzy: I don't think it's valid to nest a td in a td like this. Was that intentional? Otherwise, I think the `br` method is fine.

app/views/admin/loans/questionnaires/breakeven_rampup/_rampup_table.html.slim
```diff
@@ -0,0 +1,26 @@
+tr.group-content
+  td
+    td.text-center colspan=periods
+      = units
+      br
+      = "#{t("loan.breakeven_rampup.quantity")} / #{t("loan.breakeven_rampup.price")}"
+
+tr.group-columns
+  td.group-title = title
+  - (1..periods).each do |period|
```
linesarefuzzy: Why not `periods.each`?

app/views/admin/loans/questionnaires/breakeven_rampup/_rampup_table.html.slim
X0nic: I think that was just supposed to be an empty td.

app/views/admin/loans/questionnaires/breakeven_rampup/_rampup_table.html.slim
X0nic: I think you need a range for this. If I change the code, I get: `undefined method 'each' for 4:Fixnum`

PR #168, opened 2016-10-27
5002 add child step from context menu

emitche:
- Add child step from group menu


emitche:
The backend/controller parts of this need work. @hooverlunch, I am handing that off to you, unless you would like to pair. Currently, this does add a step into the timeline from the project group. But new steps are broken because of the parent_id being undefined and being returned as either 'undefined' or '0'.

hooverlunch:
Paired with Liz today, most work done, Tom to look into blank groups and sorting of options in Parent list.

PR #169, opened 2016-10-28
5134 period fields

X0nic:
Made some changes to the styles

project_with_cleaning_products_cooperative_project_-_madeline

### DIFF
--- a/app/models/breakeven_table_question.rb
+++ b/app/models/breakeven_table_question.rb
@@ -133,10 +133,10 @@ class BreakevenTableQuestion
       next unless product.present? && product[:name].present?
       {
         name: product[:name],
-        quantity: product.fetch(:quantity, 0),
-        amount: product.fetch(total_key, 0),
+        quantity: product.fetch(:quantity, 0).to_f,
+        amount: product.fetch(total_key, 0).to_f,
         total: product.fetch(total_key, 0).to_f * product.fetch(:quantity, 0).to_f,
-      }.merge(rampup(product[:quantity], product[total_key]))
+      }.merge(rampup(product[:quantity].to_f, product[total_key].to_f))
     end.compact
   end
 
@@ -205,7 +205,8 @@ class BreakevenTableQuestion
   end
 
   def periods
-    data_hash[:periods] || 1
+    return data_hash[:periods].to_i if data_hash[:periods].present?
+    1
   end
 
   def units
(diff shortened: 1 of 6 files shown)


app/views/admin/loans/questionnaires/_breakeven_table_question.html.slim
```diff
@@ -25,6 +25,21 @@
         = render 'admin/loans/questionnaires/breakeven_product_row', question: question
 
   .table-container
+    h4.title = t("loan.breakeven.rampup")
```
X0nic: Used the same table style layout as the rest of the subform to keep consistency. Not sure if it would work better as normal form controls.

X0nic:
Not sure if the table style is what was wanted or not. 

project_with_new_era__llc_-_madeline

app/models/breakeven_table_question.rb
```diff
@@ -85,6 +85,15 @@ def initialize(breakeven)
   #   periods: 4,
   #   units: 'Months'
   # }
+
+  PERIOD_TYPES = %i(days weeks months quarters years).freeze
```
linesarefuzzy: Why freeze?

app/views/admin/loans/questionnaires/_breakeven_table_question.html.slim
```diff
@@ -25,6 +25,24 @@
...
+          th = t("loan.breakeven.units")
+      tbody
+        / Hidden row is used to generate new rows
+        / Is at top instead of bottom to not affect striped rows
+        tr.ui-state-default data-group="rampup" class="hidden"
+
+        - if data
+          = render 'admin/loans/questionnaires/breakeven_rampup_row', f: f, periods: data[:periods], units: data[:units], question: question
+        - else
+          = render 'admin/loans/questionnaires/breakeven_rampup_row', f: f, question: question
```
linesarefuzzy: I'm not sure I see why this should be in a partial. Couldn't you just replace `periods` and `units` in the partial with `data[:periods]` and `data[:units]` and even do away with this conditional? Seems that could save several lines of code and some complexity. Your thoughts?

app/models/breakeven_table_question.rb
X0nic: Basically, it's to protect from constants getting mutated accidentally. It seems that rubocop also thinks it's a good idea. http://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/MutableConstant

app/views/admin/loans/questionnaires/_breakeven_table_question.html.slim
X0nic: I was debating with myself if I should do that or not. I like to be more explicit with what gets passed in, but you are right we would save some complexity.

PR #170, opened 2016-11-01
5300: Add initial transaction menu

X0nic:
project_with_djeva_-_madeline

It is hard to read the url in the url bar, so here is it decoded in Chrome.

la_base_united_states_internal

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -93,9 +93,9 @@ module ApplicationHelper
   #  "loan[currency_id]"=>2 ... }
   def to_query_values(object)
     allowed_attribs = object.attributes.select do |_key, value|
-      [Fixnum, String, BigDecimal].any? { |type| value.is_a?(type) }
+      [Fixnum, String, BigDecimal, Date].any? { |type| value.is_a?(type) }
     end
 
-    allowed_attribs.inject({}) { |acc, (key, value)| acc.update("#{object.class.to_s.downcase}[#{key}]" => value) }
+    allowed_attribs.inject({}) { |acc, (key, value)| acc.update("#{object.class.to_s.downcase}[#{key}]" => value.to_s) }
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -53,4 +53,16 @@ module ApplicationHelper
     no_records = grid.current_page_records.length < 1
     render "admin/common/grid", no_records: no_records, grid: grid
   end
+
+  def labase_new_disbursment_url(loan:)
+    "http://internal.labase.org/transactionManager.php?Preset=true&TransactionType=12&Loan=#{loan.id}"
+  end
+
+  def labase_new_repayment_url(loan:)
+    "http://internal.labase.org/transactionManager.php?Preset=true&TransactionType=57&Loan=#{loan.id}"
+  end
+
+  def labase_schedule_url(loan:)
+    "http://internal.labase.org/LoanSchedule.php?&LoanID=#{loan.id}"
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -54,15 +54,48 @@ module ApplicationHelper
     render "admin/common/grid", no_records: no_records, grid: grid
   end
 
+  # Returns the new disbursment url, as well as addtional loan and org information.
+  # This information maybe be required, if the loan does not exist in the old system.
   def labase_new_disbursment_url(loan:)
-    "http://internal.labase.org/transactionManager.php?Preset=true&TransactionType=12&Loan=#{loan.id}"
+    disbursment_transaction_type = 12
+    labase_url('transactionManager.php', loan,
+      'Preset': true, 'TransactionType': disbursment_transaction_type, 'Loan': loan.id)
   end
 
+  # Returns the new repayment url, as well as addtional loan and org information.
+  # This information maybe be required, if the loan does not exist in the old system.
   def labase_new_repayment_url(loan:)
-    "http://internal.labase.org/transactionManager.php?Preset=true&TransactionType=57&Loan=#{loan.id}"
+    repayment_transaction_type = 57
+    labase_url('transactionManager.php', loan,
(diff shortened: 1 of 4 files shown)


X0nic:
Hey @linesarefuzzy, this is basically done. Just missing the date fields right now. How do we want to format the dates for the other system?

linesarefuzzy:
I'd say YYYY-MM-DD. That's pretty much an international standard, for computers at least.

X0nic:
Looks like if I just include Date's but not DateTime's everything comes out properly.
internal_labase_org_loanschedule_php_loanid_1460_loan_5bamount_5d_160000_0_loan_5bcurrency_id_5d_1_loan_5bdivision_id_5d_13_loan_5bid_5d_1460_loan_5blength_months_5d_6_loan_5bname_5d_project_with_grafica_22_de_mayo_loan_5borganization_id_5d

app/helpers/application_helper.rb
```diff
@@ -53,4 +53,49 @@ def render_index_grid(grid)
...
+  end
+
+  # Returns the schedule url, as well as addtional loan and org information.
+  # This information maybe be required, if the loan does not exist in the old system.
+  def labase_schedule_url(loan:)
+    labase_url('LoanSchedule.php', loan, 'LoanID': loan.id)
+  end
+
+  # Takes the page_url merges labase values, with loan values, and returns properly formatted URI
+  def labase_url(page_url, loan, labase_query_values)
```
X0nic: `labase` for the accounting system. Is this a good name?

app/helpers/application_helper.rb
```diff
@@ -53,4 +53,49 @@ def render_index_grid(grid)
     no_records = grid.current_page_records.length < 1
     render "admin/common/grid", no_records: no_records, grid: grid
   end
+
+  # Returns the new disbursment url, as well as addtional loan and org information.
+  # This information maybe be required, if the loan does not exist in the old system.
+  def labase_new_disbursment_url(loan:)
```
X0nic: The url that gets generated is pretty long and verbose, I opted to not include an example in the URL. They look like this: `http://internal.labase.org/LoanSchedule.php?LoanID=1460&loan%5Bamount%5D=160000.0&loan%5Bcurrency_id%5D=1&loan%5Bdivision_id%5D=13&loan%5Bid%5D=1460&loan%5Blength_months%5D=6&loan%5Bname%5D=Project%20with%20Gr%C3%A1fica%2022%20de%20Mayo&loan%5Borganization_id%5D=454&loan%5Borganization_snapshot_id%5D=1325&loan%5Bprimary_agent_id%5D=255&loan%5Bpublic_level_value%5D=hidden&loan%5Brate%5D=32.5&loan%5Bsecondary_agent_id%5D=145&loan%5Bsigning_date%5D=2016-10-03&loan%5Bstatus_value%5D=prospective&organization%5Balias%5D=La%20Cabril&organization%5Bcity%5D=Avellaneda&organization%5Bcountry_id%5D=1&organization%5Bdivision_id%5D=2&organization%5Bemail%5D=quintanacarlos68%40gmail.com&organization%5Bid%5D=454&organization%5Bindustry%5D=Gr%C3%A1fica&organization%5Blegal_name%5D=Cooperativa%20de%20Trabajo%20Gr%C3%A1fica%2022%20de%20MayoLimitada&organization%5Bname%5D=Gr%C3%A1fica%2022%20de%20Mayo&organization%5Bneighborhood%5D=&organization%5Bprimary_phone%5D=4208%201150&organization%5Breferral_source%5D=Red%20Gr%C3%A1fica&organization%5Bsector%5D=Gr%C3%A1fica&organization%5Bstate%5D=&organization%5Bstreet_address%5D=Rivadavia%20700%20%28Avellaneda%29&organization%5Btax_no%5D=30714015458`

And like this when decoded by chrome:
internal_labase_org_loanschedule_php_loanid_1460_loan_5bamount_5d_160000_0_loan_5bcurrency_id_5d_1_loan_5bdivision_id_5d_13_loan_5bid_5d_1460_loan_5blength_months_5d_6_loan_5bname_5d_project_with_grafica_22_de_mayo_loan_5borganization_id_5d

app/helpers/application_helper.rb
```diff
@@ -53,4 +53,49 @@ def render_index_grid(grid)
     no_records = grid.current_page_records.length < 1
     render "admin/common/grid", no_records: no_records, grid: grid
   end
+
```
linesarefuzzy: This is great stuff but it feels wrong to put it in application_helper. How about a new helper file? "old_system_helper"? Or maybe it even belongs in its own class? Thoughts? I bet @hooverlunch has some, if he wants to comment.

In either case, I'd make the transaction type ids and the base_uri into constants.

app/helpers/application_helper.rb
```diff
@@ -53,4 +53,49 @@ def render_index_grid(grid)
...
+  # =>
+  # {"loan[id]"=>1364,
+  #  "loan[amount]"=>#<BigDecimal:7feb498d5658,'0.1E5',9(18)>,
+  #  "loan[currency_id]"=>2 ... }
+  def to_query_values(object)
+    allowed_attribs = object.attributes.select do |_key, value|
+      [Fixnum, String, BigDecimal, Date].any? { |type| value.is_a?(type) }
+    end
+
+    allowed_attribs.inject({}) { |acc, (key, value)| acc.update("#{object.class.to_s.downcase}[#{key}]" => value.to_s) }
```
linesarefuzzy: This is very concise, but a bit dense and not easy to tell what's going on. I think I'd prefer an `each`. But it's not a big deal.

app/helpers/application_helper.rb
X0nic: Yeah I agree, we should move this. Seems like overkill to put in its own class, but I would be fine with either way.

hooverlunch: I'm fine with just a separate helper in this case since the code will eventually go away anyway.

X0nic: Good point, probably not worth spending too much time on it, since it will be going away.

PR #171, opened 2016-11-02
5112 show log summaries

emitche:
- Add log summaries to timeline table

### DIFF
--- a/app/models/project_step.rb
+++ b/app/models/project_step.rb
@@ -81,6 +81,10 @@ class ProjectStep < TimelineEntry
     project_logs.count
   end
 
+  def latest_logs(limit = 3)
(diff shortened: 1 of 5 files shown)


emitche:
Keeping the "More..." text on the same line as the last summary was overly complex. I decided to place the "more" link on its own line.

app/models/project_step.rb
```diff
@@ -81,6 +81,10 @@ def logs_count
     project_logs.count
   end
 
+  def latest_logs(limit = 3)
+    project_logs.order(date: :desc, updated_at: :desc).limit(limit)
```
emitche: As this is being sorted by date and time, it is unlikely that updated_at will be needed.

emitche:
Thanks!

PR #174, opened 2016-11-11
5107 add log from step menu

emitche:
- Add new log from timeline table

### DIFF
--- a/app/assets/javascripts/backbone/views/log_modal_view.coffee
+++ b/app/assets/javascripts/backbone/views/log_modal_view.coffee
@@ -2,7 +2,7 @@ class MS.Views.LogModalView extends Backbone.View
 
   initialize: (options) ->
     @parentView = options.parentView
-    @timelineView = options.timelineView
+    @timelineTableView = options.timelineTableView
     @submitted = false
 
   events:
@@ -46,7 +46,7 @@ class MS.Views.LogModalView extends Backbone.View
 
   submitSuccess: (e, data) ->
     MS.loadingIndicator.hide()
-    if @parentView
-      @parentView.replaceWith(data)
+    if @timelineTableView
+      @timelineTableView.refresh()
     else
-      @timelineView.refresh()
+      @parentView.replaceWith(data)
--- a/app/assets/javascripts/backbone/views/log_modal_view.coffee
+++ b/app/assets/javascripts/backbone/views/log_modal_view.coffee
@@ -2,6 +2,7 @@ class MS.Views.LogModalView extends Backbone.View
 
   initialize: (options) ->
     @parentView = options.parentView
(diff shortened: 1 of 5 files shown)


emitche:
This method allows for the timeline table view to have access to each ProjectStepView and make use of existing methods that were in place from the previous timeline.

app/assets/javascripts/backbone/views/log_modal_view.coffee
```diff
@@ -2,6 +2,7 @@ class MS.Views.LogModalView extends Backbone.View
 
   initialize: (options) ->
     @parentView = options.parentView
+    @timelineTableView = options.timelineTableView
```
hooverlunch: In please try to avoid adding references to other views like this. Consider this an anti-pattern. I know we have done this in the past but it tends to produce unmaintainable code. The better way to enable inter-view communication is by using callbacks and/or promises. We are already using callbacks with the ProjectStepModalView. Promises are really neat too. Check them out if you are not familiar with them.

app/assets/javascripts/backbone/views/project_step_view.coffee
```diff
@@ -9,7 +9,9 @@ class MS.Views.ProjectStepView extends Backbone.View
     @initTypeSelect()
     @persisted = params.persisted
     @duplicate = params.duplicate
-    @context = @$el.data('context')
+    @timelineTableView = params.timelineTableView
```
hooverlunch: I don't think Fuzzy was aware but this class is for the old timeline and I want to remove it once we are fully moved over. So best not to add code to here.

app/assets/javascripts/backbone/views/project_step_view.coffee
```diff
@@ -84,13 +86,21 @@ class MS.Views.ProjectStepView extends Backbone.View
...
+      stepId = @stepId
+    else
+      stepId = @$(link).data('parent-step-id')
+
     unless @logModalView
-      @logModalView = new MS.Views.LogModalView(el: $("<div>").appendTo(@$el), parentView: this)
+      if @timelineTableView
+        @logModalView = new MS.Views.LogModalView(el: $('.timeline-table .log-modal'), timelineTableView: @timelineTableView)
+      else
+        @logModalView = new MS.Views.LogModalView(el: $("<div>").appendTo(@$el), parentView: this)
```
hooverlunch: Note that in this class there are multiple places where you are checking an attribute (`@timelineTableView`) and doing different things depending on it. Consider this an anti-pattern. It indicates that a separate class may be needed or some other kind of refactor.

emitche:
This pattern does not match the desired pattern. A separate branch with this functionality has been merged into develop.

PR #179, opened 2016-11-15
5088: Add new status filter for timeline table

X0nic:


app/controllers/admin/loans_controller.rb
```diff
@@ -165,8 +165,10 @@ def prep_form_vars
   def prep_timeline
     filters = {}
     filters[:type] = params[:type] if params[:type].present?
+    filters[:status] = params[:status] if params[:status].present?
     @loan.root_timeline_entry.filters = filters
     @type_options = ProjectStep.step_type_option_set.translated_list
+    @status_options = [ ['Finalized', 'finalized'], ['Incomplete', 'incomplete'], ['Complete', 'complete'] ]
```
X0nic: I need help with creating a `translated_list` like `ProjectStep.step_type_option_set`

X0nic: There is also no Draft state yet, so I am using Finalized for now.

app/controllers/admin/loans_controller.rb
```diff
@@ -165,8 +165,12 @@ def prep_form_vars
   def prep_timeline
     filters = {}
     filters[:type] = params[:type] if params[:type].present?
+    filters[:status] = params[:status] if params[:status].present?
     @loan.root_timeline_entry.filters = filters
     @type_options = ProjectStep.step_type_option_set.translated_list
+    @status_options = ProjectStep::COMPLETION_STATUSES.map do |status|
+      [ I18n.t("project_step.completion_status.#{status}"), status ]
+    end.sort
```
hooverlunch: Nice. Though you should need to sort these. Just go with a sensible order in the constant.

app/views/admin/loans/timeline/_top_controls.html.slim
```diff
@@ -30,6 +30,11 @@ header.timeline-header.row
         include_blank: I18n.t("project_step.all_types"),
         class: "form-control"
 
+      = select_tag :status,
+        options_for_select(@status_options),
+        include_blank: "All Statuses",
```
hooverlunch: Please i18n the blank option as with the above one.

PR #180, opened 2016-11-16
Add batch actions to timeline table

X0nic:
project_with_nail_salon_ues_-_madeline

### DIFF
--- a/app/assets/javascripts/backbone/views/timeline_batch_actions_view.coffee
+++ b/app/assets/javascripts/backbone/views/timeline_batch_actions_view.coffee
@@ -1,18 +1,15 @@
 class MS.Views.TimelineBatchActionsView extends Backbone.View
 
-  el: 'body'
-
   events: (params) ->
     'confirm:complete .batch-actions .batch-action': 'adjustForm'
-    'click #adjust-dates-confirm': 'hideAdjustDatesModal'
-    'shown.bs.modal #adjust-dates': 'showAdjustDatesModal'
-  
+    'click .adjust-dates-confirm': 'hideAdjustDatesModal'
+
   adjustForm: (e) ->
     item = e.currentTarget
-    methodKey = $(item).attr('data-method-key')
-    actionKey = $(item).attr('data-action-key')
+    methodKey = @$(item).attr('data-method-key')
+    actionKey = @$(item).attr('data-action-key')
 
-    $form = $(item).closest('form')
+    $form = @$(item).closest('form')
 
     $form.find('input[name=_method]').attr('value', methodKey)
     $form.attr('action', actionKey)
@@ -20,8 +17,5 @@ class MS.Views.TimelineBatchActionsView extends Backbone.View
     $form.submit()
 
   hideAdjustDatesModal: (e) ->
-    $('#adjust-dates-modal').modal('hide')
-    @.adjustForm(e)
-
-  showAdjustDatesModal: (e) ->
-    $('#adjust-dates-modal').focus()
+    @$('.adjust-dates-modal').modal('hide')
+    @adjustForm(e)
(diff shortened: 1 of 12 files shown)


X0nic:
This is now complete. I have noticed something that, may or may not be an issue. 

When the batch action is complete, it will always take you back to the `timeline-table`. This is also the case if you are on the `timeline-list`. I could fix this, but is it worth the effort if we will be removing the list eventually?

X0nic:
I also seem to have an issue with closing the modal dialog. It will leave behind a `modal-backdrop`. It seems bootstrap expects only one modal to be present in the DOM. Though I can't find anything specific about this in the docs. I was wondering if anyone has encountered this before, or has a good workaround?

app/views/admin/loans/timeline/_batch_actions.html.erb
```diff
@@ -12,14 +12,14 @@
   </li>
 
   <li>
-    <a id="adjust-dates" class="action" data-toggle="modal"
-      data-target="#adjust-dates-modal">
+    <a class="action adjust-dates" data-toggle="modal"
+      data-target="#<%= locals[:modal_id] %>">
```
X0nic: I am not sure why this is happening here. `modal_id` does not work, but `locals[:modal_id]` does....

app/assets/javascripts/backbone/views/timeline_batch_actions_view.coffee
```diff
@@ -1,27 +1,28 @@
...
+
   adjustForm: (e) ->
     item = e.currentTarget
-    methodKey = $(item).attr('data-method-key')
-    actionKey = $(item).attr('data-action-key')
+    methodKey = @$(item).attr('data-method-key')
+    actionKey = @$(item).attr('data-action-key')
 
-    $form = $(item).closest('form')
+    $form = @$(item).closest('form')
```
hooverlunch: Thanks for these

app/views/admin/loans/timeline/_batch_actions.html.erb
hooverlunch: It's because you're using the short form of `render`, so you're actually passing `locals` as a hash. If you just change it to `render 'foo', modal_id: 123` it should work.

app/views/admin/loans/timeline/_batch_actions.html.erb
```diff
@@ -12,14 +12,14 @@
...
       <%= t("loan.adjust_dates") %>
     </a>
   </li>
 
   <li>
-    <a id="finalize" data-method-key="PATCH"
+    <a data-method-key="PATCH"
```
hooverlunch: Another nice tidy-up!

app/views/admin/loans/timeline/_batch_actions.html.erb
hooverlunch: Also please add a comment saying modal_id can be replaced with a hard-coded value when the old timeline goes away.

app/views/admin/loans/timeline/_list_header.html.slim
```diff
@@ -18,15 +18,15 @@ header.timeline-header
...
+        input.step-ids type="hidden" name="step-ids"
 
       .btn-group role="group" aria-label="Batch Actions"
         button.btn.btn-default.dropdown-toggle type="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"
           => t("loan.batch_actions")
           span.caret
           span.sr-only = t(:dropdown_toggle)
 
-        = render "admin/loans/timeline/batch_actions"
+        = render "admin/loans/timeline/batch_actions", locals: {modal_id: 'adjust-dates-modal-timeline-list'}
```
hooverlunch: Can you please add comments to all places where the `modal_id` local appears saying that we can get rid of it once old timeline is removed.

PR #183, opened 2016-11-17
5425: Ensure log alert when no summary is present shows

emitche:
Alert should show when there is no summary and user tries to save.


emitche:
A Redmine issue has been added: #5425.

PR #189, opened 2016-11-28
5451: Updated loan index table

X0nic:
* Take out "Created" and put in "Point Person"
* Add "Target End Date" column AND filter

### DIFF
--- a/app/views/admin/loans/_loans_grid.html.slim
+++ b/app/views/admin/loans/_loans_grid.html.slim
@@ -36,9 +36,9 @@
       - number_to_currency(loan.amount)
     - g.column name: t('activerecord.attributes.loan.currency'), attribute: 'symbol',
       assoc: :currency, filter: false, in_html: false
-    
+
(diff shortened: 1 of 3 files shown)


app/views/admin/loans/_loans_grid.html.slim
```diff
@@ -53,7 +53,9 @@
     - g.column name: t('activerecord.attributes.loan.first_interest_payment_date'),
       attribute: 'first_interest_payment_date', filter: false, in_html: false
     - g.column name: t('activerecord.attributes.loan.target_end_date'),
-      attribute: 'target_end_date', filter: false, in_html: false
+      attribute: 'target_end_date', in_html: true,
+      detach_with_id: :target_end_date_filter do |loan|
+      - loan.target_end_date ? l(loan.target_end_date) : ''
```
hooverlunch: Please use the `ldate` helper which does the nil check for you.

hooverlunch:
Nevermind I took care of it :)

PR #190, opened 2016-11-29
Select user division on login

X0nic:


app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,17 @@
+class SessionsController < Devise::SessionsController
```
X0nic: This seems to be the best way to do something after login is complete

PR #192, opened 2016-12-01
5449 agent filters

X0nic:


app/filters/conditions_generator_agent_filter.rb
```diff
@@ -0,0 +1,8 @@
+class ConditionsGeneratorAgentFilter < Wice::Columns::ConditionsGeneratorColumn
+  def generate_conditions(_table_alias, search_string)
+    comparator = 'people.name ILIKE ? OR secondary_agents_loans.name ILIKE ?'
+    param = '%' + search_string + '%'
```
hooverlunch: Use string interpolation?

app/filters/conditions_generator_agent_filter.rb
```diff
@@ -0,0 +1,8 @@
...
+    val = [comparator, param, param]
+    val
```
hooverlunch: Can you just remove the `val` stuff and put the array literal as the last line?

app/views/admin/loans/_loans_grid.html.slim
```diff
@@ -36,13 +36,20 @@
...
-    - g.column name: t('activerecord.attributes.loan.secondary_agent'), attribute: 'name',
-      assoc: :secondary_agent, filter: false, in_html: false
-    - g.column name: t('activerecord.attributes.loan.representative'), attribute: 'name',
-      assoc: :representative, filter: false, in_html: false
+
+    - g.column name: t('activerecord.attributes.loan.primary_agent'), in_csv: false, attribute: 'name', assoc: :primary_agent,
+      filter: true, in_html: true, filter_type: :custom_agent_filter, detach_with_id: :agent_filter do |loan|
+      - primary_name = loan.primary_agent ? loan.primary_agent.name : ""
+      - secondary_name = loan.secondary_agent ? ", #{loan.secondary_agent.name}" : ""
+      - primary_name + secondary_name
```
hooverlunch: This is a bit awkward and too much logic for a view I think. Please move this into the model as a method called `agent_names` that returns an array with nils removed. Then you can just do `agent_names.join(", ")` here.

app/views/admin/loans/_loans_grid.html.slim
```diff
@@ -53,7 +60,9 @@
     - g.column name: t('activerecord.attributes.loan.first_interest_payment_date'),
       attribute: 'first_interest_payment_date', filter: false, in_html: false
     - g.column name: t('activerecord.attributes.loan.target_end_date'),
-      attribute: 'target_end_date', filter: false, in_html: false
+      attribute: 'target_end_date', in_html: true,
+      detach_with_id: :target_end_date_filter do |loan|
+      - loan.target_end_date ? l(loan.target_end_date) : ''
```
hooverlunch: Please use the `ldate` helper in application_helper that does the nil check for you.

config/initializers/wice_grid_config.rb
```diff
@@ -89,6 +89,9 @@
   # the value is a 2-element array with 2 strings, the first should be a name of view processor class inherited from
   # Wice::Columns::ViewColumn, the second should be a name of conditions generator class inherited from
   # Wice::Columns::ConditionsGeneratorColumn .
+  Wice::Defaults::ADDITIONAL_COLUMN_PROCESSORS = {
+    custom_agent_filter:  ['Wice::Columns::ViewColumnString', 'ConditionsGeneratorAgentFilter'],
```
hooverlunch: Nice. OOP FTW!

PR #193, opened 2016-12-02
4999 show linked steps on mouseover

emitche:
- Show relationships between steps when hovering over step
- Highlight start and end date of active step
- Highlight start date of any dependent steps, if existing
- Highlight completed date of precedent step, if it exists

### DIFF
--- a/app/assets/javascripts/backbone/views/timeline_table_view.coffee
+++ b/app/assets/javascripts/backbone/views/timeline_table_view.coffee
@@ -29,7 +29,8 @@ class MS.Views.TimelineTableView extends Backbone.View
     'click #project-step-menu a[data-action=duplicate]': 'duplicateStep'
     'click ul.dropdown-menu li.disabled a': 'handleDisabledMenuLinkClick'
     'change form.filters': 'refresh'
-    'click .step-start-date': 'showRelatedSteps' # on mouse over or on mouse enter
+    'mouseenter .step-start-date': 'showRelatedSteps'
+    'mouseleave .step-start-date': 'hideRelatedSteps'
 
   refresh: ->
     MS.loadingIndicator.show()
@@ -125,6 +126,7 @@ class MS.Views.TimelineTableView extends Backbone.View
     $table = $step.closest('tbody')
 
     $step.addClass('highlighted')
+    $step.closest('tr').find('.step-end-date').addClass('highlighted')
 
     # Highlight start date of dependent steps
     $table.find(".step-start-date[data-precedent-id=#{currentId}]").addClass('highlighted')
@@ -135,4 +137,4 @@ class MS.Views.TimelineTableView extends Backbone.View
   hideRelatedSteps: (e) ->
     $step = @$(e.currentTarget)
     $table = $step.closest('tbody')
-    $table.find(td).removeClass('highlighted')
+    $table.find('td').removeClass('highlighted')
--- a/app/assets/javascripts/backbone/views/timeline_table_view.coffee
+++ b/app/assets/javascripts/backbone/views/timeline_table_view.coffee
@@ -29,8 +29,8 @@ class MS.Views.TimelineTableView extends Backbone.View
     'click #project-step-menu a[data-action=duplicate]': 'duplicateStep'
     'click ul.dropdown-menu li.disabled a': 'handleDisabledMenuLinkClick'
     'change form.filters': 'refresh'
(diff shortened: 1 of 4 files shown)


app/views/admin/loans/timeline/_step.html.slim
```diff
@@ -2,7 +2,7 @@
 - completed = entry.completed? ? "completed" : ""
 
 / Note that we have to repeat these classes on all td tags because we don't always have access to the tr tag
-- classes = "#{completed} #{draft}"
+- classes = "#{completed} #{draft} project-step"
```
emitche: Added so that hovering over any part of the step will show related step relationships.

app/assets/javascripts/backbone/views/timeline_table_view.coffee
```diff
@@ -116,3 +118,24 @@ class MS.Views.TimelineTableView extends Backbone.View
...
+
+  showRelatedSteps: (e) ->
+    $step = @$(e.currentTarget).closest('tr').find('.step-start-date')
+    currentId = $step.data('id')
+    precedentId = $step.data('precedent-id')
+    $table = $step.closest('tbody')
+
+    # Highight start and end date of active step
+    $step.addClass('highlighted')
+    $step.closest('tr').find('.step-end-date').addClass('highlighted')
```
hooverlunch: Now that I see this I think it worked slightly different in the old system:

* If you mouseover a start date, it highlights the precedent step's end date (if exists)
* If you mouseover an end date, it hightlights the subsequent steps' start date (if they exist) 
* If you mouseover any other cell, nothing happens

Can we adjust it this way please? Sorry for not being clearer. I think you have all the pieces here, just a matter of rearranging them.

app/assets/javascripts/backbone/views/timeline_table_view.coffee
```diff
@@ -29,6 +29,8 @@ class MS.Views.TimelineTableView extends Backbone.View
     'click #project-step-menu a[data-action=duplicate]': 'duplicateStep'
     'click ul.dropdown-menu li.disabled a': 'handleDisabledMenuLinkClick'
     'change form.filters': 'refresh'
+    'mouseenter .project-step': 'showRelatedSteps'
+    'mouseleave .project-step': 'hideRelatedSteps'
```
hooverlunch: Nice. I didn't know about mouseenter vs mouseover until today!

app/assets/stylesheets/admin/_colors.scss
```diff
@@ -34,3 +34,5 @@ $barelyLate: hsl(56, 100%, 66%);
 $superLate: hsl(0, 74%, 54%);
 
 $veryLate: $orange2;
+
+$projectHoverColor: rgb(255, 255, 221);
```
hooverlunch: I would just call this `$highlight` or `$highlightColor`. The fact that it's used on hover in this case is incidental. We don't seem to be putting `Color` on the end of these things either...

app/views/admin/loans/timeline/_step.html.slim
hooverlunch: See above. You can probably remove this.

app/views/admin/loans/timeline/_step.html.slim
```diff
@@ -20,9 +20,9 @@
         i.fa.fa-fw.fa-calendar
     .summary = entry.summary
 
-  td class=classes
+  td class="#{classes} step-date step-start-date" data-id=entry.id data-precedent-id=entry.schedule_parent_id
     = entry.display_start_date ? l(entry.display_start_date, format: :short) : ""
-  td class=classes
+  td class="#{classes} step-date step-end-date"  data-id=entry.id data-precedent-id=entry.schedule_parent_id
```
hooverlunch: These data attribs can go on the `tr` right? Then you don't need to duplicate? Classes look good.

emitche: Sometimes the step is not in a tr if it is on the same line as the start of a group. I'll look into that though.

emitche:
Thanks, Tom, for this review.

app/views/admin/loans/timeline/_step.html.slim
hooverlunch: Great point! I forgot about that!

That's the same reason we're reusing the `classes` thing.

This is fine as is.

emitche:
@hooverlunch Changes have been made.

PR #195, opened 2016-12-05
4785: Condense content inside date picker

emitche:
Fixes date picker being cut off underneath topmost header


hooverlunch:
Check failure can be ignored for now.

PR #197, opened 2016-12-06
5545 friendly errors

linesarefuzzy:


app/assets/javascripts/backbone/views/application_view.coffee
```diff
@@ -8,10 +8,12 @@ class MS.Views.ApplicationView extends Backbone.View
   initialize: ->
     MS.loadingIndicator = @$('#glb-load-ind')
     MS.errorModal = @$('#glb-error-modal')
+    MS.errorModalDefaultText = I18n.t('error_notification')
```
hooverlunch: Does this need to be a global? The above two do because they are accessed from other classes. But I don't think that's the case here?

app/assets/javascripts/backbone/views/application_view.coffee
```diff
@@ -30,3 +32,19 @@ class MS.Views.ApplicationView extends Backbone.View
...
         self.$el.off 'click', hide # Unregister for performance reasons
     @$el.on 'click', hide
+
+  showErrorModal: (error = MS.errorModalDefaultText) ->
+    MS.errorModal.find('.modal-body').text(error)
+    MS.errorModal.modal('show')
+
+  handleAjaxErrors: ->
+    $(document)
+      .ajaxError (e, jqXHR) =>
```
hooverlunch: Can this be on previous line? So many indents...

app/assets/javascripts/backbone/views/application_view.coffee
```diff
@@ -30,3 +32,19 @@ class MS.Views.ApplicationView extends Backbone.View
...
+        e.stopPropagation()
```
hooverlunch: Not sure if this is necessary since we're listening on `document` but I guess it couldn't hurt.

app/assets/javascripts/backbone/views/timeline_view.coffee
```diff
@@ -48,11 +47,6 @@ class MS.Views.TimelineView extends Backbone.View
       MS.loadingIndicator.hide()
       @addSteps(html)
 
-  submitError: (e) ->
-    e.stopPropagation()
-    MS.errorModal.modal('show')
-    MS.loadingIndicator.hide()
-
```
hooverlunch: Yay!

app/controllers/admin/admin_controller.rb
```diff
@@ -5,7 +5,19 @@ class Admin::AdminController < ApplicationController
   before_action :authenticate_user!
   after_action :verify_authorized
 
+  rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
+
   def admin_controller?
     true
   end
+
+  # Should be overridden on ajax-type controllers
```
hooverlunch: Remove?

PR #198, opened 2016-12-06
5550: Add additional & private notes to automatic emails

linesarefuzzy:


linesarefuzzy:
I actually don't think this needs a review.

PR #199, opened 2016-12-07
5480 logs tab in loan

emitche:
- Show list of logs for specific loan in loan tab
- Show at most 50 logs per page


emitche:
The failed tests are also failing on develop in my local environment.

emitche:
See #202. Redmine 5481 is builds off of 5480, so the pull requests are combined.

PR #201, opened 2016-12-12
5063: Use unique ids for each chart

X0nic:


app/views/admin/loans/questionnaires/_breakeven_charts.html.slim
```diff
@@ -2,21 +2,21 @@
...
+    div id="breakeven-fixed-cost-chart-#{id}"
   .chart
     h4 = t('loan.breakeven.total_cost_breakdown')
-    #breakeven-costs-chart
+    div id="breakeven-costs-chart-#{id}"
 
 javascript:
   $(function() {
-    new MS.Views.LoanChartsView({breakeven_data: #{json report}});
+    new MS.Views.LoanChartsView({breakeven_data: #{json report}, id: "#{id}"});
```
X0nic: Would it make more sense to refactor `.summary-chart.breakeven-chart` to have a unique id, and pass that in as the EL?

app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -37,7 +37,7 @@ class MS.Views.LoanChartsView extends Backbone.View
       total = product.total
       data.addRow [name, total]
 
-    chart = new google.visualization.PieChart(document.getElementById('breakeven-revenue-chart'));
+    chart = new google.visualization.PieChart(document.getElementById("breakeven-revenue-chart-#{@id}"))
```
hooverlunch: Single quotes in JS please.

app/assets/javascripts/backbone/views/loan_charts_view.coffee
```diff
@@ -111,7 +111,7 @@ class MS.Views.LoanChartsView extends Backbone.View
     data.addRow ["Fixed Costs", @breakevenData["total_fixed_costs"]]
 
     options = @defaultChartOptions
-    chart = new google.visualization.PieChart(document.getElementById('breakeven-costs-chart'))
+    chart = new google.visualization.PieChart(document.getElementById("breakeven-costs-chart-#{@id}"))
```
hooverlunch: Can you refactor to use `@$(...)` instead of getElementById. Can you also please refactor all these very similar lines into one helper method.

app/views/admin/loans/questionnaires/_breakeven_charts.html.slim
hooverlunch: Yes, absolutely. Just give it a data-id attribute and give the `div`s classes instead of IDs.

app/views/admin/loans/questionnaires/_breakeven_charts.html.slim
```diff
@@ -2,21 +2,21 @@
...
-    #breakeven-revenue-chart
+    div id="breakeven-revenue-chart-#{id}"
   .chart
     h4 = t("loan.breakeven.production_cost_by_product")
-    #breakeven-production-cost-chart
+    div id="breakeven-production-cost-chart-#{id}"
   .chart
     h4 = t("loan.breakeven.profit_by_product")
-    #breakeven-product-profit
+    div id="breakeven-product-profit-#{id}"
```
hooverlunch: For consistency, this one should end in `chart` also, no?

hooverlunch:
One question, are you sure this solves the problem? I thought it was because `google.charts.load()` was getting called multiple times, and we had discussed a different solution.

If you found that it was something else, that's great, but please reference this in your PR notes, otherwise I get confused :)

X0nic:
Yes, it ended up being a different problem than the multiple calls to `google.charts.load()`. Just happened to be reusing elements with the same ID. Ill make sure next time to note a change like that.

PR #203, opened 2016-12-14
4617 division locales

X0nic:


app/assets/stylesheets/admin/forms/_forms.scss
```diff
@@ -30,6 +30,13 @@ form {
   }
 }
 
+.checkbox {
```
X0nic: Not sure if there is a better way to deal with this with simple form. Not super familiar with how the bootstrap wrappers work.

app/models/division.rb
```diff
@@ -102,4 +103,15 @@ def has_noncascading_dependents?
...
+  def locales
+    return [] unless self[:locales].present?
+    self[:locales].sort.select(&:present?).map(&:to_sym)
+  end
+
+  def locale_names
+    locales.map do |locale|
+      I18n.t("locale_name.#{locale}", locale: locale)
+    end
+  end
```
hooverlunch: Nice stuff here.

app/views/admin/divisions/_form.html.slim
```diff
@@ -22,6 +22,10 @@
     .view-element = @division.default_currency.try(:name)
     = f.input_field :default_currency_id, collection: @currency_choices, include_blank: true
 
+  = f.input :locales
+    .view-element = @division.locale_names.join(', ')
+    = f.input_field :locales, as: :check_boxes, collection: locale_options, include_blank: true, checked: @division.locales
```
hooverlunch: Very smooth.

hooverlunch:
I did make a few changes. Please take a look at my commits. Thanks.

PR #205, opened 2016-12-15
5453 logs list

emitche:
- Edit log in logs list
- Delete log in logs list
- Expand/hide all in logs list
- Refactor logs list for multiple contexts (logs page, org logs, loan logs tab)


emitche:
These are the remaining problems with this feature:
- Logs list modal shown from the more link for a step in timeline table is not working
- Edit and delete actions inside logs list modal are not complete
- The edit and delete functions may be broken in old timeline
- Logs list modal may need some styling tweaks including expand all link

emitche:
The base is a different branch. If this needs to be based on develop, that is fine.

PR #208, opened 2016-12-28
5116 bugfix date change on unfinalized steps

emitche:
- Ensure steps that are not finalized can be moved
- Ensure changing a step date with dependents also moves the dependents
- Ensure steps with precedent steps cannot be moved
- Indicate steps with precedent steps on the calendar

### DIFF
--- a/app/controllers/admin/timeline_step_moves_controller.rb
+++ b/app/controllers/admin/timeline_step_moves_controller.rb
@@ -27,6 +27,7 @@ class Admin::TimelineStepMovesController < Admin::AdminController
     render nothing: true
   end
 
+  # A change of date for a step that does not require a corresponding log
   def simple_move
     @step = ProjectStep.find(params[:id])
     authorize @step, :update?
--- a/app/controllers/admin/timeline_step_moves_controller.rb
+++ b/app/controllers/admin/timeline_step_moves_controller.rb
@@ -30,7 +30,7 @@ class Admin::TimelineStepMovesController < Admin::AdminController
   def simple_move
     @step = ProjectStep.find(params[:id])
     authorize @step, :update?
-    @step.update_attributes(:scheduled_start_date => params[:scheduled_start_date])
+    @step.update_attributes(scheduled_start_date: params[:scheduled_start_date])
     render nothing: true
   end
 
--- a/app/controllers/admin/timeline_step_moves_controller.rb
+++ b/app/controllers/admin/timeline_step_moves_controller.rb
@@ -27,6 +27,13 @@ class Admin::TimelineStepMovesController < Admin::AdminController
     render nothing: true
   end
 
+  def simple_move
+    @step = ProjectStep.find(params[:id])
+    authorize @step, :update?
+    @step.update_attributes(:scheduled_start_date => params[:scheduled_start_date])
+    render nothing: true
+  end
+
   private
 
   def project_step_move_params
--- a/app/models/calendar_event.rb
+++ b/app/models/calendar_event.rb
@@ -14,7 +14,9 @@ class CalendarEvent
   attr_accessor :completion_status
   attr_accessor :time_status
   attr_accessor :num_of_logs
+
   attr_accessor :has_precedent
+  alias_method :has_precedent?, :has_precedent
 
   def self.build_for(model)
     puts "model: #{model.inspect}"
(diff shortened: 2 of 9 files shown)


app/serializers/calendar_event_serializer.rb
```diff
@@ -1,8 +1,9 @@
 class CalendarEventSerializer < ActiveModel::Serializer
-  attributes :start, :html, :id, :model_id, :editable, :is_finalized, :completed, :model_type, :event_type
+  attributes :start, :html, :id, :model_id, :editable, :is_finalized, :completed, :model_type, :event_type, :has_precedent
```
emitche: What is the recommended way to list attributes when the line is too long?

emitche: What is the recommended way to list attributes when the line is too long?

app/models/calendar_event.rb
```diff
@@ -94,6 +95,7 @@ def initialize_project_step(step)
     @step_type = step.step_type_value
     @completion_status = step.completion_status
     @time_status = step.days_late && step.days_late > 0 ? "late" : "on_time"
+    @has_precedent = step.schedule_parent_id ? true : false
```
emitche: Do I need true and false here?

app/models/calendar_event.rb
```diff
@@ -14,6 +14,7 @@ class CalendarEvent
   attr_accessor :completion_status
   attr_accessor :time_status
   attr_accessor :num_of_logs
+  attr_accessor :has_precedent
```
hooverlunch: Boolean accessors like this should always have a `?` at the end. So you could either define it as a regular method or you could add `alias_method :has_precedent?, :has_precedent`. I recommend the latter as it's less code and you'll need the non `?` version in the serializer.

app/serializers/calendar_event_serializer.rb
```diff
@@ -1,8 +1,9 @@
...
 
   def editable
     return false if object.event_type == "ghost_step"
+    return false if object.has_precedent
```
hooverlunch: Use the `?` version here. You should also combine these two using `||`.

app/views/admin/calendar/_event.html.slim
```diff
@@ -12,7 +12,8 @@
   - when "project_step"
 
     div.calendar-event.cal-step [
-      class="cal-step-#{cal_event.step_type} #{cal_event.time_status}"
+      class="cal-step-#{cal_event.step_type} #{cal_event.time_status}
+        #{cal_event.has_precedent ? 'has-precedent': ''} "
```
hooverlunch: Very nice. Can use `?` version here too and throughout the templates.

app/views/admin/calendar/_legend.html.slim
```diff
@@ -13,14 +13,23 @@ div#legend-content
             td
             th.bold = t("calendar.status.incomplete")
           tr
+            td: span: i.fa.fa-pencil-square-o
```
hooverlunch: Nice! I didn't know about this syntax...

app/views/admin/calendar/_legend.html.slim
```diff
@@ -13,14 +13,23 @@ div#legend-content
...
+            th = t("calendar.num_of_logs")
+          tr
+            td: span.original-date
+            th = t("calendar.original_date")
+          tr
+            td: span: i.fa.fa-link
+            th = t("calendar.has_precedent")
```
hooverlunch: Very nice to remember this!

config/routes.rb
```diff
@@ -46,7 +46,11 @@
     resources :project_groups
 
     # Does it make sense to surround in separate namespace?
-    resources :timeline_step_moves
+    resources :timeline_step_moves do
```
hooverlunch: Can we get rid of the comment on line 48 please.

app/controllers/admin/timeline_step_moves_controller.rb
```diff
@@ -27,6 +27,14 @@ def create
     render nothing: true
   end
 
+  # A change of date for a step that does not require a corresponding log
```
emitche: @hooverlunch: Is this enough for a message? I spoke with linesarefuzzy about adding this method to this file specifically. There's a bit more going on here than just the saving of the log. The other methods refer to the Timeline StepMove model, but that's not needed here. We are simply changing a date in this method.

PR #209, opened 2016-12-30
5598: Initial work to get project model

X0nic:


app/models/basic_project.rb
```diff
@@ -0,0 +1,43 @@
+# == Schema Information
+#
+# Table name: basic_projects
+#
+#  created_at         :datetime         not null
+#  division_id        :integer
```
linesarefuzzy: It is nullable on other models (but validated as present). Maybe there should be a new issue to change them all to not null.

app/models/basic_project.rb
```diff
@@ -0,0 +1,43 @@
...
+#  id                 :integer          not null, primary key
+#  primary_agent_id   :integer
+#  project_type_value :string
```
linesarefuzzy: Actually we are removing this field. It was an artifact from the old system.

app/models/basic_project.rb
```diff
@@ -0,0 +1,43 @@
...
+#  secondary_agent_id :integer
+#  start_date         :date
+#  status_value       :string
```
linesarefuzzy: I think so

linesarefuzzy:
I'm down with STI (yeah you know I?). Re polymorphism, wouldn't the normal setup be to have the association on the parent class, like how Loan has many timeline entries rather than project steps and groups?

X0nic:
@hooverlunch I have moved everything around for Loans and BasicProject to inherit from project. 

I am not sure I have fixed all of the relations. Specifically the references to project that already exist. Things seem to work though. 

Also, there is some strangeness on ProjectLog. Is the comment about basic project still accurate?

hooverlunch:
You can try putting the `has_many :timeline_entries` association on Project instead of Loan and BasicProject. But I'm not sure that the `belongs_to` from TimelineEntry to Project will work. It might complain if you try to assign a Loan to the belongs_to :project association because it's the wrong class. I thought I did a spike and it didn't work, but it's worth trying again. Just please be sure to test everything in both directions, and also test the `has_many :project_logs, through: :timeline_entries, source: :project, source_type: 'ProjectStep'` one on Loan. That would move to Project.

hooverlunch:
Oh just seeing your comments Nathan. I will take a look at all this tomorrow. So please leave assigned to me.

hooverlunch:
And thanks!

PR #210, opened 2016-12-30
5579 make loan name editable

emitche:
- Ensure user can change loan name
- Make a loan default name method for auto-generated name
- Replace instances where loan.name is used in project
- Ensure displayed name works across edit, new, and show loan form/page

Unrelated bug fix:
- Fix displayed loan currency and amounts so that the new loan form can load


emitche:
I do have one question inside Redmine about 3 files that may or may not need the attribute name to be replaced.

app/views/admin/loans/_form.html.slim
```diff
@@ -38,11 +39,11 @@
     = f.input_field :status_value, collection: Loan.status_options, include_blank: true
 
   = f.input :amount
-    .view-element = format_currency(@loan.amount, @loan.currency, tooltip: false)
+    .view-element = @loan.id ? format_currency(@loan.amount, @loan.currency, tooltip: false) : ''
```
emitche: The new loan form was broken here and in two other spots. If a loan does not exist yet, then these places where loan.id is used would not get rendered, which ensures the new loan form can load.

config/locales/en/loans.en.yml
```diff
@@ -12,6 +12,8 @@ en:
     confirm_response_deletion: "WARNING: Are you sure you want to permanently delete these responses?"
     criteria: "Criteria"
     criteria_not_entered: "Criteria have not been entered yet."
+    default_name_message: "If left blank, a name will be automatically generated for this loan."
+    default_loan_name: "%{org} - %{date}"
```
emitche: Removes the need to use `.html_safe` in the methods and as a bonus adds internationalization. I did replace `&ndash;` with `-`, which should be fine.

app/views/admin/loans/_form.html.slim
```diff
@@ -16,9 +16,10 @@
   = authorized_form_field(simple_form: f, model: @loan, field_name: :organization,
     choices: @organization_choices, include_blank_choice: false, classes: 'hidden-print')
 
-  / = f.input :name
-  /   .view-element = @loan.name
-  /   = f.input_field :name
+  = f.input :name
+    .view-element = @loan.id ? @loan.display_name : ''
```
hooverlunch: The proper way to check if an object is persisted is to use `.persisted?` or `.new_record?` (the opposite). Checking `id` works but is less clear.

app/views/admin/loans/_form.html.slim
hooverlunch: Nice work. See my notes above.

app/models/loan.rb
```diff
@@ -117,9 +117,15 @@ def agent_names
...
+  def default_name
     date = signing_date || created_at.to_date
-    "#{organization.name} &ndash; #{I18n.l date}".html_safe
+    I18n.t("loan.default_loan_name", org: organization.name, date: I18n.l(date))
+  end
+
+  def display_name
+    if self.name.blank?
+      self.default_name
+    else self.name end
```
hooverlunch: Separate lines please.

hooverlunch: Actually, you don't need `self` anywhere here, and you can/should use `? :`. So `name.blank? ? default_name : name`. Please try to remember to think about when `self` is necessary. Usually only on assignment. I have pointed this out several times before.

app/views/admin/loans/_form.html.slim
hooverlunch: Also I don't get why `display_name` would break if the loan isn't persisted. Maybe because `organization` is nil? If so, you should be able to handle that in the model instead of in the template.

app/views/admin/loans/_loans_grid.html.slim
```diff
@@ -8,11 +8,12 @@
     - g.column name: t('common.id'), attribute: 'id', filter: true, in_csv: false, detach_with_id: :id_filter do |loan|
       - link_to(loan.id, admin_loan_path(loan.id))
     - g.column name: csv_id, attribute: 'id', in_html: false, filter: false
-    - g.column name: t('activerecord.attributes.loan.name'), attribute: 'name',
-      in_html: false, filter: false
+
+    - g.column name: t('activerecord.attributes.loan.name'), attribute: 'name' do |loan|
+      - loan.display_name
```
hooverlunch: This column has lost its `in_html: false, filter: false`. Was that intended?

config/locales/en/loans.en.yml
hooverlunch: I'm not quite sure I follow you here. A string from I18n will only be html_safe if the key ends in `html`: http://guides.rubyonrails.org/i18n.html#using-safe-html-translations. But that shouldn't matter because you've removed the only html. 

So just changing `&ndash;` to - should be enough to remove the need for .html_safe, and I'm fine with that. But this doesn't need to be in here as there is nothing to translate.

app/models/calendar_event.rb
```diff
@@ -119,7 +119,7 @@ def initialize_loan_start(loan)
 
   def initialize_loan_end(loan)
     @start = loan.target_end_date
-    @title = "End " + loan.name
+    @title = "End " + loan.display_name
```
hooverlunch: Should be using string interpolation here and above. Please try to avoid `+` with strings. If you can't use interpolation you should be using `<<`. Please look up the difference and let me know if it's not clear.

app/views/admin/loans/_loans_grid.html.slim
emitche: Yes. Loan name was only displaying in the CSV before. Now it displays in both places so only one item is needed (no specifying in_html and in_csv required, as it displays in both).

app/models/calendar_event.rb
emitche: This probably needs to be translated too, so will add these to translations.

config/locales/en/loans.en.yml
emitche: Smart. Thanks.

hooverlunch:
Nice catch on the need to translate 'End'!!

app/views/admin/loans/_loans_grid.html.slim
emitche: I am confirming loan.display_name shows up in both the loan index page and an export of the loans data.

app/views/admin/loans/_loans_grid.html.slim
```diff
@@ -32,8 +33,13 @@
       custom_filter: Loan.loan_type_options, detach_with_id: :loan_type_filter do |loan|
       - loan.loan_type
     - g.column name: t('activerecord.attributes.loan.amount'), attribute: 'amount',
-      filter: false do |loan|
-      - format_currency(loan.amount, loan.currency, tooltip: false)
+      filter: false, in_csv: false do |loan|
+      - if loan.amount && loan.currency
```
emitche: Same as before. If displayed loan amount becomes a method, helper, or some other attribute, it could be used here. However, the specifics are outside the scope of this feature.

I've only fixed this so that CSV export is properly working so that we can see that loan.display_name is exporting properly.

PR #212, opened 2017-01-03
5587: Add string/short answer option for loan responses

emitche:
- Bug fix: Add text input for loan response strings


app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
```diff
@@ -84,6 +84,11 @@
               = f.input_field :"#{question.attribute_sym}[text]", as: value.text_form_field_type,
                 value: value.text
 
+            - if value.has_string?
+              - if value.string.present?
+                .view-element.answer = simple_format(value.string)
```
hooverlunch: Keep mind for future that `simple_format` only adds `<p>` and `<br>` tags which don't make sense for a one-line input field. But this is fine for now.

PR #213, opened 2017-01-04
5577: Redirect to loan show when only one record found

X0nic:


app/helpers/application_helper.rb
```diff
@@ -54,6 +54,15 @@ def render_index_grid(grid)
     render "admin/common/grid", no_records: no_records, grid: grid
   end
 
+  def render_loan_grid(grid)
+    if grid.all_pages_records.count == 1 && grid.filtered_by == ['id']
```
X0nic: Maybe it would make sense to generalize this? I am going to guess the other grids would have the same attributes, but there are none with ID filters to test on.

hooverlunch: Yes you should generalize. Otherwise it doesn't belong in application_helper.

You should be able to do `redirect_to grid.all_pages_records.first` right? Then it doesn't have to mention `loan` at all? You could just call it `render_index_grid_with_redirect_check` or something.

BUT

Now that I think of it, it seems pretty strange to be redirecting in a helper. I can imagine my future self tearing my hair out trying to figure out why this action is redirecting. This really belongs in the controller. Please find a way to do it there. e.g. add a method called `redirect_if_one_match_and_id_filter(relation)` to ApplicationController and call it right after the `filter_by_params` in the LoansController, passing `@loans`.

X0nic: As far as I understand how wice_grid works, is that it does the filtering on its own. `filter_by_params` is never called. Wice_grid wouldn't even let me access the 'all_pages_records' in the controller. `render_index_grid_with_redirect_check` is going to be the easier approach for sure. I could look into a filter based approach, but it is going to require some research first.

hooverlunch:
Yikes. I see what you mean though, and I think you're right. Ok to go for
the helper approach.

PR #214, opened 2017-01-04
5407 show logs in loan calendar

emitche:
- Show logs list at bottom of calendar step modal
- Hide logs list in calendar step edit mode
- Adjust language for condensing logs in list


app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -1,18 +1,21 @@
-
 - if @logs.empty?
   = content_tag(:div, t('common.no_records'), class: 'alert alert-info')
 - else
   .show-actions.hidden-print
     a data-expand-all=".logs-index" href="#" = t('common.expand_all')
-    a data-hide-all=".logs-index" href="#" = t('common.hide_all')
+    - if @step && @context && @context == "calendar"
```
hooverlunch: If `@context` is `nil` then it won't equal `"calendar"` so you don't need to check `@context`.

app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -1,18 +1,21 @@
...
+      a data-hide-all=".logs-index" href="#" = t('common.hide_all')
 
-  - if @step
-    h5
-      - if @step.milestone?
-        i.fa.fa-fw.fa-flag-o
-      - elsif @step.checkin?
-        i.fa.fa-fw.fa-calendar
-      = @step.summary
+  - unless @context && @context == "calendar"
```
hooverlunch: Same here.

app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -1,18 +1,21 @@
...
+    - if @step
+      h5
+        - if @step.milestone?
+          i.fa.fa-fw.fa-flag-o
+        - elsif @step.checkin?
+          i.fa.fa-fw.fa-calendar
```
hooverlunch: This code exists in two places I believe, and there is already a `project_step_icon` method in project_step_helper.rb. Can you please refactor both to use that helper method and make sure they still work. Thanks!

app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -1,18 +1,21 @@
...
 - if @logs.empty?
   = content_tag(:div, t('common.no_records'), class: 'alert alert-info')
 - else
   .show-actions.hidden-print
     a data-expand-all=".logs-index" href="#" = t('common.expand_all')
-    a data-hide-all=".logs-index" href="#" = t('common.hide_all')
+    - if @step && @context && @context == "calendar"
+      a data-hide-all=".logs-index" href="#" = t('common.condense_all')
+    - else
+      a data-hide-all=".logs-index" href="#" = t('common.hide_all')
```
hooverlunch: The only thing changing here is the link text, yes? It would be better to use a ternary operator inside the `t` method call. That way you're not duplicating the rest of the code.

app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -30,31 +33,35 @@
             .item
               i.fa.fa-commenting-o>
               = log.progress_metric || t('project_step.no_status')
-            - unless params[:org].present? || params[:loan].present? || params[:step].present?
+
+            - unless @context.present? && @context == "calendar"
```
hooverlunch: Don't need to check `present?` here either

app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -30,31 +33,35 @@
...
+              - unless params[:org].present? || params[:loan].present? || params[:step].present?
+                .item
+                  a href=admin_organization_path(step.project.organization)
+                    i.fa.fa-building-o>
+                    = step.project.organization.name
+              - unless params[:step].present?
+                .item
+                  a href=(admin_loan_path(step.project) + "/timeline-table#step-#{step.id}")
```
hooverlunch: Please use `<<` instead of `+`.

app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -30,31 +33,35 @@
...
+                    = project_step_icon(step)
+                    = step.summary.present? ? step.summary : t('log.untitled_step')
```
hooverlunch: Here you are checking if summary is present but not above. Is that on purpose?

I like to put this kind of logic into the model with a method like `summary_or_untitled`. Then you can reuse it wherever you want a friendly display of the summary. Else you end up duplicating this same logic all over.

app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -30,31 +33,35 @@
...
+
+            - unless @step && @context && @context == "calendar"
```
hooverlunch: Don't need to check `@context`.

app/views/admin/project_logs/_log_list.html.slim
hooverlunch: I now see this code preexisted. Please take a look anyway :)

app/views/admin/project_logs/_log_list.html.slim
hooverlunch: Also, what is the purpose of checking `@step`? I see a mixture of if statements here, some checking `@step`, some not. What happens if we change them all to just `@context == "calendar"`? That would be much clearer...

app/views/admin/project_steps/_modal_content.html.slim
```diff
@@ -99,6 +99,14 @@ section.step-fields class=(@mode == :show_and_form ? 'show-view' : 'edit-view')
           .view-element = @step.is_finalized ? t("reply_yes") : t("reply_no")
           = f.input_field :is_finalized, as: :boolean, boolean_style: :inline
 
+    - if @logs.length > 0 && context == "calendar"
```
hooverlunch: Please use `.any?` instead of `.length > 0`.

app/views/admin/project_logs/_log_list.html.slim
emitche: I will look and see. (I didn't add this code in this feature. It already existed. I just moved the code inside an if statement.)

app/views/admin/project_logs/_log_list.html.slim
emitche: :)

emitche:
Thanks for the review. I will get to this by tomorrow morning. Your comments are clear.

hooverlunch:
Yes I knew that :)

On Wed, Jan 4, 2017 at 2:08 PM, Elizabeth Mitchell <notifications@github.com

PR #216, opened 2017-01-04
5611: Show an error message if date is missing

X0nic:


X0nic:
Currently, this will just show an error dialog when the user tries to submit a missing date. It doesn't say much in the ticket, so I am not sure if this is required or not.

I did look into what we could do. There is already a backbone view ensuring the summary is populated. It will show an error in JS if that field is missing. I could do something similar, but since its not really the rails way, I am not sure we would want to go that way.

I tried to modify the form to re-render if the validation failed on the server side. While that would work, there is a JS event that catches all ajax errors and displays a modal dialog. Some refactoring will be required to make it configurable somehow. 

Anyways, this works. I'll need to know what way we would like to proceed.

app/views/admin/project_logs/_form.html.slim
```diff
@@ -1,14 +1,9 @@
 .log-form
-  = simple_fields_for(@log) do |f|
+  = simple_fields_for(@log, remote: true) do |f|
+    = error_notification(f)
 
     = f.hidden_field(:project_step_id)
 
-    // The hide/show for this alert is in LogFormModalView and MoveStepModalView.
```
X0nic: I removed the JS validations. I assume we want to validate in rails now?

X0nic:
@hooverlunch I assume we want the summary validation moved into rails, but what about other things like `agent`? While the inputs won't not let you pick one, if it ever does happen it will break viewing notes for that step.

PR #217, opened 2017-01-05
4623 notes

linesarefuzzy:


linesarefuzzy:
Ah, I left failing tests. Working on that now.

linesarefuzzy:
Done.

app/views/admin/notes/_form.html.slim
```diff
@@ -0,0 +1,20 @@
...
+  
+  = f.error_notification
+  
+  = f.input :text, as: :text, label: "Note"
+  
+  = f.input :notable_id, as: :hidden
+  = f.input :notable_type, as: :hidden
+
+  button.btn.btn-default.cancel> type="button" = t :cancel
+  = f.submit 'Save', class: 'submit btn btn-success'
```
hooverlunch: I18n

app/controllers/admin/organizations_controller.rb
```diff
@@ -28,6 +28,10 @@ def show
     @org = Organization.find(params[:id])
     authorize @org
     prep_form_vars
+
+    @notes = @org.notes.order(created_at: :desc)
```
hooverlunch: Have you checked for N+1 here? I bet you need to load at least the author and notable of each note due to the policy check. Or maybe your fancy loader thing is handling it ;) Worth looking at the log.

app/views/admin/notes/_show.html.slim
```diff
@@ -0,0 +1,23 @@
+header.small
+  .post-info
+    .item
+      i.fa.fa-calendar-o>
+      = l note.created_at.to_date
+    .item
+      a href=admin_person_path(note.author)
+        i.fa.fa-user>
+        = note.author.name
```
hooverlunch: Technically you should be using delegators for this kind of thing `e.g. author_name` so you don't violate LoD. It's a good habit to get in.

app/views/admin/notes/_show.html.slim
```diff
@@ -0,0 +1,23 @@
...
+    - if policy(note).edit?
+      .item
+        a.edit-action href="#"
+          i.fa.fa-pencil>
+          = t(:edit)
+    - if policy(note).destroy?
+      .item
+        => link_to admin_note_path(note), method: :delete, class: 'delete-action',
+            data: { confirm: t('confirm_deletion', thing: t('common.note')) },
```
hooverlunch: This kind of I18n interpolation is not a good idea because in some languages (e.g. French) the word before `thing` depends on `thing`. In French it could be 'ce' or 'cette'. I know this is used in a couple other places but in most places we have separate strings. Can you please refactor this and the other two and get rid of that I18n key?

config/locales/en/notes.en.yml
```diff
@@ -0,0 +1,8 @@
+en:
+  notes:
+    new: "New Note"
+  admin:
+    notes:
+      form:
+        edit_item: "Edit Note"
+        new_item: "New Note"
```
hooverlunch: Nice.

spec/policies/note_policy_spec.rb
```diff
@@ -1,5 +1,41 @@
...
+  let(:record) { create(:note, notable: org) }
+
+  context "division member" do
+    let(:user) { create(:user, :member, division: division) }
+
+    it { should permit_action(:create) }
+
+    context "other's note" do
+      it { should forbid_action(:edit) }
+      it { should forbid_action(:update) }
```
hooverlunch: Use `permit_edit_and_update_actions`?

spec/policies/note_policy_spec.rb
```diff
@@ -1,5 +1,41 @@
...
+  subject { described_class.new(user, record) }
+
+  let(:division) { create(:division) }
+  let(:org) { create(:organization, division: division) }
+  let(:record) { create(:note, notable: org) }
+
+  context "division member" do
+    let(:user) { create(:user, :member, division: division) }
+
+    it { should permit_action(:create) }
```
hooverlunch: Use `permit_new_and_create_actions`?

hooverlunch: Also I believe `should` is deprecated. Use `is_expected.to`?

app/assets/javascripts/backbone/views/note_view.coffee
```diff
@@ -0,0 +1,56 @@
...
+
+  update: (e) ->
+    $form = @$(e.target).closest('form')
+    # We send form data via ajax so we can capture the response from server
+    $.post($form.attr('action'), $form.serialize())
+      .done (response) =>
+        @$('.view-block').html(response)
+        @$('.view-block').show()
+        @$('.form-block').hide()
+      .fail (response) =>
```
X0nic: How did you get this part to work @linesarefuzzy ? On 5611 I get `ajaxError()` from `application_view` jumping in the way and blocking my error handler.

linesarefuzzy: Actually I just removed this because it was redundant with the main application view's error handler. I'm not sure which was taking precedence or how they might have interfered with each other.

PR #219, opened 2017-01-13
5586 media upload bug

linesarefuzzy:


app/assets/javascripts/backbone/views/error_handler.coffee
```diff
@@ -0,0 +1,25 @@
...
+    MS.errorModal.find('.modal-body').text(error)
+    MS.errorModal.modal('show')
+
+  handleAjaxErrors: ->
+    $(document).ajaxError (e, jqXHR, ajaxSettings, errorString) =>
+      e.stopPropagation()
+      $('.modal').modal('hide')
+      MS.loadingIndicator.hide()
+      switch parseInt(jqXHR.status)
+        when 0, 200 then # do nothing - false alarm (no idea why ajaxError sometimes catches these)
```
X0nic: That only seems to happen on file uploads. Maybe a bug in JQuery?

app/assets/javascripts/backbone/views/error_handler.coffee
```diff
@@ -0,0 +1,25 @@
...
+        when 403
```
X0nic: What happens on a 422? In my branch I had to specifically handle the 422, or I did not get validation errors.

app/assets/javascripts/backbone/views/error_handler.coffee
```diff
@@ -0,0 +1,25 @@
+class MS.Views.ErrorHandler extends Backbone.View
+
+  el: 'body'
+
+  initialize: ->
+    MS.loadingIndicator = @$('#glb-load-ind')
```
hooverlunch: This doesn't belong in this view, I think. It's not strictly related to error handling.

app/assets/javascripts/backbone/views/error_handler.coffee
hooverlunch: Weird. It's fine how you've handled it though. Good comment.

app/assets/javascripts/backbone/views/error_handler.coffee
hooverlunch: I think it's correct to also ignore 422s here. They should be handled by the individual Views, not here.

hooverlunch: @X0nic Can you please add a commit doing this and address the loadingindicator thing above. Then you can please mark as complete.

app/assets/javascripts/backbone/views/error_handler.coffee
X0nic: @hooverlunch I don't think it can move... Otherwise, we have no way to hide the loading indicator when the error happens.

hooverlunch: But it's a global...

PR #220, opened 2017-01-17
5635: Adding loan id to loan pages

X0nic:
I think this should match what was requested in the ticket. Here are some screens.

845__project_with_idelgraff_-_madeline

845__project_with_idelgraff_-_madeline

### DIFF
--- a/app/views/admin/loans/_form.html.slim
+++ b/app/views/admin/loans/_form.html.slim
@@ -16,6 +16,10 @@
   = authorized_form_field(simple_form: f, model: @loan, field_name: :organization,
     choices: @organization_choices, include_blank_choice: false, classes: 'hidden-print')
 
+  = f.attribute :id
+    .static-text-as-field
+      = @loan.id
+
   = f.input :name
     .view-element = @loan.display_name
     = f.input_field :name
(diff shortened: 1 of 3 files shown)


app/views/admin/loans/show.html.slim
```diff
@@ -1,4 +1,4 @@
-- content_for(:title, @loan.display_name)
+- content_for(:title, "#{@loan.id}, #{@loan.display_name}")
```
hooverlunch: I like it. Nice and simple. Can you please change the comma to a colon though, and put a #. So like 

#5674: Project with Sassy Pants

Thanks!

app/views/admin/loans/_form.html.slim
```diff
@@ -16,6 +16,10 @@
   = authorized_form_field(simple_form: f, model: @loan, field_name: :organization,
     choices: @organization_choices, include_blank_choice: false, classes: 'hidden-print')
 
+  = f.attribute :id
+    .static-text-as-field
+      = @loan.id
```
hooverlunch: Please add this to the en.yml file as `ID`. Then it will get translated along with the other attribs, and I think ID looks better than Id (unless you're talking about Freudian psych :))

PR #221, opened 2017-01-18
5618 add kind dropdown

X0nic:


app/assets/javascripts/backbone/views/application_view.coffee
```diff
@@ -39,6 +39,11 @@ class MS.Views.ApplicationView extends Backbone.View
 
   handleAjaxErrors: ->
     $(document).ajaxError (e, jqXHR) =>
+      # Don't capture error on form validation errors.
+      # File uploads do some strange things, hence the 200 and 0
+      status = parseInt(jqXHR.status)
+      return if status == 422 || status == 0 || status == 200
```
X0nic: I could clean this up, but is there a better way than looking at the status code in the error handler?

db/migrate/20170118054424_change_media_kind_files_to_documents.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeMediaKindFilesToDocuments < ActiveRecord::Migration
+  def up
+    execute "UPDATE media SET kind = 'document' WHERE kind = 'file'"
```
X0nic: Maybe I should also rename `kind` to `kind_value`? All other translated option sets seem to be using that convention.

app/assets/javascripts/backbone/views/application_view.coffee
X0nic: Looks like #219 is going to handle some of this.

PR #222, opened 2017-01-18
WIP: 5592 projects index page

emitche:
- Add projects index view including table and filters


emitche:
In the system, there is a difference between Projects and BasicProjects. Refactoring to use BasicProjects.

PR #223, opened 2017-01-18
5074 inactive status loan questions

emitche:
- Add active/inactive checkbox in loan question form
- Adjust styling of inactive questions in manage loan questions
- Adjust styling of inactive questions inside loan questionnaire

In the develop branch, inactive questions and groups were already are hidden by default if they did not have a response.

### DIFF
--- a/app/serializers/loan_question_serializer.rb
+++ b/app/serializers/loan_question_serializer.rb
@@ -1,5 +1,5 @@
 class LoanQuestionSerializer < ActiveModel::Serializer
-  attributes :id, :name, :children, :parent_id, :fieldset, :optional, :required_loan_types, :active
+  attributes :id, :name, :children, :parent_id, :fieldset, :optional, :required_loan_types, :status
 
   def initialize(*args, loan: nil, **options)
     @loan = loan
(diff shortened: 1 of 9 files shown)


emitche:
Per Roxana: "If answered, the question will appear grayed out, but not the answer." I need to fix this, as both are grayed. I will address during the code review.

app/serializers/loan_question_serializer.rb
```diff
@@ -29,4 +29,8 @@ def optional
   def required_loan_types
     object.loan_question_requirements.map { |i| i.loan_type.label.to_s }
   end
+
+  def status
+    object.status ? object.status : "active"
```
hooverlunch: Equivalent but shorter and more conventional: `object.status || "active"`

BUT -- you need to make sure status can't be empty string ("") instead of nil. Because empty string evaluates to `true` when used in a condition. If that is a possibility, you can use `object.status.presence || "active"`. `.presence` returns `nil` if the value is `.empty?`, or just the value itself otherwise.

emitche:
@rgalusca:

This is how an inactive question that has been answered currently looks in show and edit mode. I'm not sure if in show mode you want the answer to still be black or not. Just let me know if adjustments are needed.

image
image

PR #225, opened 2017-01-18
5592 basic projects index page

emitche:
- Add basic projects index page
- Add filters for table


emitche:
The people filter is not in yet. 5728 (#224) has a small tweak that fixes and important bug so that this filter will work.

app/models/basic_project.rb
```diff
@@ -44,7 +44,21 @@
...
+  def start_date
+    signing_date
+  end
+
   def default_name
     I18n.t("common.untitled")
   end
+
+  def display_name
+    name.empty? ? default_name : name
```
hooverlunch: If `name` is `nil`, this will return `name`. `name.empty?` is `false`. You probably want `blank?`?

app/views/admin/basic_projects/_basic_projects_grid.html.slim
```diff
@@ -0,0 +1,44 @@
...
+    - g.column name: t('activerecord.attributes.basic_project.secondary_agent'), filter: true, in_html: false
+      do |project|
+      - project.secondary_agent.try(:name)
+
+    - g.column name: t("activerecord.attributes.basic_project.start_date"),
+      attribute: 'signing_date', filter: true, detach_with_id: :start_date_filter do |project|
+      - ldate(project.start_date)
+    - g.column name: t('activerecord.attributes.basic_project.length_months'),
+      attribute: 'length_months', filter: false do |project|
+        - if project.length_months
```
hooverlunch: Is this indented one too far?

app/views/admin/basic_projects/_basic_projects_grid.html.slim
```diff
@@ -0,0 +1,44 @@
...
+
+    - g.column name: t('common.created'), attribute: 'created_at', filter: false,
+      in_csv: false do |project|
+      - fuzzy_time(project.created_at)
+    - g.column name: t('common.created'), attribute: 'created_at', in_html: false, filter: false
+
+    - g.column name: t('common.updated'), attribute: 'updated_at', filter: false,
+      in_csv: false do |project|
+      - fuzzy_time(project.updated_at)
+    - g.column name: t('common.updated'), attribute: 'updated_at', in_html: false, filter: false
```
hooverlunch: Very nice job on the grid.

app/views/admin/basic_projects/index.html.slim
```diff
@@ -0,0 +1,33 @@
...
+
+      .actions
+        button.btn.btn-link.wg-external-reset-button data-grid-name="basic_projects"
+          = t("clear")
+
+        button.btn.btn-primary.wg-external-submit-button.action data-grid-name="basic_projects"
+          = t("filter")
+      .clearfix
+
+= render_index_grid(@basic_projects_grid)
```
hooverlunch: I'm noticing here that there are two lines of code in this file (and all our grid index files) that seem to say "render index grid". Should we rename all the grid partials to e.g. `basic_projects_grid_spec` or `basic_projects_grid_definition`? Then it would be clear that the first render is for the definition only, while the second is where the code actually comes in...?

config/locales/en/en.yml
```diff
@@ -13,6 +13,15 @@ en:
       person: "Member"
       project_step: "Project Step"
     attributes:
+      basic_project:
+        agents: "Agents"
+        length_months: "Length (months)"
+        name: "Name"
+        primary_agent: "Primary Agent"
+        secondary_agent: "Secondary Agent"
+        signing_date: "Start Date"
```
hooverlunch: Should this say "Signing Date"?

db/migrate/20170118185717_add_basic_project_status_options.rb
```diff
@@ -0,0 +1,15 @@
...
+      model_attribute: 'status')
+    basic_project_status.options.destroy_all
+    basic_project_status.options.create(value: 'active',
+      label_translations: {en: 'Active', es: 'Activo'})
+    basic_project_status.options.create(value: 'completed',
+      label_translations: {en: 'Completed', es: 'Completo'})
+    basic_project_status.options.create(value: 'changed',
+      label_translations: {en: 'Changed', es: 'Cambiado'})
+    basic_project_status.options.create(value: 'possible',
+      label_translations: {en: 'Possible', es: 'Posible'})
```
hooverlunch: So yeah, fine for now!

db/seeds.rb
```diff
@@ -36,6 +36,18 @@
...
+  model_attribute: 'status')
+basic_project_status.options.destroy_all
+basic_project_status.options.create(value: 'active',
+  label_translations: {en: 'Active', es: 'Activo'})
+basic_project_status.options.create(value: 'completed',
+  label_translations: {en: 'Completed', es: 'Completo'})
+basic_project_status.options.create(value: 'changed',
+  label_translations: {en: 'Changed', es: 'Cambiado'})
+basic_project_status.options.create(value: 'possible',
+  label_translations: {en: 'Possible', es: 'Posible'})
```
hooverlunch: Nice. Already translated!

config/locales/en/en.yml
emitche: So in the database, it is signing date. But this is a basic project and not a loan. So we are using signing date as start date. Using the language "Signing Date" is not to the spec and does not seem like it would match the intent of a basic project.

db/migrate/20170118185717_add_basic_project_status_options.rb
emitche: linesarefuzzy suggested using doing create only instead of `find_or_create_by` and not doing the destroy all. Would that address your concerns? There is also a seed in this branch.

It looks like you are fine with this as is, however.

db/seeds.rb
emitche: linesarefuzzy did these while we paired.

config/locales/en/en.yml
hooverlunch: 👍

PR #226, opened 2017-01-19
5680 members notes

linesarefuzzy:


app/assets/javascripts/backbone/views/notable_view.coffee
```diff
@@ -1,6 +1,4 @@
-class MS.Views.OrganizationView extends Backbone.View
-
-  el: '.coop'
```
hooverlunch: Hooray for removing explicit `el`s!

app/assets/javascripts/backbone/views/person_view.coffee
```diff
@@ -3,6 +3,7 @@ class MS.Views.PersonView extends Backbone.View
 
   initialize: ->
     @showHideUserFields()
+    new MS.Views.NotableView({ el: @el })
```
hooverlunch: Braces and spaces not needed :)

app/views/admin/organizations/show.html.slim
```diff
@@ -33,21 +33,13 @@ section.coop.details
...
-        = render partial: 'admin/notes/note', locals: { note: @new_note }
+  section.block
+    = render "admin/notes/notes_section"
 
 = render "admin/media/modal"
 
 javascript:
   $(function() { 
-    new MS.Views.OrganizationView();
+    new MS.Views.NotableView({ el: 'coop' });
```
hooverlunch: Should there be a `.` before `coop`? I'm curious how this is even working currently... Please be sure to test all concrete instantiations when you abstract something out like this...

hooverlunch: Or is there a new `<coop>` tag? That would be sweet!

linesarefuzzy: Huh. Yup, that's a typo. But it works. No idea how.

PR #227, opened 2017-01-19
5669 Remove project type from loan

X0nic:


X0nic:
Conflict fixed. Ready to merge.

PR #228, opened 2017-01-19
5636 change target_end_date to end_date

X0nic:


X0nic:
Conflicts fixed. May need to be done again depending on how these are merged.

PR #229, opened 2017-01-19
5593 project details

emitche:
- Add project show page
- Add project details and form
- Allow project to be edited and deleted
- Link to specific projects from projects index


app/controllers/admin/basic_projects_controller.rb
```diff
@@ -20,4 +20,50 @@ def index
...
     end
   end
+
+  def show
+    @project = BasicProject.find(params[:id])
+    authorize @project
+    prep_form_vars
+  end
+
+  def destroy
```
hooverlunch: Typically destroy is after update.

app/controllers/admin/basic_projects_controller.rb
```diff
@@ -20,4 +20,50 @@ def index
...
+    else
+      prep_form_vars
+      render :show
+    end
+  end
+
+  private
+
+  def prep_form_vars
+    @agent_choices = person_policy_scope(Person.in_division(selected_division).where(has_system_access: true)).order(:name)
```
hooverlunch: I believe the more conventional way to do this is `policy_scope(Person).in_division...`. I see this `person_policy_scope` stuff is being used in this app but it's not necesary to create these extra methods as Pundit gives you `policy_scope`: https://github.com/elabs/pundit#scopes

Would be great if you could refactor all these please.

hooverlunch: In particular: fix this one, see if it works, then fix the others. :)

app/controllers/admin/basic_projects_controller.rb
```diff
@@ -20,4 +20,50 @@ def index
...
+  end
+
+  def basic_project_params
+    params.require(:basic_project).permit(*(
+      [
+        :length_months, :name, :primary_agent_id, :secondary_agent_id, :signing_date,
+        :status_value
+      ]
+    ))
```
hooverlunch: Simplify as discussed :)

app/views/admin/basic_projects/_form.html.slim
```diff
@@ -0,0 +1,46 @@
+= simple_form_for([:admin, @project],
+    html: { \
+      class: 'form-horizontal organization-record',
+    },
+    wrapper: :horizontal_form_condensed,
+    defaults: { \
+      input_html: {class: 'form-element form-control'},
+    },
+  ) do |f|
```
hooverlunch: At some point we need to refactor all these forms to get rid of all this boilerplate. I think I mentioned this before. This would be a really good exercise in frontend refactoring. @rgalusca can we get a form refactoring issue on the docket for next sprint? Perhaps with a 5 hour estimate? This really needs to get done and we're saving a ton of time on this issue...

app/views/admin/basic_projects/_form.html.slim
```diff
@@ -0,0 +1,46 @@
...
+    .view-element = @project.try(:name)
+    = f.input_field :name
+
+  = f.input :status
+    .view-element = @project.try(:status)
+    = f.input_field :status_value, collection: BasicProject.status_options, include_blank: true
+
+  = authorized_form_field(simple_form: f, model: @project, field_name: :primary_agent,
+    choices: @agent_choices, classes: 'hidden-print')
+  = f.input t('activerecord.attributes.basic_project.primary_agent'), wrapper_html: {class: 'visible-print-block'}
```
hooverlunch: I'm not sure what's happening here with the field name being translated in the template. If you just put `:primary_agent` as the name, won't the translation happen automatically?

app/views/admin/basic_projects/_form.html.slim
```diff
@@ -0,0 +1,46 @@
...
+    span = @project.primary_agent.try(:name)
+
+  = authorized_form_field(simple_form: f, model: @project, field_name: :secondary_agent,
+    choices: @agent_choices, classes: 'hidden-print')
+  = f.input t('activerecord.attributes.basic_project.secondary_agent'), wrapper_html: {class: 'visible-print-block'}
+      span = @project.secondary_agent.try(:name)
+
+  = f.input :signing_date, label: t("activerecord.attributes.basic_project.start_date")
```
hooverlunch: Same here. I think the explicit label can be removed. Just change the key in the en.yml file to match the field name.

app/views/admin/basic_projects/_form.html.slim
```diff
@@ -0,0 +1,46 @@
...
+    .view-element = @project.try(:signing_date)
+    = f.input_field :signing_date, as: :date_picker
+
+  = f.input :length_months
+    .view-element = t("basic_project.months", number: @project.length_months.to_s) if @project.length_months
```
hooverlunch: Don't think you need the `to_s` here.

app/views/admin/basic_projects/show.html.slim
```diff
@@ -0,0 +1,32 @@
...
+      - if policy(@project).destroy?
+        = link_to [:admin, @project], method: :delete,
+            data: { \
+              confirm: t('.confirm_deletion'),
+              "confirm-title": t('.delete'),
+              "confirm-cancel": t(:cancel),
+              "confirm-proceed": t(:confirm)\
+          }
+          i.fa.fa-trash.fa-large>
+          = t(".delete")
```
hooverlunch: This edit and delete link code should really be refactored into a helper method too. I bet almost the exact same code shows up in about 10 places. The only thing different would be the object. We could include this in our form refactor job. Let's pair on this next week.

config/locales/en/basic_projects.en.yml
```diff
@@ -0,0 +1,11 @@
...
+  # The admin top level group is used here to allow t(".edit") style shortcuts from views
+  admin:
+    basic_projects:
+      show:
+        confirm_deletion: "WARNING: Are you sure you want to permanently delete this Project?"
+        delete: "Delete Project"
+        edit: "Edit Project"
+
+  basic_project:
+    months: "%{number} mo."
```
hooverlunch: Nice job here.

app/views/admin/basic_projects/_form.html.slim
emitche: Yes, @hooverlunch, we can pair on that when it is part of a sprint. Thanks for keeping an eye on that.

app/views/admin/basic_projects/show.html.slim
emitche: That sounds like a great idea.

app/controllers/admin/basic_projects_controller.rb
emitche: I fixed this one. As we are at end of sprint, I am going to address these others in a different pull request. That way I can test and make sure all is working well rather than rushing that change in. Thanks, @hooverlunch.

emitche:
@hooverlunch requested refactors of code outside of this feature. I will address in a separate pull request.

PR #232, opened 2017-01-24
5674 styling display logs

emitche:
- Move show/hide logs list links into logs list partial
- Ensure show/hide logs links are on same line as expand/condense logs links
- Ensure refactor works across app where logs lists are displayed
- Hide expand all link when logs list is not expanded

### DIFF
--- a/app/assets/javascripts/backbone/views/expander.coffee
+++ b/app/assets/javascripts/backbone/views/expander.coffee
@@ -12,7 +12,7 @@ class MS.Views.Expander extends Backbone.View
   expand: (e) ->
     e.preventDefault()
     $link = @$(e.currentTarget)
-    targetName = link.data('expands')
+    targetName = $link.data('expands')
     target = @$("[data-expandable='#{targetName}']")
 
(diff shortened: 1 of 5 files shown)


emitche:
I have asked @rgalusca a question about design in Redmine.

emitche:
I need to make one tweak to the UI.

app/assets/javascripts/backbone/views/expander.coffee
```diff
@@ -11,26 +11,38 @@ class MS.Views.Expander extends Backbone.View
 
   expand: (e) ->
     e.preventDefault()
-    link = @$(e.currentTarget)
-    targetName = link.data('expands')
+    $link = @$(e.currentTarget)
```
hooverlunch: Can you please update the comment at the top of this file to describe:

1. expandAll/hideAll (these never got added to the comment)
2. The nested expander scenario (that you have expandables within an expandable and you want to be able to expand all the inner expandables with one click)

I think that would make this easier to understand later.

PR #233, opened 2017-01-24
5632 Display precedent step or parent group when in view mode

X0nic:


app/views/admin/project_steps/_modal_content.html.slim
```diff
@@ -69,11 +69,15 @@ section.step-fields class=(@mode == :show_and_form ? 'show-view' : 'edit-view')
...
-        label: t("project_step.timeline_table.preceding_step"), include_blank: true
+      = f.input :parent_id
+        .view-element = @step.parent.summary_or_none if @step.parent.present?
+        = f.input_field :parent_id, collection: @parents, label_method: :summary_or_none,
+          label: t("project_step.timeline_table.group"), include_blank: false
+
+      = f.input :schedule_parent_id
+        .view-element = @step.schedule_parent.summary_or_none if @step.schedule_parent.present?
+        = f.input_field :schedule_parent_id, collection: @precedents, label_method: :summary_or_none,
+          label: t("project_step.timeline_table.preceding_step"), include_blank: true
```
hooverlunch: Can we please move the labels for these two inputs into the activerecord part of the en.yml? Or is there some reason we can't?

app/views/admin/project_steps/_modal_content.html.slim
```diff
@@ -69,11 +69,15 @@ section.step-fields class=(@mode == :show_and_form ? 'show-view' : 'edit-view')
...
 
-      = f.input :parent_id, collection: @parents, label_method: :summary_or_none,
-        label: t("project_step.timeline_table.group"), include_blank: false
-
-      = f.input :schedule_parent_id, collection: @precedents, label_method: :summary_or_none,
-        label: t("project_step.timeline_table.preceding_step"), include_blank: true
+      = f.input :parent_id
+        .view-element = @step.parent.summary_or_none if @step.parent.present?
+        = f.input_field :parent_id, collection: @parents, label_method: :summary_or_none,
+          label: t("activerecord.attributes.project_step.parent"), include_blank: false
```
X0nic: @hooverlunch What is wrong here? I changed to activerecord translations and now it's broken. No matter what I put in the translation, it always says "parent". Even if I hard code the label....

X0nic: When on develop the form looks like:

_773__ferrox_-seven_-_nov_18__2016_-_madeline


On this branch, because of the last commit it looks like:

_773__ferrox_-seven_-_nov_18__2016_-_madeline

As far as I can tell, the tranlations are set up properly. So it should not be displaying "Parent" and "Schedule Parent"

config/locales/en/en.yml
```diff
@@ -124,13 +133,6 @@ en:
...
         accent_fg_color: "Color of text on top of accent color. Defaults to banner FG color."
     labels:
-      project_step:
-        actual_end_date: "Completed Date"
-        duration: "Scheduled Duration"
-        is_finalized: "Finalized"
-        scheduled_duration_days: "Scheduled Duration"
-        scheduled_start_date: "Scheduled Start Date"
-        step_type_value: "Type"
       timeline_step_move:
```
X0nic: I wasn't able to move `timeline_step_move` since it is not an active record model.

hooverlunch: Roger

PR #234, opened 2017-01-25
5594 new project

emitche:
- Add button on projects index to new project page
- Add form to new project page
- Allow new project to be saved
- Add division to project form

### DIFF
--- a/app/views/admin/basic_projects/_form.html.slim
+++ b/app/views/admin/basic_projects/_form.html.slim
@@ -14,7 +14,7 @@
     .view-element = @project.try(:name)
     = f.input_field :name
 
-  = f.input :division_id
+  = f.input :division
     .view-element = @project.division.try(:name)
     = f.input_field :division_id, collection: @division_choices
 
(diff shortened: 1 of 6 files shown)


app/views/admin/basic_projects/_form.html.slim
```diff
@@ -14,19 +14,21 @@
     .view-element = @project.try(:name)
     = f.input_field :name
 
+  = f.input :division
+    .view-element = @project.division.try(:name)
```
hooverlunch: Good use of `try` here. 

I'm not sure why it's used on the existing code for `:name`, `:status`, and `:signing_date` though. We can assume `@project` won't be nil. Could you please refactor those.

app/views/admin/basic_projects/index.html.slim
```diff
@@ -1,4 +1,7 @@
-- content_for(:title, "Projects")
+- content_for(:title, t("basic_project.project", count: 2))
```
hooverlunch: Oh nice catch!

PR #235, opened 2017-01-26
5633 coop contact empty

X0nic:


app/controllers/admin/organizations_controller.rb
```diff
@@ -83,19 +79,23 @@ def destroy
   private
 
   def organization_params
-    params.require(:organization).permit(
+    # Not sure why params.permit wasn't honoring this field for the native simple_form array
```
X0nic: This looks like a rails or simple_form bug to me. Probably because of the strange pluralization. It's just a theory, though.

hooverlunch: Since it's an array you need to do `person_ids: []`

https://github.com/rails/strong_parameters#permitted-scalar-values

app/models/organization.rb
```diff
@@ -65,4 +67,13 @@ def active_loans
     loans.where(status_value: Loan::STATUS_ACTIVE_VALUE)
   end
 
+  private
+
+  def primary_contact_is_member
+    return if primary_contact.blank?
+    return if person_ids.include?(primary_contact_id)
+
+    errors.add(:primary_contact, I18n.t('organization.invalid_primary_contact'))
```
hooverlunch: Please use the proper Rails I18n conventions for this.

I believe it would be `activerecord.errors.models.organization.attributes.primary_contact.invalid`.

Then you just do `errors.add(:primary_contact, :invalid)`.

spec/models/organization_spec.rb
```diff
@@ -6,4 +6,14 @@
...
+
+  it 'errors when primary_contact does not belong to organization' do
+    contact = create(:person)
+    expect{ create(:organization, primary_contact: contact) }.to raise_error(ActiveRecord::RecordInvalid)
+  end
+
+  it 'primary_contact can be added when also part of organization' do
+    contact = create(:person)
+    create(:organization, primary_contact: contact, people: [contact])
+  end
```
hooverlunch: Nice specs. You could DRY them up a bit by:

1. Putting them in a `describe 'primary_contact'` block
1. Moving `contact` to a `let` block inside the `describe`

I would also please put the second `create` into an `expect { ... }.not_to raise_error` to clarify intent.

app/controllers/admin/organizations_controller.rb
X0nic: Gah, thats right.

PR #236, opened 2017-01-26
5572 Override default loans and members filters

X0nic:


app/helpers/application_helper.rb
```diff
@@ -72,4 +72,20 @@ def content_tag_if(condition, *args, &block)
       capture(&block)
     end
   end
+
+  def admin_loans_path(*args)
```
X0nic: Maybe there is a better way to override the params for these routes?

config/routes.rb
```diff
@@ -81,5 +81,5 @@
 
   get '/test' => 'static_pages#test'
 
-  root to: redirect('/admin/loans')
+  root to: redirect(path: '/admin/loans', params: { loans: { f: { status_value: ['active'] } } })
```
X0nic: I don't think we can access the url_helpers here. Can you think of a way to reduce the duplication with `admin_loans_path`?

app/helpers/application_helper.rb
```diff
@@ -72,4 +72,20 @@ def content_tag_if(condition, *args, &block)
...
+    # Default to active loans, unless overriden
+    options = { loans: { f: { status_value: ['active'] } } }
+    options = args.pop if args.last.is_a? Hash
```
hooverlunch: Won't this not work right if someone provides an unrelated option?

Shouldn't you first extract the options hash and then set the default value unless it's there already?

X0nic: I guess that depends on what we want the behavior to be.

Right now there is just one link that uses options:

red_emma_s__llc_-_madeline

With the change, it will show take you to loans, but default to only active loans. If we want to show all loans, we run into a similar problem with wice_grid where you can't choose to show all statues.

hooverlunch: I don't quite follow. For that link, couldn't you do `admin_loans_path(loans: {f: {status_value: ""}})` and the helper would see that that key is defined in the options and _not_ change it to `active`?

I am worried about us going back and forth on this too many times. We need to wrap this up. If my suggestion here doesn't work, please reach out and let's pair on this until it's done.

hooverlunch: Also note that I think the convention on this project is to only put spaces inside curlies if it's a Proc.

PR #237, opened 2017-01-27
5596 project timeline

emitche:
- Convert timeline list and timeline table to be used on all projects
- Show both timelines in basic project show page
- Replace loan_id with project_id


app/views/admin/basic_projects/show.html.slim
```diff
@@ -26,7 +25,32 @@ section.details
 
     = render 'form'
 
+section.block id="timeline-table"
+  h3 Timeline - Table
+  = render "admin/timeline/main"
+
+div id="timeline-list"
```
emitche: How content is set up is temporary. Eventually each of these will be a separate page that has tabs at the top taking the user to different pages for a basic project.

spec/factories/repayments.rb
```diff
@@ -5,7 +5,7 @@
     amount_paid { amount_due }
     interest_since_last_payment { rand(1..amount_due) }
     date_due { Faker::Date.between(Date.civil(2004, 01, 01), Date.today) }
-    loan_id { create(:loan).id }
+    project_id { create(:loan).id }
```
emitche: Should I be adjusting these specs?

spec/models/loan_spec.rb
```diff
@@ -102,11 +102,11 @@
       let!(:loan) { create(:loan) }
       let!(:project_steps) do
         project_steps = []
-        project_steps << create_list(:project_step, 2, :past, :completed, :with_logs, :for_loan, loan_id: loan.id)
-        project_steps << create_list(:project_step, 8, :past, :for_loan, loan_id: loan.id)
-        project_steps << create_list(:project_step, 2, :future, :for_loan, loan_id: loan.id)
-        project_steps << create_list(:project_step, 2, :past, :completed, :for_loan, loan_id: loan.id)
-        project_steps << create_list(:project_step, 2, :past, :with_logs, :for_loan, loan_id: loan.id)
+        project_steps << create_list(:project_step, 2, :past, :completed, :with_logs, :for_loan, project_id: loan.id)
```
emitche: Should I be adjusting these specs?

app/views/admin/timeline/_step.html.slim
```diff
@@ -48,4 +48,4 @@
     span.dropdown
       i.fa.fa-cog data-toggle="dropdown" aria-hidden="true"
 
-  / Please update timeline_table_step_column_count in LoansHelper if adding/removing columns here.
+  / Please update timeline_table_step_column_count in AdminHelper if adding/removing columns here.
```
hooverlunch: Yes!

config/routes.rb
```diff
@@ -9,7 +9,14 @@
   end
 
   namespace :admin do
-    resources :basic_projects, path: 'projects'
+    resources :basic_projects do
```
hooverlunch: This will conflict on merge because I changed the path thing. Can you please add `path: 'basic-projects'` and that should fix it.

spec/factories/loans.rb
```diff
@@ -121,8 +121,8 @@
 
     trait :with_repayments do
       after(:create) do |loan|
-        paid = create_list(:repayment, num_repayments = 2, :paid, loan_id: loan.id)
-        unpaid = create_list(:repayment, num_repayments = 3, loan_id: loan.id)
+        paid = create_list(:repayment, num_repayments = 2, :paid, project_id: loan.id)
+        unpaid = create_list(:repayment, num_repayments = 3, project_id: loan.id)
```
hooverlunch: I think these would stay as loan because repayments seem tied to loans but not projects.

spec/factories/repayments.rb
hooverlunch: See above.

Were the specs not passing when you started the branch? If so then it's someone else's job to fix them. I'll check with Fuzzy.

app/views/admin/loans/_tabs.html.slim
```diff
@@ -17,12 +17,12 @@ div class="tab-content"
         // Questionnaires are loaded dynamically by view
 
   div role="tabpanel" class="tab-pane" id="timeline-table"
-    = render "admin/loans/timeline/main", loan: @loan
+    = render "admin/timeline/main"
```
emitche: Yes.

app/views/admin/loans/show.html.slim
```diff
@@ -14,7 +14,8 @@ javascript:
     new MS.Views.LoanTabView({
       el: '.loan-content',
       calendarEventsUrl: #{json @calendar_events_url},
-      loanId: #{json @loan.id},
-      locale: "#{I18n.locale}"
+      projectId: #{json @project.id},
+      locale: "#{I18n.locale}",
+      urlComponent: "loans"
```
emitche: Good point.

spec/factories/loans.rb
emitche: Ok.

emitche:
This looks good, @hooverlunch. The changes are understandable. Thanks for the code review. I will not touch the specs. I will change that back to loan_id/whatever is in develop. Thanks!

spec/factories/repayments.rb
linesarefuzzy: Yeah, repayments were an incomplete feature from the old Cofunder front end. Let's leave them alone for now.

spec/models/loan_spec.rb
linesarefuzzy: I think so in this case.

emitche: Yes, a new step expects a project_id, not a loan_id.

app/assets/javascripts/backbone/views/timeline_table_view.coffee
```diff
@@ -5,8 +5,14 @@ class MS.Views.TimelineTableView extends Backbone.View
 
   initialize: (options) ->
     new MS.Views.AutoLoadingIndicatorView()
-    @loanId = options.loanId
-    @groupModal = new MS.Views.ProjectGroupModalView(loanId: @loanId, success: @refresh.bind(@))
+    @projectId = options.projectId
+
+    if options.projectType == 'BasicProject'
```
emitche: Written in this syntax because the below breaks, with or without the parentheses. The below returns `true` or `false` instead of the expected keys.
`@urlComponent = (options.projectType == 'BasicProject') ? 'basic-projects' : 'loans'`

The same syntax is used in another file.

app/assets/javascripts/backbone/views/timeline_view.coffee
```diff
@@ -3,7 +3,12 @@ class MS.Views.TimelineView extends Backbone.View
   el: 'section.timeline'
 
   initialize: (options) ->
-    @loanId = options.loanId
+    @projectId = options.projectId
+
+    if options.projectType == 'BasicProject'
```
emitche: See earlier comment in timeline_table_view.

app/assets/javascripts/backbone/views/timeline_table_view.coffee
hooverlunch: Yes, coffeescript doesn't like the `? :` operator. You can use `if` `else` on one line instead but this if fine.

PR #240, opened 2017-01-31
5597 project calendar

emitche:
- Add calendar to basic project page
- Revise calendar to work with Loans and BasicProjects
- Show basic project calendar events inside general calendar


app/assets/javascripts/backbone/views/calendar_view.js.coffee
```diff
@@ -6,6 +6,7 @@ class MS.Views.CalendarView extends Backbone.View
     # Initialize calendar
     @$calendar = @$('#calendar')
     @stepModal = params.stepModal
+    @urlComponent = params.urlComponent
```
emitche: @hooverlunch: Throughout Loans and BasicProjects, we will need to specify `admin/loans/some-path` and `admin/basic-projects/some-path`. Do you have a preferred method?

The method used to pass a string to the JavaScript to define the desired project-specific portion of the path.

An alternative:
```
if (params.projectType == 'BasicProject')
  @urlComponent = 'basic-projects'
else
  @urlComponent = 'loans'
```

Still another alternative is to have an admin/projects path that can do the necessary work to decide which type of project it is and forward to the appropriate controller. This method feels the cleanest.

app/controllers/concerns/project_concern.rb
```diff
@@ -0,0 +1,13 @@
+module ProjectConcern
+  extend ActiveSupport::Concern
+
+  private
+
+  def change_date(project)
```
emitche: @linesarefuzzy suggested using an inheritance pattern instead of a concern for certain methods like this. However, @hooverlunch, we appear to be going the way of concerns based on a previous code review.

linesarefuzzy: I thought inheritance was the answer so we could have a single project-type-agnostic route for change_date, as you suggested above. I think that would remove the need for the `urlComponent` param. There's no reason functions like change_date need to know the project type—they should function the same either way.

linesarefuzzy: That said, this method works fine too, so whatever @hooverlunch says is fine with me.

emitche: I like inheritance. I will see what @hooverlunch says and we will go with that. Thanks, @linesarefuzzy.

app/assets/javascripts/backbone/views/calendar_view.js.coffee
hooverlunch: For shared stuff, let's do a separate controller as Fuzzy suggests.

app/assets/javascripts/backbone/views/calendar_view.js.coffee
```diff
@@ -62,8 +63,8 @@ class MS.Views.CalendarView extends Backbone.View
       # We use a 1ms timeout so that fullCalendar can finish drawing the event in the new calendar cell.
       setTimeout =>
         if confirm(I18n.t("loan.move_date_confirm.body"))
-          loanId = @$el.find('.loan-calendar').data('loan-id')
-          $.post "/admin/loans/#{loanId}/change_date",
+          projectId = @$el.find('#calendar').data('project-id')
+          $.post "/admin/#{@urlComponent}/#{projectId}/change_date",
```
hooverlunch: So all the `@urlComponent`s should be replaceable by `projects`

app/controllers/admin/basic_projects_controller.rb
```diff
@@ -22,54 +22,64 @@ def index
...
     end
   end
 
+  def change_date
+    @basic_project = @basic_project = BasicProject.find(params[:id])
+    authorize @basic_project, :update?
+    attrib = params[:which_date] == "project_start" ? :signing_date : :end_date
+    @basic_project.update_attributes(attrib => params[:new_date])
+    render nothing: true
+  end
```
hooverlunch: How come this got added here as well as in the concern? Should be able to remove it once we have the separate controller.

app/controllers/concerns/project_concern.rb
hooverlunch: @linesarefuzzy You are correct I think. Good thinking. My mistake. Fortunately I think there was not much wasted work. We still needed to abstract, just to a different place. So all the `@urlComponent`s can I think just become `projects`.

app/models/project.rb
```diff
@@ -104,4 +104,17 @@ def agent_names
   def display_agent_names
     agent_names.join(', ')
   end
+
+  def calendar_events
+    CalendarEvent.build_for(self)
+  end
+
+  def project_events(order_by="Completed IS NULL OR Completed = '0000-00-00', Completed, Date")
```
hooverlunch: Can we add spaces around the `=` sign here please.

PR #241, opened 2017-02-01
5534: disallow blank step types, and change form to default to checkin

X0nic:


app/controllers/admin/project_steps_controller.rb
```diff
@@ -5,12 +5,13 @@ class Admin::ProjectStepsController < Admin::AdminController
   def new
     @project = Project.find(params[:project_id])
     @step = ProjectStep.new(project: @project, scheduled_start_date: params[:date],
-      parent_id: params[:parent_id], schedule_parent_id: params[:schedule_parent_id])
+      parent_id: params[:parent_id], schedule_parent_id: params[:schedule_parent_id],
+      step_type_value: 'checkin')
```
X0nic: Made the suggested change here, but I still have a hardcoded value. Is there a better way?

hooverlunch: No this is fine.

PR #242, opened 2017-02-02
5788 allow clearing of schedule_parent

X0nic:


spec/models/project_step_spec.rb
```diff
@@ -48,6 +48,16 @@
...
+  it 'scheduled_start_date can be set to blank string when scheduled_start_date of schedule_parent is nil' do
+    parent = create(:project_step, scheduled_start_date: nil)
+    step = create(:project_step, schedule_parent: parent)
+
+    step.scheduled_start_date = ''
+    step.save
+
+    expect(step.scheduled_start_date).to be_nil
+  end
+
```
hooverlunch: It's not clear to me how this spec represents the reported bug. Should we not have a spec that tests that you can set _schedule_parent_ to nil/blank? @X0nic

hooverlunch:
Looks great, thanks.

PR #245, opened 2017-02-12
5808 quickbooks frontend

X0nic:


config/initializers/quickbooks.rb
```diff
@@ -0,0 +1,13 @@
+unless Rails.env == 'test'
```
X0nic: I think it would be best if we could pair @linesarefuzzy on a way to clean this up a bit.

app/controllers/admin/quickbooks_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+    authorize :quickbooks, :disconnect?
+
+    Division.root.quickbooks_disconnect
+
+    redirect_to admin_settings_path, notice: 'Your QuickBooks account has been successfully disconnected.'
+  end
+
+  private
+
+  def qb_oauth_consumer
```
X0nic: Seemed like overkill to extract this into a class, so I left it here.

app/models/accounting/quickbooks/consumer.rb
```diff
@@ -0,0 +1,29 @@
...
+      def verify_access_token(qb_request_token:, oauth_verifier:)
+        qb_request_token.get_access_token(oauth_verifier: oauth_verifier)
+      end
+
+      def access_token(token:, secret:)
+        OAuth::AccessToken.new(@oauth_consumer, token, secret)
+      end
+    end
+  end
+end
```
hooverlunch: This looks good as its own class!

app/models/division.rb
```diff
@@ -114,4 +114,20 @@ def locale_names
...
+    self.class.connector.connected?
+  end
+
+  def quickbooks_connect(access_token:, params:)
+    self.class.connector.connect(access_token: access_token, params: params)
+  end
+
+  def quickbooks_disconnect
+    self.class.connector.disconnect
+  end
```
hooverlunch: Having all these methods prefixed with `quickbooks_` is a code smell IMO. It says that there is a bunch of functionality that should be in its own class...

BUT looking more closely really it seems like what these are are just delegations to follow LoD, which is good!

So what I recommend is:

1. Making `connector` an instance method and renaming it to `quickbooks` (right now you're setting an instance var in a class method which I don't even know how that's working!)
1. Using `delegate ..., prefix: true` to setup your delegators. Then you can delete all these methods.

app/models/division.rb
```diff
@@ -114,4 +114,20 @@ def locale_names
...
+
+  def self.connector
+    @connector ||= Accounting::Quickbooks::Connector.new Division.root.custom_data
```
hooverlunch: I don't even know how this works right now. Seems like this should be a class var. But I'd rather avoid the class method stuff anyway unless it's needed for some reason I'm not seeing. See above.

hooverlunch: Also, `connector` is not a good method name because it doesn't reference quickbooks anywhere. Again, see above for a recommendation.

app/views/admin/settings/index.html.slim
```diff
@@ -0,0 +1,27 @@
+- content_for(:title, 'Settings')
```
hooverlunch: It makes sense to have a settings page and it seems like a lot of the stuff under the Division page (e.g. colors, currency, locales, notify on new logs) should move here. @rgalusca what do you think? Obviously this would be a separate issue...

app/views/admin/settings/index.html.slim
```diff
@@ -0,0 +1,27 @@
...
+    .form-group
+      label.col-sm-2 Quickbooks Status:
+      - if Division.root && Division.root.quickbooks_connected?
+        .col-sm-10 Connected
+      - else
+        .col-sm-10 Not Connected
+
+    - if Division.root && Division.root.quickbooks_connected?
+      .form-group
+        label.col-sm-2 Disconnect from  Quickbooks:
```
hooverlunch: Extra space here.

app/views/admin/settings/index.html.slim
```diff
@@ -0,0 +1,27 @@
...
+        .col-sm-10 
+          a href="#{disconnect_admin_quickbooks_path}" Disconnect
+    - else
+      .form-group
+        label.col-sm-2 Connect to Quickbooks:
+        .col-sm-10
+          / this will display a button that the user clicks to start the flow
```
hooverlunch: Nice comment!

app/views/admin/settings/index.html.slim
```diff
@@ -0,0 +1,27 @@
...
+          ipp:connectToIntuit
```
hooverlunch: This page will surely end up needing to be refactored as a form later on. I guess it's ok for now like this.

config/initializers/quickbooks.rb
```diff
@@ -0,0 +1 @@
+Quickbooks.sandbox_mode = Rails.env.production?
```
hooverlunch: Umm, shouldn't there be a `!` here?

spec/models/accounting/quickbooks/connector_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+      expect(connector.renewable?).to be_falsey
+    end
+    it 'should not be expired' do
+      expect(connector.expired?).to be_falsey
+    end
+  end
+
+  context 'with new token' do
+    subject(:connector) do
+      Accounting::Quickbooks::Connector.new(quickbooks: { token: valid_token, secret: valid_secret, realm_id: valid_realm_id, token_expires_at: 180.days.from_now.utc })
```
hooverlunch: The fact that `quickbooks:` needs to be passed here should be an immediate code smell. Do you see why?

spec/models/accounting/quickbooks/connector_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    end
+
+    it 'should be connected' do
+      expect(connector.connected?).to be_truthy
+    end
+    it 'should not be renewable' do
+      expect(connector.renewable?).to be_falsey
+    end
+    it 'should not be expired' do
+      expect(connector.expired?).to be_falsey
```
hooverlunch: Is there a reason you're using `be_truthy` instead of `be true`? They always return booleans right? Why not be more specific?

spec/models/accounting/quickbooks/connector_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    it_behaves_like 'no token is present'
+  end
+
+  context 'with invalid token' do
+    subject(:connector) do
+      Accounting::Quickbooks::Connector.new(quickbooks: { token: invalid_token, secret: valid_secret, realm_id: valid_realm_id, token_expires_at: 180.days.from_now.utc })
+    end
+
+    it_behaves_like 'no token is present'
```
hooverlunch: Nice abstraction.

spec/models/accounting/quickbooks/connector_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  context 'with invalid realm id' do
+    subject(:connector) do
+      Accounting::Quickbooks::Connector.new(quickbooks: { token: valid_token, secret: valid_secret, realm_id: invalid_realm_id, token_expires_at: 180.days.from_now.utc })
+    end
+
+    it_behaves_like 'no token is present'
+  end
+end
```
hooverlunch: There should be a feature spec for the new settings page right? Testing connection will be hard and I don't think we should go down that road now unless it's easier than I'm thinking, but you could add a fake token and test disconnection, right?

app/models/accounting/quickbooks/connector.rb
```diff
@@ -0,0 +1,71 @@
+module Accounting
+  module Quickbooks
+    class Connector
```
hooverlunch: Seems like this should be called `Connection`? At first I thought it actually did some calling out to the QB api but it actually doesn't? Really it just stores the tokens and etc.?

app/controllers/admin/quickbooks_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    authorize :quickbooks, :authenticate?
+
+    redirect_to("https://appcenter.intuit.com/Connect/Begin?oauth_token=#{qb_request_token}")
+  end
+
+  def oauth_callback
+    authorize :quickbooks, :oauth_callback?
+    @header_disabled = true
+
+    Division.root.quickbooks_connect(access_token: qb_access_token, params: params)
```
hooverlunch: This method name makes it look like the method is doing the connection. But really it's already connected and you're just storing the data. Maybe just call it `quickbooks_save`. And instead of disconnect below you could call it `forget`.

config/initializers/quickbooks.rb
X0nic: Um wow, yeah. Good catch!

app/views/admin/settings/index.html.slim
rgalusca: @hooverlunch, 
Not sure if settings would be the right place for division-related functionalities. It would render the Manage Divisions page obsolete and I personally like the way it is right now. But, I can see that, at some point, we might need to add new stuff to Settings or even rethink the Manage Divisions page altogether. For the moment though, this is not going to be a priority.

PR #247, opened 2017-02-14
5885 bug fix status filter timeline table

emitche:
- Fix status dropdown filters in timeline table header

### DIFF
--- a/app/assets/stylesheets/admin/loans/timeline/_table.scss
+++ b/app/assets/stylesheets/admin/loans/timeline/_table.scss
@@ -17,6 +17,10 @@
     }
   }
 
+  table {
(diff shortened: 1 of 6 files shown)


emitche:
This may not be the best approach, so I welcome feedback. I wish I would have paired at the start on this.

The filters do not belong to a single attribute on a project step. Therefore, I created a new attribute. If you have a different approach, let me know.

Thanks.

emitche:
I paired with @hooverlunch  and he will look into this issue further. Recreating the wheel to get something working that may have previously worked is not a good approach.

emitche:
The approach to fix this will be different.

PR #249, opened 2017-02-15
5911 create dashboard page

emitche:
- Create dashboard page
- Create link in menu

### DIFF
--- a/app/assets/stylesheets/admin/dashboard/_dashboard.scss
+++ b/app/assets/stylesheets/admin/dashboard/_dashboard.scss
@@ -4,11 +4,9 @@
       margin-bottom: 1em;
     }
 
-    section {
-      section {
-        h2 {
-          font-size: 1.5em;
-        }
+    .section {
+      h2 {
+        font-size: 1.5em;
       }
     }
   }
--- a/app/assets/stylesheets/admin/dashboard/_dashboard.scss
+++ b/app/assets/stylesheets/admin/dashboard/_dashboard.scss
@@ -1,0 +1,15 @@
+.admin-dashboard {
+  .content {
+    .row {
+      margin-bottom: 1em;
+    }
+
+    section {
+      section {
+        h2 {
+          font-size: 1.5em;
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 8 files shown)


app/controllers/admin/dashboard_controller.rb
```diff
@@ -0,0 +1,7 @@
+class Admin::DashboardController < Admin::AdminController
+  def dashboard
+    authorize Project
```
emitche: Project will be used by 3 out of the 4 features that are a part of the dashboard. But I can remove this until it is needed, if you think that is best.

app/controllers/admin/dashboard_controller.rb
```diff
@@ -0,0 +1,7 @@
+class Admin::DashboardController < Admin::AdminController
+  def dashboard
```
emitche: When pairing with linesarefuzzy, Fuzzy requested that this action not be index or show, but instead a custom action named dashboard.

emitche:
Page title is missing and needs to be added.

app/assets/stylesheets/admin/dashboard/_dashboard.scss
```diff
@@ -0,0 +1,15 @@
+.admin-dashboard {
+  .content {
+    .row {
+      margin-bottom: 1em;
+    }
+
+    section {
+      section {
```
hooverlunch: This seems like an odd pattern and perhaps a bit brittle. Can you just use a class on the inner section perhaps?

app/controllers/admin/dashboard_controller.rb
```diff
@@ -0,0 +1,7 @@
...
+    authorize Project
+    authorize Person
```
hooverlunch: We should talk about this authz stuff. Not quite right.

app/controllers/admin/dashboard_controller.rb
hooverlunch: Hmm, I'm curious as to why...

emitche:
Per @rgalusca, the heading should be in line with the beginning of the menu.

emitche:
The page can be accessed at admin/dashboard. This feature is just the bones for the dashboard.

PR #250, opened 2017-02-15
5876 export tables to csv

emitche:
- Bug fix: Ensure CSV tables export from index grids

### DIFF
--- a/app/controllers/admin/basic_projects_controller.rb
+++ b/app/controllers/admin/basic_projects_controller.rb
@@ -15,7 +15,7 @@ class Admin::BasicProjectsController < Admin::ProjectsController
 
     @csv_mode = true
 
-    export_grid_if_requested do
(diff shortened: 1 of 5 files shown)


app/controllers/admin/basic_projects_controller.rb
```diff
@@ -15,7 +15,7 @@ def index
 
     @csv_mode = true
 
-    export_grid_if_requested do
+    export_grid_if_requested('basic_projects': 'basic_projects_grid_definition') do
```
emitche: The first part does not need quotes `basic_projects: '..'` is enough.

hooverlunch: 👍

hooverlunch: This is kind of yucky boilerplate but I guess wice_grid needs it and it's probably not worth creating a helper method to simplify it.

PR #251, opened 2017-02-16
5807 transaction list

X0nic:
This PR was built on top of the code in 5808. Use this link to see the actual [diff](https://github.com/sassafrastech/madeline_system/compare/5808_quickbooks_frontend...5807_transaction_list).


app/controllers/admin/transactions_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Admin::TransactionsController < Admin::AdminController
+
+  def index
+    authorize :transaction, :index?
+
+    auth_details = {access_token: Division.connector.access_token, company_id: Division.connector.realm_id}
+
+    # This is temporary code until the proper transaction wrappers are created.
+    @transaction_list = %w(JournalEntry Deposit Purchase).map do |transaction_type|
```
X0nic: This could be much better, but it is going to be cleaned up very soon.

app/controllers/admin/transactions_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Admin::TransactionsController < Admin::AdminController
+
+  def index
+    authorize :transaction, :index?
+
+    auth_details = {access_token: Division.connector.access_token, company_id: Division.connector.realm_id}
```
hooverlunch: I suggest adding a private helper method `qb_connection` that memoizes `Division.root.qb_connection`. Then you can use that throughout.

app/views/admin/transactions/_tbody.html.slim
```diff
@@ -0,0 +1,7 @@
+- @transaction_list.each do |transaction|
+  tr
+    td= transaction[:id]
+    td= transaction[:txn_date]
+    td= transaction[:type]
+    td= transaction[:total]
+    td= transaction[:private_note]
```
hooverlunch: We usually put spaces around the = I think. I know this is temp though.

app/views/admin/transactions/index.html.slim
```diff
@@ -0,0 +1,6 @@
+- content_for(:title, 'Transactions')
+
+table.table
+  = render 'thead'
+  tbody
+    = render 'tbody'
```
hooverlunch: This seems a bit premature to separate the head and body. Each partial also adds a bit to the load time I find. Can we combine these back into one file unless there is some need for the abstraction?

app/controllers/admin/transactions_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    @transaction_list = %w(JournalEntry Deposit Purchase).map do |transaction_type|
+      transactions = Quickbooks::Service.const_get(transaction_type).new(auth_details).query
+      transactions.map do |t|
+        {
+          type: transaction_type,
+          id: t.id,
+          txn_date: t.txn_date,
+          total: t.total,
+          private_note: t.private_note,
+        }
```
hooverlunch: This will of course turn into a class instantiation later, etc.

app/policies/transaction_policy.rb
```diff
@@ -0,0 +1,5 @@
+class TransactionPolicy < ApplicationPolicy
```
hooverlunch: Should be namespaced under Accounting

app/controllers/admin/transactions_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Admin::TransactionsController < Admin::AdminController
```
hooverlunch: Should be namespaced as `Admin::Accounting::`. I think we will end up with several controllers here and it could get awkward.

app/controllers/admin/quickbooks_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Admin::QuickbooksController < Admin::AdminController
```
hooverlunch: Sorry I missed this but this should also be namespaced under `Admin::Accounting`.

app/policies/quickbooks_policy.rb
```diff
@@ -0,0 +1,13 @@
+class QuickbooksPolicy < ApplicationPolicy
```
hooverlunch: Also should be namespaced under `Accounting`. `Admin` is not required here as that's for controllers only.

config/routes.rb
```diff
@@ -61,6 +61,18 @@
...
+
+    resources :quickbooks do
+      collection do
+        get :authenticate
+        get :oauth_callback
+        get :disconnect
+      end
+    end
+
+    resources :transactions
```
hooverlunch: Routes should also be namespaced, so /admin/accounting/transactions and /admin/accounting/quickbooks.

spec/models/accounting/quickbooks/connector_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  context 'with invalid realm id' do
+    subject(:connector) do
+      Accounting::Quickbooks::Connector.new(quickbooks: { token: valid_token, secret: valid_secret, realm_id: invalid_realm_id, token_expires_at: 180.days.from_now.utc })
+    end
+
+    it_behaves_like 'no token is present'
+  end
+end
```
hooverlunch: We should add a feature spec for the transactions page. I know the content is temporary but the page/route is not, right? Better to add it now and adjust it later b/c we'll be reminded to do so when it fails. If we wait till later we might forget.

.env.example
```diff
@@ -1,2 +1,4 @@
 export MADELINE_EMAIL_FROM=madeline@theworkingworld.org
 export MADELINE_HOSTNAME=madeline.theworkingworld.org
+export QB_OAUTH_CONSUMER_KEY=
+export QB_OAUTH_CONSUMER_SECRET=
```
hooverlunch: Another thing just occurred to me -- have we added anything to the Readme for QB stuff? Should probably be something about where to get keys, and what you can and can't do with out adding a key.

X0nic:
I have made the suggested code changes. I still need to update the readme, and add a couple feature specs. No need to hold up this PR for that, though. I will do a small followup PR.

PR #252, opened 2017-02-16
3746 dashboard upcoming events

emitche:
- Add calendar of upcoming events for current user


app/assets/javascripts/backbone/views/dashboard_calendar_view.js.coffee
```diff
@@ -0,0 +1,49 @@
...
+      firstDay: cal_start
+
+    @renderLegend()
+
+  eventRender: (calEvent) -> calEvent.html
+
+  loading: (isLoading) ->
+    MS.loadingIndicator[if isLoading then 'show' else 'hide']()
+
+  renderLegend: (e) ->
```
emitche: We are not using legend so therefore may be able to remove this, so long as a legend is not desired.

app/views/admin/dashboard/dashboard.html.slim
```diff
@@ -0,0 +1,14 @@
+section.dashboard
+  h1 = @person.name
+
+  section.block.upcoming-events.calendar
+    h2 Upcoming Steps
```
emitche: Use translation here.

app/assets/javascripts/backbone/views/dashboard_calendar_view.js.coffee
hooverlunch: This can live in the parent class and if you need it here later you can use it.

app/assets/javascripts/backbone/views/dashboard_calendar_view.js.coffee
```diff
@@ -0,0 +1,49 @@
+class MS.Views.DashboardCalendarView extends Backbone.View
```
hooverlunch: There is a fair amount of duplication here with the other calendar view. Perfect time to use inheritance. Move the methods that you want to use in both into the super class. Superclass (call it CalendarView?) extends Backbone.View. Then the child classes (TimelineCalendarView? and DashboardCalendarView) extend CalendarView.

hooverlunch: We can pair on this if you need to.

app/controllers/admin/calendar_events_controller.rb
```diff
@@ -8,6 +8,16 @@ def index
       project = Project.find(params[:project_id])
       authorize project, :show?
       project_filter = {id: project.id}
+    elsif params[:person_id]
+      person = Person.find(params[:person_id])
+      authorize person
```
hooverlunch: So as we discussed earlier, this is really checking if the current user can `index` `person` which is not what you want.

You really only need a scope here because any user should be able to get their own calendar events, so I would do `skip_authorization`. Or you could add a CalendarEventsPolicy I guess..

app/controllers/admin/calendar_events_controller.rb
```diff
@@ -8,6 +8,16 @@ def index
...
+      authorize Project
+
+      project_ids = []
+      person.agent_projects.each do |project|
+        project_ids.push(project.id)
+      end
+      project_filter = {id: project_ids}
```
hooverlunch: Here is a hint for how to simplify the above and incorporate policy_scope as we discussed:

```
project_filter = {id: policy_scope(person.agent_projects).pluck(:id)}
```

Please please please let me know if how this works this isn't clear. You may need to look up `pluck`. Please understand why using `pluck` is better than using `.map(&:id)` in this case. Feel free to use some mentoring time here.

hooverlunch: Overall, though, nice job on incorporating this new functionality into the controller!

app/views/admin/dashboard/dashboard.html.slim
hooverlunch: 👍

app/views/admin/dashboard/dashboard.html.slim
```diff
@@ -0,0 +1,14 @@
...
+    div#calendar
+
+javascript:
+  $(function(){
+    MS.DashboardCalendarView = new MS.Views.DashboardCalendarView({
```
hooverlunch: I don't think you need to assign this to anything here. We do with the other CalendarView for separate reasons.

app/assets/javascripts/backbone/views/dashboard_calendar_view.js.coffee
emitche: :D I understand completely and that is brilliant. Thanks!

app/controllers/admin/calendar_events_controller.rb
emitche: I may just use skip_authorization.

app/controllers/admin/calendar_events_controller.rb
emitche: I'll check in with you on this.

emitche:
@hooverlunch Please see comment in Redmine.

@rgalusca Please see note in Redmine.

PR #254, opened 2017-02-24
5829 Adding initial transaction model with QBO data

X0nic:


lib/tasks/migrate_data.rake
```diff
@@ -76,4 +76,17 @@ namespace :tww do
       end
     end
   end
+
+  desc "migrate some test data from quickbooks"
```
X0nic: A helper to pull over some QB transactions for testing. Otherwise the transaction list will be empty.

app/models/accounting/transaction.rb
```diff
@@ -0,0 +1,35 @@
...
+  attr_writer :qb_object
+
+  delegate :txn_date, :total, :private_note, to: :qb_object
+
+  def qb_object
+    raise 'qb_object not already set' if @qb_object.blank?
+    @qb_object
+  end
+
+  def self.where(**args)
```
X0nic: @linesarefuzzy and I decided during our pairing session, it would make most sense to override `where` here.

app/models/accounting/transaction.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+#  created_at          :datetime         not null
+#  id                  :integer          not null, primary key
+#  qb_transaction_id   :string           not null
+#  qb_transaction_type :string           not null
+#  updated_at          :datetime         not null
+#
+
+class Accounting::Transaction < ActiveRecord::Base
+  attr_writer :qb_object
```
X0nic: I would normally have this private but because it needs to be set externally, I have I public.

app/models/accounting/quickbooks/fetcher.rb
```diff
@@ -0,0 +1,53 @@
+module Accounting
+  module Quickbooks
+    class Fetcher
```
hooverlunch: Nice class name! Can we please get a general comment describing what the class does.

app/models/accounting/quickbooks/fetcher.rb
```diff
@@ -0,0 +1,53 @@
...
+        end.flatten
+      end
+
+      def populate(qb_objects)
+        @relation.each do |transaction|
+          qb = qb_objects.find { |qbo| qbo.id.to_s == transaction.qb_transaction_id }
+          transaction.qb_object = qb
+        end
+        @relation
+      end
```
hooverlunch: Can the above two methods please be private? I was confused because I couldn't understand why all three would need to be used by outside objects, but it seems only fetch does, right?

app/models/accounting/quickbooks/fetcher.rb
```diff
@@ -0,0 +1,53 @@
...
+      def auth_details
+        { access_token: qb_connection.access_token, company_id: qb_connection.realm_id }
+      end
+
+      def qb_connection
+        @qb_connection ||= Division.root.qb_connection
+      end
+    end
+  end
+end
```
hooverlunch: Overall, great job on this! Wow, very nice.

app/models/accounting/transaction.rb
hooverlunch: This is fine.

app/models/accounting/transaction.rb
```diff
@@ -0,0 +1,37 @@
...
+    Accounting::Quickbooks::Fetcher.new(transactions).fetch
+  end
+
+  # We need to override where to ensure the qb_object data gets populated when queried.
+  def self.all
+    transactions = super
+    return transactions if transactions.count < 1
+
+    Accounting::Quickbooks::Fetcher.new(transactions).fetch
+  end
```
hooverlunch: I don't think overriding these methods is a good idea because it violates the principle of least surprise. They both will now return arrays instead of relations and so they are no longer chainable... Which is fine in itself, I don't see how you could avoid this, but...

Couldn't you add a new class method called `with_qb_objs` that handles the fetch and returns an array, and then do e.g. `Transaction.where(whatever).order(whatever).with_qb_objs`. You would just put a comment on with_qb_objs that says it returns an array and should be at the end of a chain.

Also, please remember to put class methods above instance methods.

app/models/accounting/quickbooks/fetcher.rb
hooverlunch: It appears this class has no test coverage. This is concerning. We need to have good coverage on this core stuff.

What I recommend:

* We only need to test `fetch`
* Create a few transactions with two different types (using a factory)
* stub the `service` method to return a test double
* stub the `query` method on the `service` double and expect it to be called with the appropriate query methods (should be two, right?), and return a double with `entries` set to some qb object doubles with the appropriate `id`s set
* Instantiate with Transaction.all and expect that `fetch` returns the objects with the appropriate qb object doubles attached

app/models/accounting/transaction.rb
hooverlunch: We need a test here too I think, for the with_qb_objs method. For this one you should stub the constructor of Fetcher and ensure it gets passed the right thing and then ensure fetch gets called.

db/migrate/20170221234419_create_accounting_transactions.rb
```diff
@@ -0,0 +1,10 @@
+class CreateAccountingTransactions < ActiveRecord::Migration
+  def change
+    create_table :accounting_transactions do |t|
+      t.string :qb_transaction_id, null: false
+      t.string :qb_transaction_type, null: false
```
hooverlunch: These should have indices, no? Separate, I think, not compound, because they are queried separately.

lib/tasks/migrate_data.rake
```diff
@@ -76,4 +76,17 @@ namespace :tww do
...
+  task :migrate_test_qbo => :environment do
+  qb_connection = Division.root.qb_connection
+  auth_details = { access_token: qb_connection.access_token, company_id: qb_connection.realm_id }
```
hooverlunch: These two lines should be indented one level.

X0nic:
@hooverlunch I added some tests for the fetcher. They are pretty basic since the implementation is going to be changing somewhat as we discussed. I ignored the specs for `with_qb_objs` as that probably won't exist soon.

spec/models/accounting/quickbooks/fetcher_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+    let(:transactions) do
+      create(:accounting_transaction, qb_transaction_id: 10, qb_transaction_type: 'Deposit')
+      create(:accounting_transaction, qb_transaction_id: 35, qb_transaction_type: 'Deposit')
+      create(:accounting_transaction, qb_transaction_id: 49, qb_transaction_type: 'JournalEntry')
+      Accounting::Transaction.all
+    end
+
+    it 'should fetch 3 records' do
+      expect(subject.fetch.count).to eql 3
```
hooverlunch: You should almost never need `eql` as it translates to `eql?` which is 'hash equality', which is rarely needed. `eq` is preferred.

spec/models/accounting/quickbooks/fetcher_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  context 'when multiple transaction types exist' do
+    before do
+      d_service = instance_double(Quickbooks::Service::Deposit)
+      j_service = instance_double(Quickbooks::Service::JournalEntry)
+
+      allow(subject).to receive(:service).with('Deposit').and_return(d_service)
+      allow(d_service).to receive(:query).and_return(deposits)
+
+      allow(subject).to receive(:service).with('JournalEntry').and_return(j_service)
+      allow(j_service).to receive(:query).and_return([instance_double(Quickbooks::Model::JournalEntry, id: 49)])
```
hooverlunch: Can you just put the `query`  returns inside the instance_double declaration above? 

e.g. `d_service = instance_double(Quickbooks::Service::Deposit, query: deposits)`

PR #257, opened 2017-02-27
5941 refactor loan tabs

X0nic:


app/views/admin/loans/_tabs.html.slim
```diff
@@ -7,9 +7,8 @@ ul class="nav nav-tabs nav-tabs-responsive hidden-print" role="tablist"
   = render "admin/loans/tab", identifier: "loan-calendar", label: t("calendar.calendar")
 
 div class="tab-content"
-  div role="tabpanel" class="tab-pane flush" id="details"
-    section.block.loan-fields data-container='details' class=(@loan.valid? ? 'show-view' : 'edit-view')
-      // Details are loaded dynamically by view
+  div role="tabpanel" class="tab-pane" id="details"
+    = render "admin/loans/details", loan: @loan
```
hooverlunch: OH! I am thinking now maybe this file just needs to be deleted? Hence all my confusion....

app/views/admin/loans/_timeline.html.slim
```diff
@@ -0,0 +1,11 @@
...
+  = render "admin/timeline/main", project: @loan
+
+javascript:
+  $(function() {
+    new MS.Views.TimelineTableView({
+      projectId: #{json @loan.id},
+      stepModal: new MS.Views.ProjectStepModalView()
+    });
+  });
+
```
hooverlunch: This file is now basically identical to `basic_projects/timeline`.

I get why we still need this file due to your nice loop thing but perhaps at least you could move the JS initialize into `timeline/main`? I don't see a need to duplicate it.

app/views/admin/loans/show.html.slim
```diff
@@ -4,7 +4,14 @@
...
   div.loan-content
-    = render "admin/loans/tabs"
+    ul class="nav nav-tabs nav-tabs-responsive hidden-print" role="tablist"
+      - @tabs.each do |tab|
+        li role="presentation" class==("active" if tab == @tab)
+          = link_to t("loan.tab.#{tab}"), admin_loan_tab_path(@loan.id, tab: tab)
+
+    div class="tab-content"
+      div class="tab-pane active" id=@tab
+        = render "admin/loans/#{@tab}"
```
hooverlunch: Very nice stuff here. So satisfying!

config/locales/en/loans.en.yml
```diff
@@ -69,6 +69,13 @@ en:
       frozen: "Frozen"
       liquidated: "Liquidated"
     end_date: "End Date"
+    tab:
+      details: "Details"
+      questions: "Questions"
+      timeline: "Timeline"
+      timeline_list: "Timeline - List"
+      logs: "Logs"
+      calendar: "Calendar"
```
hooverlunch: Nice that you've consolidated these but can you please remove the originals if they're no longer needed?

So, several instances of old stuff not getting cleaned out here. I bet at some point you thought to yourself "I need to clean this out later" but forgot? If so, do you use a checklist while you're developing? I use Evernote for this b/c it has nice checkboxes:

image

So just a tip in case it's useful.

X0nic: Thanks for the tip @hooverlunch I did forget to clean out some of the old stuff. I usually comment stuff out, then go back and delete it. Obviously, there are issues with that method when whole files need to be removed. ;)

X0nic:
This is mergable, but we should probably wait until I add some feature specs for the tabs.

hooverlunch:
Going to merge this now because it's holding stuff up and we can do the feature specs some other time.

PR #258, opened 2017-02-27
5937 alert users to save on questionnaire

emitche:
- Alert users when exiting a questionnaire to go to a different page
- Alert users when moving to a different questionnaire
- Applies to places where a user has made changes without saving


app/controllers/admin/loans_controller.rb
```diff
@@ -39,6 +39,7 @@ def show
     @steps = @loan.project_steps
     @calendar_events_url = "/admin/calendar_events?project_id=#{@loan.id}"
     @active_tab = params[:tab].presence || "details"
+    @questionnaire_filter = params[:filter] || "criteria"
```
emitche: There is duplication in the show and questionnaires methods in this controller. Is there a recommended way to adjust this?

app/views/admin/loans/show.html.slim
```diff
@@ -18,6 +18,7 @@ javascript:
       el: '.loan-content',
       calendarEventsUrl: #{json @calendar_events_url},
       loanId: #{json @loan.id},
-      locale: "#{I18n.locale}"
+      locale: "#{I18n.locale}",
+      questionnaireFilter: #{json @questionnaire_filter}
```
emitche: Because of the current tab structure, the initial questionnaire view model is initialized inside the loan tabs view model. We can clean this up, but the tab restructuring is not in yet.

app/assets/javascripts/backbone/views/loan_tab_view.coffee
```diff
@@ -4,6 +4,7 @@ class MS.Views.LoanTabView extends Backbone.View
 
   initialize: (params) ->
     @loanId = params.loanId
+    @questionnaireFilter = params.questionnaireFilter
```
emitche: See my comment on the show.html.slim file.

app/controllers/admin/loans_controller.rb
emitche: Now that the loan tab changes are in, this can be refactored.

emitche:
This requires some refactoring given the new tab structure.

PR #259, opened 2017-02-27
3744 dashboard project list

emitche:
- Add simple project list in dashboard
- Show the latest 15 projects belonging to user
- Add filter for completed, active, and all projects


app/controllers/admin/dashboard_controller.rb
```diff
@@ -1,8 +1,17 @@
 class Admin::DashboardController < Admin::AdminController
+
+  STATUS_FILTERS = ['all', 'active', 'completed']
```
emitche: Where would this best be placed? Inside the project's model, here, elsewhere?

app/controllers/admin/dashboard_controller.rb
```diff
@@ -11,4 +16,28 @@ def prep_calendar
...
+      include: [:primary_agent, :secondary_agent],
+      per_page: 15,
+      name: "recent_projects",
+      enable_export_to_csv: false
+    )
+
+    @status_filter_options = status_filter_options
+  end
+
+  def status_filter_options
```
emitche: Recommendations about how best to do this and the best place to put this code are welcome.

app/views/admin/dashboard/_recent_projects_grid_definition.html.slim
```diff
@@ -0,0 +1,27 @@
...
+      - @empty_list = true
+      - content_tag(:div, t('common.no_records'), class: 'alert alert-info')
+
+    - g.column name: t('common.id'), attribute: 'id', filter: false, in_csv: false,
+      ordering: false do |project|
+      - if project.type == "BasicProject"
+        - link_to(project.id, admin_basic_project_path(project.id))
+      - else
+        - link_to(project.id, admin_loan_path(project.id))
+    - g.column name: csv_id, attribute: 'id', in_html: false, filter: false
```
emitche: This line may not be needed, given that this table will not be exported to CSV. There are a few spots where in_html and in_csv are defined, but may be not needed.

app/assets/javascripts/backbone/views/dashboard_projects_view.coffee
```diff
@@ -0,0 +1,33 @@
+class MS.Views.DashboardProjectsView extends Backbone.View
+  el: '.projects-grid'
+
+  initialize: ->
+    @$all_projects = @$('.project-status').closest('tr')
+    @$active_projects = @$('.project-status.active').closest('tr')
+    @$completed_projects = @$('.project-status.completed').closest('tr')
+
+  events:
+    'change #completion_status': 'filterProjects'
```
hooverlunch: Can we use a dash instead of underscore in this ID?

app/assets/javascripts/backbone/views/dashboard_projects_view.coffee
```diff
@@ -0,0 +1,33 @@
+class MS.Views.DashboardProjectsView extends Backbone.View
+  el: '.projects-grid'
+
+  initialize: ->
+    @$all_projects = @$('.project-status').closest('tr')
+    @$active_projects = @$('.project-status.active').closest('tr')
+    @$completed_projects = @$('.project-status.completed').closest('tr')
```
hooverlunch: CamelCase for variable names please.

app/controllers/admin/dashboard_controller.rb
hooverlunch: Here is fine if it's just for this view, which I think it is. Please use the `%w(...)` notation though.

app/controllers/admin/dashboard_controller.rb
```diff
@@ -11,4 +16,28 @@ def prep_calendar
...
+    filter_options = []
+    STATUS_FILTERS.each do |item|
+        filter_options.push([I18n.t("dashboard.status_options.#{item}"), item])
+    end
+    return filter_options
```
hooverlunch: Try to remember to use blocks and `map` for things like this.

`STATUS_FILTERS.map { |f| [I18n.t("dashboard.status_options.#{f}"), f] }`

And since it's so simple now you really don't even need the separate function...

app/views/admin/dashboard/_recent_projects_grid_definition.html.slim
hooverlunch: Yeah seems like it shouldn't be needed.

emitche:
@rgalusca Take a look at the revised size of the dropdown filter. If you would like it adjusted, let me know in which ways. Thanks!

hooverlunch:
@emitche For things like this it might be an idea to provide a small screenshot of the dropdown so that @rgalusca doesn't have to fire up here environment and check out the branch?

emitche:
@hooverlunch Thanks. Will do in the future.

emitche:
image

PR #260, opened 2017-02-28
5886 add original date and duration

emitche:
- Add fields for original date and original duration in step modal
- Only show fields if they exist on step and if step is finalized


app/views/admin/project_steps/_modal_content.html.slim
```diff
@@ -77,10 +77,20 @@ section.step-fields class=(@mode == :show_and_form ? 'show-view' : 'edit-view')
         .view-element = @step.schedule_parent.summary_or_none if @step.schedule_parent.present?
         = f.input_field :schedule_parent_id, collection: @precedents, label_method: :summary_or_none, include_blank: true
 
+      - if @step.old_start_date && @step.is_finalized
```
hooverlunch: Please add `?` to `is_finalized` and also below.

config/locales/en/en.yml
```diff
@@ -117,6 +117,8 @@ en:
         actual_end_date: "Completed Date"
         duration: "Scheduled Duration"
         is_finalized: "Finalized"
+        old_duration_days: "Old Duration"
+        old_start_date: "Old Start Date"
```
hooverlunch: I think for display purposes, `Original` is better that `Old` unless you've been told otherwise by Roxana/client :)

PR #261, opened 2017-03-02
5928 sort groups

linesarefuzzy:


app/models/project_group.rb
```diff
@@ -54,6 +54,21 @@ def summary_or_none
     summary.blank? ? "[#{I18n.t("none")}]" : summary.to_s
   end
 
+  def scheduled_start_date
+    return @scheduled_start_date if defined?(@scheduled_start_date)
+    @scheduled_start_date = reload.children.map(&:scheduled_start_date).compact.min
```
linesarefuzzy: I don't know why all these `reload`s are necessary, but they make me nervous.

spec/models/project_group_spec.rb
```diff
@@ -41,18 +41,63 @@
...
+      it "should be correct for interior node" do
+        group = root.children[0]
+        expect(group.descendant_leaf_count).to eq group.children.size
+      end
     end
 
-    it "should be correct for interior node" do
-      group = root.children[0]
-      expect(group.descendant_leaf_count).to eq group.children.size
+    describe "max_descendant_group_depth" do
```
linesarefuzzy: This section I just moved from `timeline_entry_spec` (because these only apply to project groups).

app/models/project_group.rb
hooverlunch: Me too! Who says they are necessary? Does it not work without them?

app/models/project_group.rb
```diff
@@ -54,6 +54,21 @@ def summary_or_none
...
+  end
+
+  def scheduled_end_date
+    return @scheduled_end_date if defined?(@scheduled_end_date)
+    @scheduled_end_date = reload.children.map(&:scheduled_end_date).compact.max
+  end
+
+  def scheduled_duration_days
+    return @scheduled_duration_days if defined?(@scheduled_duration_days)
+    @scheduled_duration_days = scheduled_end_date - scheduled_start_date if scheduled_end_date && scheduled_start_date
```
hooverlunch: Shouldn't we set it to nil here and memoize that if start or end isn't set? So use a ternary?

app/models/timeline_entry.rb
```diff
@@ -49,12 +49,13 @@ class TimelineEntry < ActiveRecord::Base
 
   delegate :division, :division=, to: :project, allow_nil: true
 
-  scope :by_date, -> { order("scheduled_start_date is null desc, scheduled_start_date, scheduled_duration_days") }
+  # NOTE: This will only work for steps, but must be defined here in the parent class
```
hooverlunch: Comment cookies for you!! (Gluten free)

spec/models/project_group_spec.rb
hooverlunch: 👍

spec/models/project_group_spec.rb
```diff
@@ -41,18 +41,63 @@
...
+          g4 = create(:project_group, parent: root2)
+          g5 = create(:project_group, parent: root2)
+            create_dated_step(g5, "2017-03-01", 0)
+          g6 = create(:project_group, parent: root2)
+            g6_g1 = create(:project_group, parent: g6)
+              create_dated_step(g6_g1, "2017-01-10", 5)
+            g6_g2 = create(:project_group, parent: g6)
+              create_dated_step(g6_g2, "2017-01-01", 5)
+          s1 = create_dated_step(root2, nil, 0)
+          s2 = create_dated_step(root2, "2017-02-28", 30)
```
hooverlunch: This is so fucking hot. `create_dated_step`. YES.

app/models/project_group.rb
hooverlunch: Ok so I figured this out I think. I can describe it to you when you're in later. For now, just know that I moved the reloads into the spec where they belong and the specs are sill passing. So you can calm your nerves.

app/models/project_group.rb
hooverlunch: I made this change since I had the code open.

spec/models/project_group_spec.rb
linesarefuzzy: 🔥

PR #262, opened 2017-03-02
5828 transaction index on loans

X0nic:


X0nic:
This is blocked again because of silly QBO API strangeness. Now the issue seems to be that it is impossible to view transactions for a specific account. There is no filter implemented on QB side. If you try to request transactions for a specific account, you get `"QueryValidationError: property 'DepositToAccountRef' is not queryable","code":"4001"`. There are possible workarounds, but again they are less than optimal. Need to chat before proceeding. 

/cc @hooverlunch @linesarefuzzy

X0nic:
I found a minor bug. Will need to fix before I submit for code review.

app/controllers/admin/accounting/transactions_controller.rb
```diff
@@ -2,16 +2,12 @@ class Admin::Accounting::TransactionsController < Admin::AdminController
...
-    @transactions = Accounting::Transaction.all.with_qb_objects
-  end
-
-  private
-
-  def auth_details
-    { access_token: qb_connection.access_token, company_id: qb_connection.realm_id }
-  end
+    Accounting::Quickbooks::AccountFetcher.new.fetch
+    Accounting::Quickbooks::TransactionFetcher.new.fetch
```
hooverlunch: This looks nice.

app/controllers/admin/loans_controller.rb
```diff
@@ -160,4 +162,17 @@ def prep_questionnaire
...
+
+  def prep_transactions
+    begin
+      Accounting::Quickbooks::AccountFetcher.new.fetch
+      Accounting::Quickbooks::TransactionFetcher.new.fetch
+    rescue
+      flash.now[:error] = 'Error connecting to quickbooks'
+    end
+
+    @transactions = ::Accounting::Transaction.where(project_id: @loan.id)
```
hooverlunch: Why is leading `::` needed here but not above?

app/models/accounting/quickbooks/account_fetcher.rb
```diff
@@ -0,0 +1,35 @@
...
+
+        acc_transaction.update_attributes!(quickbooks_data: qb_object.as_json, name: qb_object.name)
+      end
+
+      def service
+        ::Quickbooks::Service::Account.new(auth_details)
+      end
+
+      def auth_details
+        { access_token: qb_connection.access_token, company_id: qb_connection.realm_id }
```
hooverlunch: No space inside {} for hashes plz.

app/models/accounting/quickbooks/account_fetcher.rb
```diff
@@ -0,0 +1,35 @@
+module Accounting
+  module Quickbooks
+    class AccountFetcher
```
hooverlunch: There is quite a bit of repeated code here and in TransactionFetcher. Could you refactor to use inheritance?

app/models/accounting/transaction.rb
```diff
@@ -2,37 +2,50 @@
...
-  def self.with_qb_objects
-    transactions = all
-    return transactions if transactions.count < 1
+  def txn_date
+    quickbooks_data[:txn_date]
+  end
+
+  def total
+    quickbooks_data[:total]
+  end
```
hooverlunch: You are both delegating and defining these methods. Please do one or the other?

app/models/accounting/transaction.rb
```diff
@@ -2,37 +2,50 @@
...
-    Accounting::Quickbooks::Fetcher.new(transactions).fetch
+  def private_note
+    quickbooks_data[:private_note]
   end
 
-  def qb_object
-    raise ArgumentError, 'qb_object not already set. Make sure to append #with_qb_objects at the end of an AR chain.' if @qb_object.blank?
-    @qb_object
+  def quickbooks_data
+    super.with_indifferent_access
```
hooverlunch: This is clever but it took me a minute to figure out what you're doing. Would `read_attribute` be clearer?

Also, using either method means any changes to quickbooks_data via this accessor won't be persisted. Is that ok?

app/views/admin/dashboard/dashboard.html.slim
```diff
@@ -15,8 +15,8 @@
               = select_tag(:completion_status, options_for_select(@status_filter_options, "all"),
                 class: "form-control")
 
-      - unless @empty_list
-        = render_index_grid(@recent_projects_grid)
+      / - unless @empty_list
+      = render_index_grid(@recent_projects_grid)
```
hooverlunch: Please remove commented code.

app/views/admin/loans/_transactions.html.slim
```diff
@@ -0,0 +1,14 @@
...
+
+  = grid(@transactions_grid) do |g|
+    - g.column name: 'Id', attribute: 'qb_id'
+    - g.column name: 'Type' do |transaction|
+      - transaction.qb_transaction_type
+    - g.column name: 'Date' do |transaction|
+      - transaction.txn_date
+    - g.column name: 'Total' do |transaction|
+      - transaction.total
+    - g.column name: 'Note' do |transaction|
```
hooverlunch: These grid headers should be i18n'd. Preferably using the Wice default lookup location instead of calling `t` for each one, if possible.

db/migrate/20170301172505_create_accounting_accounts.rb
```diff
@@ -0,0 +1,12 @@
+class CreateAccountingAccounts < ActiveRecord::Migration
+  def change
+    create_table :accounting_accounts do |t|
+      t.string :name, null: false
+      t.string :qb_account_classification, unique: true
+      t.string :qb_id, null: false, index: true, unique: true
+      t.json :quickbooks_data
+
+      t.timestamps null: false
```
hooverlunch: Nice constraints and indices.

db/migrate/20170301172852_add_accounting_account_ref_to_accounting_transaction.rb
```diff
@@ -0,0 +1,8 @@
+class AddAccountingAccountRefToAccountingTransaction < ActiveRecord::Migration
+  def change
+    add_reference :accounting_transactions, :accounting_account, index: true, foreign_key: true #, null: false
```
hooverlunch: Why is this commented out?

db/migrate/20170301172852_add_accounting_account_ref_to_accounting_transaction.rb
```diff
@@ -0,0 +1,8 @@
...
+    add_reference :accounting_transactions, :project, index: true, foreign_key: true
+    add_column :accounting_transactions, :quickbooks_data, :json
+    rename_column :accounting_transactions, :qb_transaction_id, :qb_id
```
hooverlunch: Nice refactor.

spec/models/accounting/quickbooks/account_fetcher_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe Accounting::Quickbooks::AccountFetcher, type: :model do
+  subject { described_class.new }
+
+  it 'should fetch all records for Account' do
+    service = instance_double(Quickbooks::Service::Account)
```
hooverlunch: Could add `all: []` here as you do below.

app/controllers/admin/loans_controller.rb
X0nic: Good catch. The perils of catching `Exception`.

app/views/admin/dashboard/dashboard.html.slim
X0nic: Gah. That was supposed to be stashed. It was for another PR.

app/controllers/admin/loans_controller.rb
hooverlunch: Oh, wow. Yeah I missed that. Please get rid of any generic `rescue`s.

X0nic: Yeah, will do. I added it becase `quickbooks-ruby` did not have a base exception to catch. Ill just catch it in the fetcher, and rethrow one I can catch.

app/models/accounting/transaction.rb
X0nic: Should still be fine, I didn't override `def quickbooks_data=`. I suppose you couldn't update the field without replacing the whole block of json. I doubt it will be an issue. We really don't want anyone besides quickbooks changing the data here.

app/views/admin/loans/_transactions.html.slim
X0nic: I don't think wice_grid respects I18N for column names. I took a quick look at the docs and source, this appears to be the case.

PR #264, opened 2017-03-02
5933 edit functionality for dashboard steps

emitche:
- Open step modal from dashboard calendar
- Refactor calendar view models to move more logic into the parent view model


app/assets/javascripts/backbone/views/calendar_view.coffee
```diff
@@ -1,6 +1,28 @@
 # Common functions for calendars application-wide.
 class MS.Views.CalendarView extends Backbone.View
 
+  events:
+    'click .cal-step': 'stepClick'
+
+  prepareVariables: (params) ->
+    @$calendar = @$('#calendar')
+    @stepModal = params.stepModal
```
hooverlunch: Does the DashboardCalendarView use this? Perhaps this can stay in the initializer for GeneralCalendarView?

app/assets/javascripts/backbone/views/calendar_view.coffee
```diff
@@ -1,6 +1,28 @@
...
+
+  prepareDefaultSettings: (params, context) ->
```
hooverlunch: This is a neat refactor. Well done! Here are a few ways to make it simpler:

* Remove this method and just set `defaultSettings` inside the `initializeCalendar` method. Doesn't even have to be an instance variable.
* Pass the special settings as an argument to the `initializeCalendar` method instead of using an instance var.

app/assets/javascripts/backbone/views/calendar_view.coffee
emitche: Both use it. This particular feature (adding ability to edit steps from dashboard) will need tge step modal.

PR #267, opened 2017-03-03
5951 basic project logs

linesarefuzzy:


linesarefuzzy:
Probably for next sprint.

app/controllers/admin/projects_controller.rb
```diff
@@ -19,8 +19,14 @@ def steps
...
     authorize @project, :show?
-    prep_timeline(@project)
-    render partial: "admin/timeline/table", locals: {project: @project}
+    if request.xhr?
+      prep_timeline(@project)
+      render partial: "admin/timeline/table", locals: {project: @project}
+    elsif @project.is_a? Loan
+      redirect_to admin_loan_tab_path(@project, tab: 'timeline')
+    elsif @project.is_a? BasicProject
+      redirect_to admin_basic_project_tab_path(@project, tab: 'timeline')
```
hooverlunch: `is_a?` is often a code smell (see Metz book, duck typing is usually better) but here I don't see a way to do that without coupling the route logic to the Loan/BasicProject models...

Although you could have a helper method here called `project_tab_path` that does something like `send("admin_#{@project.model_name.singular)_tab_path", ...)` behind the scenes. I think that would be preferable.

app/controllers/admin/projects_controller.rb
```diff
@@ -35,4 +41,12 @@ def prep_timeline(project)
...
     end
   end
+
+  def prep_logs(project)
+    @org = Organization.find(params[:org]) if params[:org]
+    @step = ProjectStep.find(params[:step]) if params[:step]
+    @logs = ProjectLog.in_division(selected_division).filter_by(project: project.id).
+        order('date IS NULL, date DESC, created_at DESC').
+        page(params[:page]).per(params[:per_page])
+  end
```
hooverlunch: Love to see code getting moved into the superclass!

app/views/admin/loans/_logs.html.slim
```diff
@@ -11,7 +9,7 @@ javascript:
   $(function() {
     new MS.Views.LogListView({
       el: 'section.log-list',
-      refreshUrl: #{json refresh_url},
+      refreshUrl: #{json "/admin/logs?project=#{@loan.id}"},
```
hooverlunch: Not sure why the `org` stuff is able to be deleted here or why it was here in the first place, but I'm trusting you, and this is cleaner!

linesarefuzzy: Yeah, I'm not sure why it was there either, or why it used to work. It seems like it was just copied from the analogous partial in logs index.

app/controllers/admin/projects_controller.rb
linesarefuzzy: Ok, but I think defining a helper with that name might be confusing since people would think to look for it in routes. How about I just drop that `send` code in here?

linesarefuzzy: (btw I also try to avoid `is_a` but didn't see an obvious way around it here)

PR #268, opened 2017-03-03
5981 bug fix change project start end date in calendar

emitche:
- Fix bug when moving project start and end date in dashboard and main calendar

### DIFF
--- a/app/assets/javascripts/backbone/views/calendar_view.coffee
+++ b/app/assets/javascripts/backbone/views/calendar_view.coffee
@@ -23,8 +23,7 @@ class MS.Views.CalendarView extends Backbone.View
         .done => @refresh()
         .fail => revertFunc()
       else
-        stepId = event.model_id
-        $.post("/admin/timeline_step_moves/#{stepId}/simple_move",
+        $.post("/admin/timeline_step_moves/#{event.model_id}/simple_move",
           _method: "PATCH"
           scheduled_start_date: event.start.format('YYYY-MM-DD'))
           .done => @refresh()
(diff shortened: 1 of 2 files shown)


emitche:
For sprint 1.8.

app/assets/javascripts/backbone/views/calendar_view.coffee
```diff
@@ -23,8 +23,7 @@ class MS.Views.CalendarView extends Backbone.View
         .done => @refresh()
         .fail => revertFunc()
       else
-        stepId = event.model_id
```
hooverlunch: Nice simplification!!!

app/assets/javascripts/backbone/views/calendar_view.coffee
```diff
@@ -33,8 +32,7 @@ class MS.Views.CalendarView extends Backbone.View
       # We use a 1ms timeout so that fullCalendar can finish drawing the event in the new calendar cell.
       setTimeout =>
         if confirm(I18n.t("loan.move_date_confirm.body"))
-          projectId = @$el.find('#calendar').data('project-id')
```
hooverlunch: I'm curious why this was ever here instead of using `event.model_id`. Perhaps just an oversight. In any case if this works then great! As long as we have tested it in all the various calendar locations.

app/views/admin/calendar/_event.html.slim
```diff
@@ -1,7 +1,8 @@
 - case cal_event.event_type
   - when "project_start", "project_end"
 
-    div.calendar-event class="#{cal_event.event_type == 'project_start' ? 'project-start' : 'project-end'}"
+    div.calendar-event class="#{cal_event.event_type.gsub('_', '-')}"
```
hooverlunch: Please see my commit here. Make sense?

emitche: Makes sense.

PR #270, opened 2017-03-03
6009 Remove ProjectStepModalView from LoanTabView.

X0nic:
It is no longer needed because of the tab refactor.


X0nic:
@linesarefuzzy just fixed this in an unrelated bug.

PR #271, opened 2017-03-13
Loan health check

X0nic:
- [x] Loan marked active and no contract attached
- [x] A step more than a day late and incomplete
- [x] Less than 80% of the required questions answered
- [x] Less than one step every month
  - [x] Specifically, every 30 day period between the loan start date and target end date should have at least one step.
  - [x] For this calculation, use the step's end date rather than start date.
- [x] Loan marked active and no logs in longer than a month

- [x] The backend should return a list of specific warnings.


app/models/loan.rb
```diff
@@ -189,4 +189,49 @@ def ensure_currency
...
+
+  def days_old
+    (end_date.beginning_of_day - signing_date.beginning_of_day) / (24 * 60 * 60)
+  end
+
+  private
+
+  def most_recent_log_date
+    project_logs.maximum(:date) || Time.at(0)
+  end
```
hooverlunch: The new methods in this file are nicely crafted, but I'm concerned about performance. Running the health checks for a single loan will kick off at least 5 non trivial queries:

* media
* timeline_entries x2
* progress_pct, which loads in the full LoanResponseSet and associated LoanQuestions
* project_logs

It's my understanding that this health check feature is designed to be run for a bunch of loans at once. Doing this will therefore kick of a ton of big queries and be super slow.

app/models/loan.rb
```diff
@@ -189,4 +189,49 @@ def ensure_currency
   def amount_formatted
     ensure_currency.format_amount(amount)
   end
+
+  def signed_contract?
+    media.where(kind_value: 'contract').count > 0
```
hooverlunch: You can use `.any?` instead of `.count > 0`. Both will result in a COUNT query.

app/models/loan.rb
hooverlunch: I think the best way around this is probably to cache all the necessary values and update them whenever the appropriate records are updated.

So...

* When media are updated, cache whether there is a contract
* When a step is saved, cache whether there are any past due or recently updated steps
* When a ResponseSet is updated, calculate the progress_pct and cache that (and also perhaps refactor the Loan details page to use this value instead of calculating directly)
* When a log is created, update the max log date

If all these things are stored as columns in the projects table, then calculating the health checks can be done entirely in memory once the Loans are loaded on the dashboard.

If you agree with these changes, I would like to be involved in designing how they will be implemented because I am worried about a bunch of callback couplings and things getting brittle and scary... I have some ideas on how to avoid that.

app/jobs/recalculate_loan_job.rb
```diff
@@ -0,0 +1,14 @@
...
+  def perform(loan_id: )
+    check = LoanHealthCheck.where(project_id: loan_id).first
+
+    unless check
+      Rails.logger.debug "Creating health check for loan #{loan_id}"
+      check = LoanHealthCheck.create(project_id: loan_id)
+    end
+
+    check.recalculate
+  end
```
hooverlunch: Nice.

app/models/loan.rb
```diff
@@ -68,6 +69,9 @@ class Loan < Project
 
   validates :organization_id, presence: true
 
+  before_create :build_loan_health_check
+  after_commit :recalculate_loan
```
hooverlunch: Nice use of after_commit!

app/jobs/recalculate_loan_job.rb
```diff
@@ -0,0 +1,14 @@
+class RecalculateLoanJob < ActiveJob::Base
```
hooverlunch: Can we please call this RecalculateLoanHealthJob, and elsewhere in the code add the 'health' piece? There could be other things about Loans that we are calculating so I want it to be clear this is just about health.

app/models/loan.rb
```diff
@@ -85,6 +89,10 @@ def self.filter_by_params(params)
     scoped
   end
 
+  def recalculate_loan
```
hooverlunch: add _health

app/models/loan.rb
```diff
@@ -189,4 +197,14 @@ def ensure_currency
...
   end
+
+  def active?
+    status_value == 'active'
+  end
+
+  def healthy?
+    return false unless loan_health_check
+    loan_health_check.healthy?
+  end
```
hooverlunch: So nice and clean! I love this solution!

app/models/loan_health_check.rb
```diff
@@ -0,0 +1,84 @@
...
+#
+#  index_loan_health_checks_on_project_id  (project_id)
+#
+# Foreign Keys
+#
+#  fk_rails_6441c92912  (project_id => projects.id)
+#
+
+class LoanHealthCheck < ActiveRecord::Base
+  belongs_to :loan, class_name: Project, foreign_key: :project_id
```
hooverlunch: This seems odd. If it's just for loans, can't you use class_name Loan and foreign_key loan_id?

spec/models/loan_health_check_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+      context 'with criteria' do
+        let(:progress) { 57 }
+        before do
+          allow(check).to receive(:loan).and_return(loan)
+          allow(loan).to receive(:criteria).and_return(instance_double(LoanResponseSet, progress_pct: progress))
+          check.recalculate
+        end
+
+        it 'matches criteria progress_pct' do
+          is_expected.to eq progress
```
hooverlunch: Hmm this is a fun bit of syntactic sugar!

spec/models/loan_health_check_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+    let(:check) { loan.loan_health_check }
+    let(:loan) { create(:loan, :prospective) }
+    subject { check }
+
+    describe '.progress_pct' do
+      subject { check.progress_pct }
+
+      context 'without criteria' do
+        it 'returns 0' do
+          expect(subject).to eq 0
```
hooverlunch: Use `is_expected_to` here?

spec/models/loan_health_check_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+    describe '.healthy' do
+      subject { check.healthy? }
+
+      context 'with no #health_warnings' do
+        before do
+          allow(check).to receive(:health_warnings).and_return([])
+        end
+        it { is_expected.to be true }
+      end
+      context 'with 1 #health_warnings' do
```
hooverlunch: Generally we do blank lines between context blocks and after before blocks.

spec/models/loan_health_check_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+                :with_a_number_of_recent_project_steps,
+                :with_a_number_of_old_project_steps,
+                step_count: 29,
+                old_step_count: 5)
+            end
+
+            it 'has found 29 steps' do
+              expect(loan.timeline_entries.merge(ProjectStep.recent).count).to eq 29
+            end
+            it { is_expected.to include :sporadic_loan_updates }
```
hooverlunch: I thought we had realized this was the wrong rule? Wasn't it just that there needs to be at least one step in each future 30 day period until the end date or something? Did you discuss with Fuzzy?

spec/models/loan_health_check_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+
+          context 'and logs within 30 days' do
+            let(:loan) { create(:loan, :active, :with_contract, :with_recent_logs) }
+
+            it { is_expected.to_not include :active_without_recent_logs }
+          end
+        end
+      end
+    end
+  end
```
hooverlunch: Wow, great coverage here!

app/models/loan.rb
X0nic: Funny story about this. I only know about `after_commit` because I have used Sidekiq in the past. Sidekiq picks up jobs so fast, that if you use after_save Sidekiq is already processing your job before your data is committed to the DB.

PR #272, opened 2017-03-13
5805 dashboard list of users

emitche:
- Add list of users with projects lists
- Refactor projects list to be reused on page


app/views/admin/dashboard/dashboard.html.slim
```diff
@@ -32,6 +32,16 @@
     section.log-list
       = render "admin/project_logs/log_list"
 
+  section.section.people
+    h2 Users
```
emitche: Use translations.

emitche:
This feature is awaiting feedback from the client before completion.

app/controllers/admin/dashboard_controller.rb
```diff
@@ -34,6 +35,26 @@ def prep_projects_grid
     @status_filter_options = STATUS_FILTERS.map { |f| [I18n.t("dashboard.status_options.#{f}"), f] }
   end
 
+  # Prepare grids for all users inside selected division
+  def prep_projects_grids
+    @people = @division.people.where(has_system_access: true).where.not(id: @person.id)
```
hooverlunch: Can you also please add a scope to Person called `with_system_access` and refactor any places that have the full `where` like you have. I think there are around three others.

app/controllers/admin/dashboard_controller.rb
```diff
@@ -34,6 +35,26 @@ def prep_projects_grid
...
+
+    @people_with_projects = []
+    @grids = {}
+    @people.each do |person|
+      if person.agent_projects.length > 0
+        @people_with_projects << person
+        @grids["person_#{person.id}"] = initialize_grid(
```
hooverlunch: Simplification: can you just use `person` by itself as the hash key? In Ruby you can use objects (or anything really) as hash keys

app/views/admin/dashboard/_recent_projects_grid_definition.html.slim
```diff
@@ -1,4 +1,4 @@
-= define_grid(@recent_projects_grid, hide_csv_button: true,
+= define_grid(grid, hide_csv_button: true,
```
hooverlunch: Nice reuse!!!!

app/controllers/admin/dashboard_controller.rb
hooverlunch: Also, the var should probably be called `@people_grids` if this is the only thing it's being used for?

hooverlunch: It's unfortunate that this is going to fire N queries where N is the number of users being loaded. Generally we try to avoid this (it's called the "N+1 queries" problem). But I don't see an easy way around it here.

emitche: I spoke with @linesarefuzzy  about this. Wice Grid appears to make its own query every time a table is loaded/initialized. Because of this, though it would be great to just load the projects once and filter them per each user, that is not how Wice Grid works.

hooverlunch: Yup, agreed.

app/controllers/admin/dashboard_controller.rb
emitche: I'm not fully understanding this part.

app/controllers/admin/dashboard_controller.rb
```diff
@@ -34,6 +35,26 @@ def prep_projects_grid
...
+          person.agent_projects.order(created_at: :desc, updated_at: :desc).limit(15),
+          include: [:primary_agent, :secondary_agent],
+          per_page: 5,
```
emitche: Limits were added for performance reasons, but it seems this is loading fast enough. So I agree. Have removed the earlier limits.

emitche:
The last two requested refactors are not quick refactors. I have made the smaller changes I can make. If this needs to go in to a release quickly, the remaining changes can be added as a separate pull request later. These refactors do not affect the functionality. There is still time in the remaining estimate to address these refactors as needed.

PR #274, opened 2017-03-15
5990: Remove background from current day

emitche:
- Remove yellow background on current day
- Applies to all calendars


app/assets/stylesheets/admin/calendar/_calendar.scss
```diff
@@ -41,6 +41,12 @@
     background: $white;
   }
 
+  .fc {
```
hooverlunch: Perhaps this isn't needed?

PR #275, opened 2017-03-15
5882 require log on duration change

emitche:
- Revise how old duration is set
- Show log modal when duration is changed and step is finalized


app/models/project_step.rb
```diff
@@ -320,6 +320,10 @@ def pending_days_shifted
     0
   end
 
+  def pending_duration_change?
+    is_finalized? ? scheduled_duration_days_changed? : false
```
hooverlunch: You can replace this with `is_finalized? && schedule_duration_days_changed?`

As a rule of thumb, you shouldn't need the ternary operator if your goal is to return a boolean.

app/controllers/admin/project_steps_controller.rb
```diff
@@ -59,18 +59,24 @@ def update
...
+
     if %w(timeline_table calendar).include?(params[:context])
-      valid ? render(json: {id: @step.id, days_shifted: days_shifted}) : render_modal_content(422)
+      valid ? render(json: options) : render_modal_content(422)
     else
       render partial: '/admin/project_steps/project_step', locals: {
         step: @step,
         mode: valid ? :show : :edit,
         days_shifted: days_shifted,
+        duration_changed: duration_changed,
```
hooverlunch: Is this used anywhere? I'm thinking not. Which is fine I think since this is for the timeline list view. So I don't think this change will work there, but that's ok in my book because I still really want to get rid of this old code!!! But I think you can remove this line as it's not doing anything presently.

app/models/project_step.rb
```diff
@@ -351,11 +355,14 @@ def handle_old_start_date_logic
...
-    # the original scheduled date.
+    # Set old duration days once and only if the step is finalized (not a draft)
     return unless persisted? && scheduled_duration_days_changed? && is_finalized?
 
-    self.old_duration_days = scheduled_duration_days_was
+    # By default, old_duration_days is set to 0.
+    # Only remember old duration if a duration has been set other than the default and then changed.
+    unless old_duration_days > 0
+      self.old_duration_days = scheduled_duration_days_was
+    end
```
hooverlunch: This seems to make sense.

PR #276, opened 2017-03-15
5887: When there is an empty timeline, display a message

emitche:
Show message when there are no groups or steps in timeline


emitche:
I've removed some obsolete code. If this belongs in a separate pull request, let me know.

emitche:
I added a commit that was for a different by mistake. Deleting.

PR #278, opened 2017-03-15
5887 empty timeline

emitche:
- Show message when there are no groups or steps in timeline
- Remove obsolete code that applied to projects with incorrect root timeline entry


emitche:
I've removed some obsolete code. If those changes (commit 2) belong in a separate pull request, let me know.

app/views/admin/timeline/_main.html.slim
```diff
@@ -2,15 +2,12 @@
...
-    - if project.root_timeline_entry
-      - if project.root_timeline_entry.group?
-        .table-wrapper
-          = render "admin/timeline/table", project: project
-      / If user has created a step through the timeline list, the timeline table is incompatible
-      - elsif project.root_timeline_entry.step?
-        p This loan is not yet compatible with the timeline table.
+
+    - if project.root_timeline_entry.children.empty?
+      .alert.alert-info = t("project_step.no_steps")
```
hooverlunch: I don't think an alert will look very good against the gray background of the timeline. In general I think alerts are designed to be displayed at the top of the main content area. They have a lot of visual weight (large padding, background color, border, etc.) to draw attention at the top of the page.

For this situation, I think just a `p` tag with style `text-align: center; margin: 40px 0` will look very nice. We could define a general class for this like `.no-results`. I think we should use it for the no results message for grids as well.

This is also more typical, from what I've seen.

PR #280, opened 2017-03-18
6096 errors on duplicate step

linesarefuzzy:
resolve "that day doesn't exist in that month" issue use last day of month instead

### DIFF
--- a/app/models/timeline/step_duplication.rb
+++ b/app/models/timeline/step_duplication.rb
@@ -88,9 +88,16 @@ module Timeline
         reference_date = date.beginning_of_month + interval - 1.month
         first_try = Chronic.parse("#{month_repeat_on} of next month", now: reference_date)
 
-        # Sometimes 5th weekday of month doesn't exist
-        if first_try.nil? && month_repeat_on =~ /\A5th/
-          Chronic.parse("#{month_repeat_on.sub('5th', '4th')} of next month", now: reference_date)
+        if first_try.nil?
+          # Sometimes 5th weekday of month doesn't exist
+          if month_repeat_on =~ /\A5th/
+            Chronic.parse("#{month_repeat_on.sub('5th', '4th')} of next month", now: reference_date)
+          # Error if day to repeat doesn't exist in some months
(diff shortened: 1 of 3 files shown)


app/models/timeline/step_duplication.rb
```diff
@@ -59,18 +59,7 @@ def basis_weekday_key
...
-      elsif (8 <= day) && (day < 15)
-        2
-      elsif (15 <= day) && (day < 22)
-        3
-      elsif (22 <= day) && (day < 29)
-        4
-      else
-        5
-      end
+      (day - 1).div(7) + 1
```
hooverlunch: I admit I was not aware of `div`. Cute.

hooverlunch: Glad we have test coverage on this...

config/locales/en/en.yml
```diff
@@ -400,7 +400,7 @@ en:
       num_2: "second"
       num_3: "third"
       num_4: "fourth"
-      num_5: "fifth"
+      num_5: "last" # This reflects the actual behavior of the repeat function
```
hooverlunch: Nice detail.

app/models/timeline/step_duplication.rb
```diff
@@ -99,9 +88,16 @@ def apply_time_interval(date, frequency, time_unit, month_repeat_on)
...
+        if first_try.nil?
+          # Sometimes 5th weekday of month doesn't exist
+          if month_repeat_on =~ /\A5th/
+            Chronic.parse("#{month_repeat_on.sub('5th', '4th')} of next month", now: reference_date)
+          # Error if day to repeat doesn't exist in some months
+          elsif month_repeat_on.to_i > 28
+            (date + interval).end_of_month
+          else
+            raise "Error parsing date string for monthly repeat"
+          end
```
hooverlunch: Can you please update the specs to reveal the previous flaw here? Any time you fix a bug in a model like this it's a good idea to cover the change with a spec. Even better is to do it TDD style.

PR #281, opened 2017-03-21
6111 with system access

emitche:
- Replace people filters with condensed `with_system_access` filter


emitche:
In #272, changes were requested to other files not part of the original feature. A filter was being added to condense people filters. This branch applies the filter to other files.

PR #283, opened 2017-03-22
5884 style changes timeline completed steps

emitche:
- Remove gray background on steps
- Show completed message with checkbox in step status
- Show completed days late message in step status, if applicable

### DIFF
--- a/app/assets/stylesheets/admin/loans/timeline/_table.scss
+++ b/app/assets/stylesheets/admin/loans/timeline/_table.scss
@@ -81,10 +81,6 @@
     }
   }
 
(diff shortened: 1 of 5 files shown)


emitche:
This makes sense. Make the status information into a helper method and move the logic out of the model.

emitche:
I decided not to refactor the admin_status for the timeline list, as the timeline list displays the same information but in different spots (whether or not an item is completed, how late a step is). The condensed status duplicates info in the timeline list and is more confusing. As we have already moved to timeline table, it does not make sense to make big refactors to timeline list.

PR #284, opened 2017-03-23
3690 quickboks dev env

emitche:
- Add additional instructions for connecting QuickBooks

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,16 +96,16 @@ On server:
 In order to test any of the Quickbooks features, you will need to create Intuit Developer account and sandbox.
 
 1. Visit https://developer.intuit.com
-2. Click "Sign In"
-3. Create an account, or sign in
-4. Click on "My Apps"
-5. Click on "Select APIs" under "Just start coding"
-6. Select "Accounting" and click "Create app"
-7. Click on "Keys"
-8. Copy the OAuth Key and Secret into your .env file. Use `.env.example` as a template.
-9. While logged into the application (Madeline site), visit Manage > Settings.
-10. Click on the button that says "Connect to Quickbooks."
-11. Sign in to the Quickbooks account for your developer account.
-12. Click authorize to connect your account data to Madeline.
+1. Click "Sign In"
+1. Create an account, or sign in
+1. Click on "My Apps"
+1. Click on "Select APIs" under "Just start coding"
+1. Select "Accounting" and click "Create app"
+1. Click on "Keys"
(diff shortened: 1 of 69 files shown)


README.md
```diff
@@ -103,5 +103,9 @@ In order to test any of the quickbook features, you will need to create Intuit D
 6. Select "Accounting" and click "Create app"
 7. Click on "Keys"
 8. Copy the OAuth Key and Secret into your .env file. Use `.env.example` as a template.
+9. While logged into the application (Madeline site), visit Manage > Settings.
+10. Click on the button that says "Connect to Quickbooks."
+11. Sign in to the Quickbooks account for your developer account.
+12. Click authorize to connect your account data to Madeline.
```
hooverlunch: Note that it's better in Markdown to use `1.` for every item in a numbered list like this as it makes adding items in between easier. But this is fine for now.

app/controllers/admin/accounting/transactions_controller.rb
```diff
@@ -7,7 +7,8 @@ def index
 
     @transactions = Accounting::Transaction.all
   rescue
-    flash.now[:error] = 'Error connecting to quickbooks'
+    flash.now[:error] = "Error connecting to QuickBooks. Visit
+      #{view_context.link_to('Settings', admin_settings_path)}.".html_safe
```
hooverlunch: Wow very nice use of `view_context` :)

hooverlunch: One thing: you can avoid use of html_safe, I think, if you do like this:

`"Error connecting to QuickBooks. Visit " << view_context.link_to('Settings', admin_settings_path)`

app/controllers/admin/loans_controller.rb
```diff
@@ -170,7 +170,8 @@ def prep_transactions
     rescue Accounting::Quickbooks::FetchError => e
       Rails.logger.error e
       Rails.logger.error e.cause
-      flash.now[:error] = 'Error connecting to quickbooks'
+      flash.now[:error] = "Error connecting to QuickBooks. Visit
+        #{view_context.link_to('Settings', admin_settings_path)}.".html_safe
```
hooverlunch: Same thing here with html_safe. Best to avoid it if at all possible.

README.md
linesarefuzzy: This is probably because of my use of Typora when editing this file.

PR #288, opened 2017-03-27
6128: Hide Manage Logs link when no logs exist in dashboard

emitche:
- Hide "Manage Logs" link when there are no logs in dashboard


emitche:
@hooverlunch You may not need to review this as it is a tiny change and @linesarefuzzy has approved it. If it doesn't need additional review, just mark complete. Thanks!

PR #289, opened 2017-03-27
5826 Remove inactive and retired questions from toc.

X0nic:


app/views/admin/loans/questionnaires/_questionnaire_toc.html.slim
```diff
@@ -1,6 +1,6 @@
-  ol
+  ul
```
X0nic: Currently the Toc numbering is done in Css. It does not use the same method as the body, so the numbers would not line up. Chatted to @linesarefuzzy about this one, and because the numbering is going to be refactored soon, we decided to just disable it for now on the Toc.

app/models/loan_question.rb
```diff
@@ -137,7 +137,7 @@ def active?
   end
 
   def child_groups
-    children.select(&:group?)
+    children.order(:position).select(&:group?)
```
hooverlunch: Please avoid doing AR scopes on the children method because it forces a cascading series of db queries. We are using the closure_tree full-tree eager loading functionality but that only works if you keep everything in memory. I have fixed this as I noticed another spot where it also needed fixing. Need to override children to make it not possible to do this in future. Will likely do that in an upcoming refactor.

PR #290, opened 2017-03-28
6124 quickbooks account settings

emitche:
- Add form to connect specific QuickBooks accounts
- Adjust settings page


app/assets/stylesheets/admin/_settings.scss
```diff
@@ -0,0 +1,16 @@
...
+
+    .note {
+      font-style: italic;
+    }
+
+    .form-horizontal {
+      margin-top: 1em;
+    }
+  }
+}
```
hooverlunch: Looks good here.

app/controllers/admin/settings_controller.rb
```diff
@@ -1,7 +1,26 @@
 class Admin::SettingsController < Admin::AdminController
 
   def index
-    authorize :setting, :index?
+    authorize :setting
```
hooverlunch: Nice fix.

app/views/admin/settings/index.html.slim
```diff
@@ -1,27 +1,48 @@
...
-          a href="#{disconnect_admin_accounting_quickbooks_path}" Disconnect
+        label.col-sm-4.text-right = t("quickbooks.connection.disconnect_from_qb")
+        .col-sm-8
+          a href="#{disconnect_admin_accounting_quickbooks_path}" = t("quickbooks.connection.disconnect")
     - else
       .form-group
-        label.col-sm-2 Connect to Quickbooks:
-        .col-sm-10
+        label.col-sm-4.text-right = t("quickbooks.connection.connect_to_qb")
+        .col-sm-8
```
hooverlunch: Nice cleanups here!

app/views/admin/settings/index.html.slim
```diff
@@ -1,27 +1,48 @@
...
+      url: admin_settings_path,
+      method: 'patch',
+      html: {class: 'form-horizontal'},
+      wrapper: :horizontal_form_expanded,
+    ) do |f|
+
+    = error_notification(f)
+    = f.input :principal_account_id, collection: @accounts
+    = f.input :interest_receivable_account_id, collection: @accounts
+    = f.input :interest_income_account_id, collection: @accounts
```
hooverlunch: What happens if you don't have any accounts? Important to always test the zero-case. I think these will be empty dropdowns which is kind of an eyesore. Perhaps if there are no accounts you can just show a message saying 'No accounts have been retrieved from QuickBooks yet.'

app/views/admin/settings/index.html.slim
```diff
@@ -1,27 +1,48 @@
...
+
+    .text-right
+      = f.submit t(:save), class: "btn btn-primary"
```
hooverlunch: Should this save button appear outside the block? Else we will eventually end up with multiple save buttons on the page which will be a bad UX? Thoughts?

app/views/admin/settings/index.html.slim
hooverlunch: Actually now that I look at this more there is something a bit troubling. We are replicating the HTML for a form which is not very DRY. Imagine in the future we want to change the form template a bit (quite likely), we will forget about this one-off duplication and this piece of the form will not match.

Here is an idea: add these two 'fields' to the form below using the simple form block syntax (not all form fields need to be actual fields...) Then we get the styling for free.

app/views/admin/settings/index.html.slim
```diff
@@ -1,27 +1,48 @@
...
+
+section.block
+  h2 = t("quickbooks.accounts.accounts")
+  p.note = t("quickbooks.accounts.note")
+
+  = simple_form_for([:admin, @division],
+      url: admin_settings_path,
+      method: 'patch',
+      html: {class: 'form-horizontal'},
+      wrapper: :horizontal_form_expanded,
```
hooverlunch: What is the purpose of adding this new wrapper? I'd like to make it work with the existing one if possible since one is simpler than two and thus easier to maintain...

config/locales/en/settings.en.yml
```diff
@@ -0,0 +1,15 @@
...
+      accounts: "QuickBooks Accounts"
+      note: "Note: These account settings will only apply to transactions created in the future."
+    connection:
+      connect_to_qb: "Connect to QuickBooks"
+      connected: "Connected"
+      connection: "QuickBooks Connection"
+      disconnect: "Disconnect"
+      disconnect_from_qb: "Disconnect from QuickBooks"
+      not_connected: "Not Connected"
+      status: "QuickBooks Status"
```
hooverlunch: When you refactor to move these fields into the form, some of this should move into the simple_form translations.

app/views/admin/settings/index.html.slim
emitche: These labels are longer than others in the application. I can use the other wrapper. I have no more time for refactoring in this issue, however, so it remains as is for now.

app/views/admin/settings/index.html.slim
emitche: The save button is part of that section of the page. Placing the save button outside of the context of that form adds a confusing button that does not relate at all to the QuickBooks connection settings. If a user hit the save button when there was nothing to save, the page would break.

I welcome you to talk with @rgalusca  about styling of pages in the future. I am keeping this as if for the time being as I do not have remaining time to address style changes in this feature.

PR #291, opened 2017-03-29
6146 bug fix csv export

emitche:
Fix CSV export in basic projects, divisions, loans, organizations, and people grids

### DIFF
--- a/app/views/admin/basic_projects/_basic_projects_grid_definition.html.slim
+++ b/app/views/admin/basic_projects/_basic_projects_grid_definition.html.slim
@@ -1,5 +1,5 @@
 = define_grid(@basic_projects_grid, hide_csv_button: true,
-    html: {class: 'rowlink table-hover', data: {link: @csv_mode ? nil : 'row'}}) do |g|
+    html: {class: 'rowlink table-hover', data: {link: 'row'}}) do |g|
 
     - g.blank_slate { set_no_grid_records_flag }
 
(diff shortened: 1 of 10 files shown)


app/views/admin/basic_projects/_basic_projects_grid_definition.html.slim
```diff
@@ -1,5 +1,5 @@
 = define_grid(@basic_projects_grid, hide_csv_button: true,
-    html: {class: 'rowlink table-hover', data: {link: @csv_mode ? nil : 'row'}}) do |g|
+    html: {class: 'rowlink table-hover', data: {link: 'row'}}) do |g|
```
emitche: This condition has no impact on successful exporting to CSV. Data link information does not show up in the table when exporting.

app/views/admin/divisions/_divisions_grid_definition.html.slim
```diff
@@ -1,5 +1,5 @@
 = define_grid(@divisions_grid, show_filters: :show, hide_csv_button: true,
-  html: { class: 'rowlink table-hover', data: { link: @csv_mode ? nil : 'row' } }) do |g|
+  html: { class: 'rowlink table-hover', data: { link: 'row' } }) do |g|
```
emitche: Do you prefer `{class: 'class'}` or `{ class: 'class' }` for syntax? I've kept each of these files with the same syntax they were before I edited them.

PR #292, opened 2017-03-30
5870 tree in group dropdown

linesarefuzzy:


app/models/project_group.rb
```diff
@@ -62,6 +62,11 @@ def summary_or_none
     summary.blank? ? "[#{I18n.t("none")}]" : summary.to_s
   end
 
+  # For the "Group" dropdown in step edit form
```
hooverlunch: I'd prefer to avoid comments that specifically name pieces of the interface because later if something else uses this (likely) then the comment becomes incomplete. Can you please change to something more generic. Also the method name could perhaps be `indented_option_label` to be a bit more clear about the function.

app/models/project_group.rb
```diff
@@ -99,6 +104,11 @@ def filtered_children
     end
   end
 
+  # Returns a flat array of the descendant groups, pre-ordered. For the "Group" dropdown in step edit form.
```
hooverlunch: Same thing here.

app/models/project_group.rb
```diff
@@ -62,6 +62,11 @@ def summary_or_none
     summary.blank? ? "[#{I18n.t("none")}]" : summary.to_s
   end
 
+  # For the "Group" dropdown in step edit form
+  def option_label
+    ("&nbsp; &nbsp; " * [depth - 1, 0].max).html_safe + summary_or_none
```
hooverlunch: It's more efficient to use `<<` instead of `+` in this case b/c it appends to the string instead of creating a whole new one.

spec/models/project_group_spec.rb
```diff
@@ -104,30 +127,15 @@
...
-        expect(g6.reload.filtered_children).to eq [g6_g2, g6_g1]
+        expect(@root2.reload.filtered_children).to eq [@g2, @g6, @g1, @s2, @g5, @g3, @g4, @s1]
+        expect(@g1.reload.filtered_children).to eq [@g1_s4, @g1_s2, @g1_s1, @g1_s3]
+        expect(@g6.reload.filtered_children).to eq [@g6_g2, @g6_g1]
+      end
+    end
+
+    describe "self_and_descendant_groups_preordered" do
+      it "should return flat pre-ordered array of groups" do
+        expect(@root2.reload.self_and_descendant_groups_preordered).to eq [@root2, @g2, @g6, @g6_g2, @g6_g1, @g1, @g5, @g3, @g4]
```
hooverlunch: Nice refactoring and test here. Can there please also be a couple for the `option_label` method? Should be quick.. just one for the root, first level, and interior level nodes.

hooverlunch:
Crap I didn't mean to put the red X on there! Sorry!

PR #295, opened 2017-04-05
6133 child groups

emitche:
- Allow parent group to be changed on existing group or new group
- Add specs to test parent group changes

### DIFF
--- a/app/controllers/admin/project_groups_controller.rb
+++ b/app/controllers/admin/project_groups_controller.rb
@@ -54,6 +54,7 @@ class Admin::ProjectGroupsController < Admin::AdminController
 
   def render_modal_partial(status: 200)
     link_params = params.slice(:project_id, :parent_id)
+    find_possible_parent_groups
     render partial: "admin/project_groups/modal_content", status: status
   end
 
@@ -63,6 +64,11 @@ class Admin::ProjectGroupsController < Admin::AdminController
 
   def find_timeline_entry
     @entry = ProjectGroup.find(params[:id])
+    @project = Project.find(@entry.project_id)
     authorize @entry
+  end
+
+  def find_possible_parent_groups
+    @parents = @project.timeline_groups_preordered
   end
 end
--- a/app/controllers/admin/project_groups_controller.rb
+++ b/app/controllers/admin/project_groups_controller.rb
@@ -69,6 +69,6 @@ class Admin::ProjectGroupsController < Admin::AdminController
   end
 
   def find_possible_parent_groups
-    @parents = @project.timeline_groups_preordered
+    @parents = @project.timeline_groups_preordered.to_a - [@entry]
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/admin/project_groups_controller.rb
```diff
@@ -54,6 +54,7 @@ def destroy
 
   def render_modal_partial(status: 200)
     link_params = params.slice(:project_id, :parent_id)
+    @parents = @project.timeline_groups_preordered.to_a - [@entry] - @entry.descendants.to_a
```
hooverlunch: I think you can use `@entry.self_and_descendants.to_a` instead of `[@entry] - @entry.descendants.to_a`.

app/views/admin/project_groups/_modal_content.html.slim
```diff
@@ -12,8 +12,8 @@ div.modal-body
 
     = error_notification(f, dont_show_default: true)
 
-    - if @entry.parent.present?
-      = f.input :parent_id, collection: [@entry.parent], label_method: :summary, disabled: true
+    = f.input :parent_id, collection: @parents.map { |i| [indented_option_label(i, :summary_or_none), i.id] },
+      include_blank: false
```
hooverlunch: Nice job to use the new `indented_option_label` function! I'm glad the existence of this managed to get communicated.

hooverlunch: One thing, I think `i` doesn't make sense as a loop variable. Perhaps `p` for parent or `g` for group?

hooverlunch: Question: How are we handling the case where the parent is root? I'm guessing root shows up in `@parents` and then since we're using `summary_or_none` it is shown as "[None]" or so in the dropdown? Just checking... Always important to think about the edge cases.

spec/models/project_group_spec.rb
```diff
@@ -76,6 +76,25 @@
...
+    end
+
+    it 'should disallow changing the parent of a regular group to nil' do
+      expect { child1.update(parent: nil) }.to raise_error(ArgumentError)
+    end
+
+    it 'should allow the parent to be changed otherwise' do
+      expect(child1.update(parent: child2)).to be true
+    end
+  end
```
hooverlunch: Looks nice!

app/views/admin/project_groups/_modal_content.html.slim
emitche: Yes, `[None]` is the root group. It is always an option. Since the root group is not editable from the interface, when a group is edited, the root is always the base and an available option.

app/controllers/admin/project_groups_controller.rb
emitche: I was unable to get `@entry.self_and_descendants` working. It either is empty when used as `[@entry.self_and_descendants]` or results in an error when used as `@entry.self_and_descendants.to_a`.

hooverlunch: That's fine. You can leave as is.

emitche:
The failing test is related to the rspec test not equating `&` with `&amp;` Visually those are the same thing. I am not sure how to fix the test for that.

PR #298, opened 2017-04-10
6185 add ping page

X0nic:


X0nic:
Are there any other checks we would like to add from Mealyzer @linesarefuzzy

app/views/ping/index.text.erb
```diff
@@ -0,0 +1,6 @@
+<% content_for(:title){ "Ping" } -%>
+Madeline
+App is running.
+Delayed Job is <%= @dj ? "" : "NOT " %>running.
+<%= "There are no stuck jobs." if !@stuck %>
+<%= "Everything is OK." if @ok %>
```
hooverlunch: Perhaps this would be a bit better for the last two lines:

```
There <%= @stuck ? "ARE" : "are no" %> stuck jobs.
Everything is <%= @ok ? "OK" : "NOT OK" %>.
```

PR #300, opened 2017-04-11
6017 display media type on media attachment

emitche:
- Rearrange what is displayed for media information
- Display content type on media item
- Add icons to non-image files
- On hover of shortened filename, show full filename
- If no media items present in a section, do not display the section

### DIFF
--- a/app/helpers/media_helper.rb
+++ b/app/helpers/media_helper.rb
@@ -40,4 +40,12 @@ module MediaHelper
       end
     end
   end
+
+  def media_visuals(media)
+    media.select do |media_item| media_item.visual? end
+  end
+
+  def media_documents(media)
+    media.select do |media_item| !media_item.visual? end
+  end
 end
--- a/app/helpers/media_helper.rb
+++ b/app/helpers/media_helper.rb
@@ -3,17 +3,21 @@ module MediaHelper
     if media_item.thumbnail?
       image_tag(media_item.item.thumb.url)
     else
-      media_extension = File.extname(media_item.item.file.path).downcase
-      # content_tag(:div, class: "generic-thumbnail") do
-        content_tag(:div, media_extension, class: "extension")
-        # concat(image_tag("file-blank.png"))
-      # end
+      # media_extension = File.extname(media_item.item.file.path).downcase
+      # # content_tag(:div, class: "generic-thumbnail") do
+      # content_tag(:div, media_extension, class: "extension")
+      # # concat(image_tag("file-blank.png"))
+      # # end
+      content_tag(:div, media_item.kind_value.capitalize, class: "media-block")
     end
   end
 
   def media_title(media_item)
     ext = File.extname(media_item.item.file.path)
     file_name = truncate(File.basename(media_item.item.file.path, ext), length: 15)
-    content_tag(:span, "#{file_name}#{ext.downcase}")
+    content_tag(:div, class: "media-title") do
+      concat(content_tag(:span, file_name))
+      concat(content_tag(:div, ext.downcase))
+    end
   end
 end
--- a/app/helpers/media_helper.rb
+++ b/app/helpers/media_helper.rb
@@ -22,7 +22,7 @@ module MediaHelper
   end
 
   def media_icon_class(media_item)
-    if media_item.visual_media?
+    if media_item.visual?
       if media_item.kind_value == "video"
         "file-video-o"
       else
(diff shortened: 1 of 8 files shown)


app/models/media.rb
```diff
@@ -51,11 +51,21 @@ def alt
...
+    kind_value == 'image' ? (item_content_type.include? "image") : false
   end
 
   def recalculate_loan_health
     # Only submit a job if it is a Loan
     RecalculateLoanHealthJob.perform_later(loan_id: media_attachable_id) if media_attachable_type == 'Project'
   end
+
+  def visual_media?
+    if kind_value == "image"
```
emitche: I had trouble getting a condensed version of this working. Recommendations welcome here. Thank you.

app/helpers/media_helper.rb
```diff
@@ -3,11 +3,43 @@ def media_thumbnail(media_item)
...
     else
-      media_extension = File.extname(media_item.item.file.path).downcase
-      # content_tag(:div, class: "generic-thumbnail") do
-        content_tag(:div, media_extension, class: "extension")
-        # concat(image_tag("file-blank.png"))
-      # end
+      content_tag(:div, class: "media-block") do
+        concat(content_tag(:div, media_item.kind_value.capitalize))
+        concat(sanitize("<i class='fa fa-#{media_icon_class(media_item)}'></i>", tags: %w(i),
+          attributes: %w(class)))
```
hooverlunch: Not sure why sanitize is needed here. How about just `content_tag(:i, '', class: "fa fa-#{media_icon_class(media_item}")`?

hooverlunch: Or even better would be to add a general helper to application_helper.rb called `icon_tag` that takes the part after `fa-` as an argument and gives back the icon tag. You could use the above within there...

app/helpers/media_helper.rb
```diff
@@ -3,11 +3,43 @@ def media_thumbnail(media_item)
...
+      end
+    end
+  end
+
+  def media_title(media_item)
+    ext = File.extname(media_item.item.file.path)
+    file_name = truncate(File.basename(media_item.item.file.path, ext), length: 12)
+    full_name = File.basename(media_item.item.file.path)
+    content_tag(:div, class: "media-title", title: "#{full_name}") do
```
hooverlunch: Please remove the `"#{...}"`

app/helpers/media_helper.rb
```diff
@@ -3,11 +3,43 @@ def media_thumbnail(media_item)
...
+      when ".pdf"
+        icon_class = "file-pdf-o"
+      when ".doc"
+        icon_class = "file-word-o"
+      when ".docx"
+        icon_class = "file-word-o"
+      else
+        icon_class = "file-text-o"
+      end
+      return icon_class
```
hooverlunch: You can remove this line and all the `icon_class = `s and the method should just work. Ruby returns whatever the last-evaluated expression was in the method.

hooverlunch: Nice job on these methods!

app/models/media.rb
```diff
@@ -51,11 +51,21 @@ def alt
   end
 
   def thumbnail?
-    kind_value == 'image'
+    kind_value == 'image' ? (item_content_type.include? "image") : false
```
hooverlunch: `kind_value == 'image' && item_content_type.include? 'image'`

app/models/media.rb
hooverlunch: How about `%w(image video).include?(kind_value)`?

Also how about changing the name to just `visual?` because the media part is redundant?

app/views/admin/media/_media_item.html.slim
```diff
@@ -0,0 +1,16 @@
...
+        i.fa.fa-pencil
+      = link_to admin_media_path(media_item, link_params), class: 'media-action delete',
+        data: { \
+          confirm: t('media.confirm_deletion'),
+          'confirm-title': t('media.delete'),
+          'confirm-cancel': t(:cancel),
+          'confirm-proceed': t(:confirm),
+          'confirm-cancel-class': 'media-action btn cancel btn-default',
+          'confirm-proceed-class': 'media-action btn proceed btn-primary' } do
+        i.fa.fa-trash
```
hooverlunch: Nice refactor here. Good job moving out the appropriate parts into a partial.

emitche:
@hooverlunch: Great feedback here. Thanks a bunch!

emitche:
Of the spots that have media, there were indeed existing headings (Person and Loan).

Logs don't have a media heading, but I don't think a heading is needed. Once media is added, the headings "Visual Media" and "Documents" are added as those media types are added to the log.

This is what that looks like:

image

image

image

PR #301, opened 2017-04-11
6018: Save the state of the questionnaire tree

emitche:
- Allows nodes to remain open on page reload
- When questionnaire is saved, open nodes are remembered


app/assets/javascripts/backbone/views/loan_questionnaires_view.coffee
```diff
@@ -32,6 +32,7 @@ class MS.Views.LoanQuestionnairesView extends Backbone.View
     # Initialize the jqtree
     @tree.tree
       dragAndDrop: false
+      saveState: true
```
hooverlunch: Nice! The docs say "For this to work, you should give each node in the tree data an id field:". I'm assuming this is true here?

PR #306, opened 2017-04-13
6198: Display existing media for person

emitche:
- Show media for a person in person details page

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -65,7 +65,7 @@ class Person < ActiveRecord::Base
   has_many :project_logs, foreign_key: :agent_id, dependent: :nullify
   has_many :timeline_entries, foreign_key: :agent_id, dependent: :nullify
 
-  # The belong interferes with the expected behavior for media attachable
+  # The below interferes with the expected behavior for media attachable
   # has_many :media, foreign_key: :uploader_id, dependent: :nullify
 
   has_one :user, foreign_key: :profile_id, autosave: true, dependent: :destroy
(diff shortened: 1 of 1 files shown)


app/models/person.rb
```diff
@@ -64,7 +64,9 @@ class Person < ActiveRecord::Base
   has_many :representative_loans,  class_name: 'Loan', foreign_key: :representative_id,  dependent: :nullify
   has_many :project_logs, foreign_key: :agent_id, dependent: :nullify
   has_many :timeline_entries, foreign_key: :agent_id, dependent: :nullify
-  has_many :media, foreign_key: :uploader_id, dependent: :nullify
+
+  # The belong interferes with the expected behavior for media attachable
+  # has_many :media, foreign_key: :uploader_id, dependent: :nullify
```
emitche: See Redmine for questions.

app/models/person.rb
```diff
@@ -64,7 +64,9 @@ class Person < ActiveRecord::Base
   has_many :representative_loans,  class_name: 'Loan', foreign_key: :representative_id,  dependent: :nullify
   has_many :project_logs, foreign_key: :agent_id, dependent: :nullify
   has_many :timeline_entries, foreign_key: :agent_id, dependent: :nullify
-  has_many :media, foreign_key: :uploader_id, dependent: :nullify
+
+  # The below interferes with the expected behavior for media attachable
```
emitche: This can be completely, removed instead of commented out, if desired. Longer notes are available in Redmine.

In short, uploader_id was used in the old system, but is not being currently used. It does exist on records.

The shown media are the media attached to a specific person. The below code that is commented out was interfering with the expected behavior of Madeline, which is to show media for the current MediaAttachable object. In this case, that object is a person.

PR #307, opened 2017-04-14
5791 tweaks show logs link

emitche:
Inside calendar step modal:
- Do not show "Show Logs" if there are no logs to display
- If there are logs, show the logs count in "Show/Hide Logs"


rgalusca:
The log count has not been implemented. @hooverlunch, can you maybe do it since Liz is not here next week. I can also have this functionality as a separate tweak in another sprint.

PR #310, opened 2017-04-18
6234 media in show log

emitche:
- Show media in show log
- Inside log list, display media
- Display media inside the expand/hide section of the log
- UI changes to media including displaying items inline with each other


app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -84,3 +84,7 @@
             - if log.private_notes.present?
               .extra
                 = simple_format content_tag(:strong, t('log.private_notes')) << log.private_notes.to_s
+
+            - if log.media.present?
+              .extra.media-container.row
+                = render 'admin/media/index', media: log.media, owner: log
```
hooverlunch: Hooray for reusability!!!

PR #311, opened 2017-04-19
6191 bug error on batch finalize steps

emitche:
- On batch finalize of steps, use update method without validation
- Fix bug where batch finalize does not work for steps without step type
- Make step type value not be null or empty
- Set project groups to have a step type value
- Set project steps to have a step type value

### DIFF
--- a/app/controllers/admin/project_steps_controller.rb
+++ b/app/controllers/admin/project_steps_controller.rb
@@ -4,9 +4,13 @@ class Admin::ProjectStepsController < Admin::AdminController
 
   def new
     @project = Project.find(params[:project_id])
-    @step = ProjectStep.new(project: @project, scheduled_start_date: params[:date],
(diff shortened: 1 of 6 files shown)


app/controllers/admin/project_steps_controller.rb
```diff
@@ -4,9 +4,13 @@ class Admin::ProjectStepsController < Admin::AdminController
...
-    @step = ProjectStep.new(project: @project, scheduled_start_date: params[:date],
-      parent_id: params[:parent_id], schedule_parent_id: params[:schedule_parent_id],
-      step_type_value: 'checkin')
+    @step = ProjectStep.new(
+      project: @project,
+      scheduled_start_date: params[:date],
+      parent_id: params[:parent_id],
+      schedule_parent_id: params[:schedule_parent_id],
+      step_type_value: 'checkin'
+    )
```
hooverlunch: Nice cleanup.

PR #312, opened 2017-04-19
6250 show group in new dependent step

emitche:
- Show precedent's group in new dependent step's modal

### DIFF
--- a/app/assets/javascripts/backbone/views/timeline_table_view.coffee
+++ b/app/assets/javascripts/backbone/views/timeline_table_view.coffee
@@ -88,7 +88,12 @@ class MS.Views.TimelineTableView extends Backbone.View
 
   addDependentStep: (e) ->
     e.preventDefault()
-    @stepModal.new(@projectId, @refresh.bind(@), precedentId: @stepIdFromEvent(e))
+    @stepModal.new(
+      @projectId,
+      @refresh.bind(@),
+      precedentId: @stepIdFromEvent(e),
+      parentId: @stepParentIdFromEvent(e)
+    )
 
   parentId: (e) ->
     @$(e.target).closest(".project-group").data("id")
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/backbone/views/timeline_table_view.coffee
```diff
@@ -88,7 +88,10 @@ class MS.Views.TimelineTableView extends Backbone.View
 
   addDependentStep: (e) ->
     e.preventDefault()
-    @stepModal.new(@projectId, @refresh.bind(@), precedentId: @stepIdFromEvent(e))
+    @stepModal.new(@projectId, @refresh.bind(@),
+      precedentId: @stepIdFromEvent(e),
+      parentId: @stepParentIdFromEvent(e)
+    )
```
hooverlunch: Note the small stylistic change. It's more common to split just the hash arguments across lines while leaving the regular ones inline, unless the regular ones are super long too.

app/views/admin/timeline/_step.html.slim
```diff
@@ -44,7 +44,7 @@
       - if index == latest_logs.length - 1
         div = link_to t("project_logs.more"), "#", data: {action: 'view-logs'}
 
-  td.step-menu-col class="dropup #{classes}" data-id=entry.id
+  td.step-menu-col class="dropup #{classes}" data-id=entry.id data-parent-id=entry.parent_id
```
hooverlunch: Note that I changed `parent.id` to `parent_id`. It's best to use the latter if it's possible since calling `parent.id` may trigger a database query to populate the `parent` object. If you need any other property of the parent, e.g. `summary`, you obviously have to call `parent.summary`, but with IDs you can often just do it directly.

emitche:
The failed test is tiny:
`expected: "&nbsp; &nbsp; Humblebrag 90's typewriter loko"`
`got: "&nbsp; &nbsp; Humblebrag 90&#39;s typewriter loko"`

Visually these would be the same.

emitche:
@rgalusca As we discussed, this should be working now. The failed error is a syntax difference that is not about functionality.

PR #313, opened 2017-04-20
5679 batch adjusting dates with preceding steps

emitche:


app/assets/javascripts/backbone/views/timeline_batch_actions_view.coffee
```diff
@@ -23,6 +24,12 @@ class MS.Views.TimelineBatchActionsView extends Backbone.View
 
     $form.submit()
 
+  checkForDependentSteps: (stepIds) ->
+    @$('#dependent-steps-notice').hide()
+    boxes = @$(".select-step[data-id][data-precedent-id]:checked")
+    if boxes.length > 0
+      @$('#dependent-steps-notice').show()
+
```
hooverlunch: Wow, slick.

app/models/timeline/date_adjustment.rb
```diff
@@ -28,7 +28,10 @@ def days_adjustment
     end
 
     def adjust_scheduled_date(step, days_adjustment)
-      if step.scheduled_start_date && days_adjustment != 0
+      # If step has a precedent step, do not change the date
+      if step.schedule_parent_id
+        false
+      elsif step.scheduled_start_date && days_adjustment != 0
```
hooverlunch: This looks correct but it needs test coverage. I'd suggest a spec that sets up four steps, one of them with a schedule_parent, attempts to adjust the dates of three of them (including the dependent one AND the one it depends on), checks that the operation still works and that all four steps still have the right start dates. (The dependent one should also change by virtue of its dependence, and this is an important thing to double check.)

Also, the proper test driven way to do this would be to comment out the new code, write the test, watch it blow up like it used to, then uncomment your new code and watch it pass. Few things are more satisfying in life than this, IMHO :)

There should already be a spec file for this class, so this would just be adding to it.

If you need to pair on this, let me know.

config/locales/en/loans.en.yml
```diff
@@ -18,6 +18,7 @@ en:
     delete: "Delete Loan"
     delete_responses: "Delete Responses"
     delete_selected: "Delete"
+    dependent_steps_notice: "Some of the selected steps depend on other steps. Their scheduled start dates cannot be adjusted directly."
```
hooverlunch: Nice wording :)

hooverlunch: Actually... if the warning is showing, the operation will still work, right? Why do we need to show a warning at all? Won't the end result be that the date shows as updated?

I guess this only won't work if the depended-upon step is _not_ in the set of steps being adjusted. Should we test for that? I think it should be possible...

If we do, perhaps this should read "Some of the selected steps depend on other steps that are not selected. You can proceed, but the start dates of those steps will not be adjusted."

app/models/timeline/date_adjustment.rb
emitche: Testing is in a separate issue and I will make sure these tests are included.

app/assets/javascripts/backbone/views/timeline_batch_actions_view.coffee
```diff
@@ -23,6 +24,23 @@ class MS.Views.TimelineBatchActionsView extends Backbone.View
 
     $form.submit()
 
+  # Check the selected steps for any steps with precedents
+  # If the step has a precedent step, show a message
+  checkForUnselectedPrecedents: (stepIds) ->
+    @$('#steps-notice').hide()
+
+    dependents = @$(".select-step[data-id][data-precedent-id]:checked")
+    stepIds = @$('.step-ids').val()
```
hooverlunch: There are a few odd things going on with `stepIds` here it seems. First of all, it's passed in as an arg and then overwritten. That's easily enough fixed.

But also it seems like in TimelineSelectStepsView, we're assigning an array of IDs to the value of the `.step-ids` input. If you try this out (I had to to make sure!) you'll see that this sets the input's value to a string of comma separated IDs like "124,224,824" etc.

Then below you're doing `includes`, which is a string function that checks if a substring is present in the string. This is a problem because `"124,224,824".includes(24)` will be true, which is not what you want.

Note that `disabled = stepIds.length < 1` works because it's checking that the string of IDs is empty.

To fix this you could split the ID string by comma and test against that. I would add a method `selectedStepIds` that does this (`(@$('.step-ids').val() || '').split(',')`) and call it from both places.

Then I don't think you need to pass `stepIds` to `checkForUnselected...`.

app/assets/javascripts/backbone/views/timeline_batch_actions_view.coffee
```diff
@@ -23,6 +24,23 @@ class MS.Views.TimelineBatchActionsView extends Backbone.View
...
+    unselectedPrecedentIds = []
+
+    dependents.each ->
+      stepId = $(this).data('id')
+      precedentId = $(this).data('precedent-id')
+      unselectedPrecedentIds.push(precedentId) unless stepIds.includes(precedentId)
+
+    if unselectedPrecedentIds.length > 0
+      @$('#steps-notice').show()
```
emitche: Yes, this is true.

PR #314, opened 2017-04-24
6081 qb latest changes

X0nic:


app/models/accounting/account.rb
```diff
@@ -16,7 +16,19 @@
...
   has_many :transactions, inverse_of: :account, foreign_key: :accounting_account_id, dependent: :destroy
+
+  def self.find_or_create_from_qb_object(qb_object)
+    account = create_with(name: qb_object.name).find_or_create_by qb_id: qb_object.id
+    account.tap do |a|
+      a.update_attributes!(
+        name: qb_object.name,
+        qb_account_classification: qb_object.classification,
+        quickbooks_data: qb_object.as_json
+      )
```
hooverlunch: Kind of odd to set the name twice. Perhaps do `find_or_initialize_by` and then you won't need `create_with`?

app/models/accounting/quickbooks/account_fetcher.rb
```diff
@@ -2,7 +2,7 @@ module Accounting
   module Quickbooks
     class AccountFetcher < FetcherBase
       def types
-        ['Account']
+        [Accounting::Account::QB_TRANSACTION_TYPE]
```
hooverlunch: Yay

app/models/accounting/quickbooks/connection.rb
```diff
@@ -50,6 +50,17 @@ def realm_id
         data[:realm_id] unless data.blank?
       end
 
+      def last_updated_at
+        DateTime.parse(data[:last_updated_at]) unless data.blank? || data[:last_updated_at].blank?
```
hooverlunch: How are we handling time zones here? This app is globally in Eastern time right now but that will undoubtedly change. I'd like to standardize on using `Time.zone.parse` in most places. We should also ensure that the date that gets stored in the JSON below includes a timezone (we can convert them all to UTC if you like and include `UTC`, or just write out the local timezone, but one or the other needs to be there). So perhaps explicitly doing the string conversion instead of relying on the default one?

app/models/accounting/quickbooks/connection.rb
```diff
@@ -50,6 +50,17 @@ def realm_id
...
+      end
+
+      def last_updated_at=(date)
+        root = Division.root
+        root.quickbooks_data = data.merge(last_updated_at: date)
```
hooverlunch: See above. Let's include timezone here please if it's not already.

app/models/accounting/quickbooks/full_fetcher.rb
```diff
@@ -0,0 +1,20 @@
...
+  module Quickbooks
+    class FullFetcher
+      attr_reader :qb_connection
+
+      def initialize(qb_connection = Division.root.qb_connection)
+        @qb_connection = qb_connection
+      end
+
+      def fetch_all
+        started_fetch_at = DateTime.current
```
hooverlunch: Can we use `Time.zone.now` please.

app/models/accounting/quickbooks/updater.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    class Updater
+      attr_reader :qb_connection
+
+      def initialize(qb_connection = Division.root.qb_connection)
+        @qb_connection = qb_connection
+      end
+
+      def update(updates_since: nil)
+        update_started_at = DateTime.current
```
hooverlunch: `Time.zone.now` as well here please, unless there is a reason not to.

app/models/accounting/transaction.rb
```diff
@@ -26,11 +26,20 @@
...
-  TRANSACTION_TYPES = %w(JournalEntry Deposit Purchase).freeze
+  QB_TRANSACTION_TYPES = %w(JournalEntry Deposit Purchase).freeze
 
   belongs_to :account, inverse_of: :transactions, foreign_key: :accounting_account_id
   belongs_to :project, inverse_of: :transactions, foreign_key: :accounting_account_id
 
+  def self.find_or_create_from_qb_object(qb_object)
+    transaction_type = qb_object.class.name.demodulize
+
+    transaction = find_or_create_by qb_transaction_type: transaction_type, qb_id: qb_object.id
```
hooverlunch: Use `find_or_initialize_by` here also to save a DB operation?

config/locales/en/settings.en.yml
```diff
@@ -10,3 +10,8 @@ en:
       connection: "QuickBooks Connection"
       disconnect: "Disconnect"
       not_connected: "Not Connected"
+      never_synced: "Never"
+      full_sync: "Full Sync"
+      disconnect_message: "Your QuickBooks account has been successfully disconnected."
+      full_sync_message: "Your QuickBooks data has been syncronized"
```
hooverlunch: Period at end please.

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'rails_helper'
+
+RSpec.describe Accounting::Quickbooks::Updater, type: :model do
+  let(:connection) { instance_double(Accounting::Quickbooks::Connection, last_updated_at: last_updated_at) }
+  subject { described_class.new(connection) }
+
+  let(:generic_service) { instance_double(Quickbooks::Service::ChangeDataCapture, since: double(all_types: [])) }
```
hooverlunch: Please group `let`s together, then subject, then blank line then before, blank line, then describe.

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  describe '#update' do
+    context 'with updates_since is nil' do
+      context 'when last_updated_at is nil' do
+        let(:last_updated_at) { nil }
+
+        it 'throws error' do
+          expect { subject.update }.to raise_error(Accounting::Quickbooks::FullSyncRequiredError)
+        end
+      end
+      context 'when last_updated_at is 31 days ago' do
```
hooverlunch: Blank line between contexts please.

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      context 'when last_updated_at is 30 days ago' do
+        let(:last_updated_at) { 30.days.ago }
+
+        it 'does not throw error' do
+          expect { subject.update }.not_to raise_error
+        end
+      end
+    end
+
+    context 'with updates_since is 31 days ago' do
```
hooverlunch: s/with/when/

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  subject { described_class.new(connection) }
+
+  let(:generic_service) { instance_double(Quickbooks::Service::ChangeDataCapture, since: double(all_types: [])) }
+  before do
+    allow(subject).to receive(:service).and_return(generic_service)
+    allow(connection).to receive(:last_updated_at=)
+  end
+
+  describe '#update' do
+    context 'with updates_since is nil' do
```
hooverlunch: s/with/when

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      context 'when last_updated_at is 30 days ago' do
+        let(:last_updated_at) { 30.days.ago }
+
+        it 'throws error' do
+          expect { subject.update(updates_since: 31.days.ago) }.to raise_error(Accounting::Quickbooks::FullSyncRequiredError)
+        end
+      end
+    end
+
+    context 'with updates_since is 30 days ago' do
```
hooverlunch: s/with/when

app/models/accounting/quickbooks/updater.rb
```diff
@@ -0,0 +1,77 @@
+module Accounting
+  module Quickbooks
+    class FullSyncRequiredError < StandardError; end
+
+    class Updater
```
hooverlunch: Really need a comment on this class. Please try to put comments on all classes. What does it do? How is it different than FullFetcher? FullFetcher needs a comment too please.

spec/models/accounting/quickbooks/updater_spec.rb
hooverlunch: Although I'm not seeing where the `updates_since` param is actually getting used. Can we ditch it? It's adding a lot of complexity to these tests.

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      context 'when last_updated_at is 30 days ago' do
+        let(:last_updated_at) { 30.days.ago }
+
+        it 'does not throw error' do
+          expect { subject.update(updates_since: 30.days.ago) }.not_to raise_error
+        end
+      end
+    end
+  end
+end
```
hooverlunch: These tests are thorough but they only cover whether the FullSync error is thrown.

There is a lot of other logic in this model that needs coverage. Can you not stub the `service` method and return some fake transaction data and test that the right things happen? I'm not sure how you are even building this code without tests of this sort. How do you know it works?

If you want to pair on how to set up these tests, I'm happy to. But the fact that this is an external API is no excuse for not testing our model logic here. Would you agree?

app/models/accounting/quickbooks/connection.rb
hooverlunch: Update: I see you're using .utc in other places with quickbooks data. I think we need to pair on this to get it straight. I don't want this to get ugly. Can you please reach out to me.

spec/models/accounting/quickbooks/updater_spec.rb
hooverlunch: As discussed: test creation, update, and deleteing of existing objects in the update process

app/controllers/concerns/transaction_listable.rb
```diff
@@ -3,8 +3,10 @@ module TransactionListable
 
   def initialize_transactions_grid(project_id = nil)
     begin
-      ::Accounting::Quickbooks::AccountFetcher.new.fetch
-      ::Accounting::Quickbooks::TransactionFetcher.new.fetch
+      ::Accounting::Quickbooks::Updater.new.update
+    rescue Accounting::Quickbooks::FullSyncRequiredError => e
+      Rails.logger.error e
+      flash.now[:error] = t('quickbooks.full_sync_required', settings: view_context.link_to(t('menu.settings'), admin_settings_path)).html_safe
```
X0nic: Maybe I am doing something wrong, but this wouldn't work for me.

X0nic:
Nearly done all the changes, but still need to add some additional test coverage.

X0nic:
It might make sense to merge this for the deployment @hooverlunch. I can finish up the new specs tomorrow, without holding things up.

hooverlunch:
Ok. Can you start a new issue for the specs then please.

app/models/accounting/quickbooks/account_fetcher.rb
```diff
@@ -1,13 +1,14 @@
 module Accounting
   module Quickbooks
+
+    # Responsible for grabbing all quickbooks accounts and inserting or updating Accounting::Account
```
hooverlunch: YES

app/models/accounting/quickbooks/transaction_fetcher.rb
```diff
@@ -1,13 +1,14 @@
 module Accounting
   module Quickbooks
+
+    # Responsible for grabbing all quickbooks accounts and inserting or updating Accounting::Account
```
hooverlunch: I love all these!

app/controllers/admin/accounting/quickbooks_controller.rb
```diff
@@ -9,15 +9,15 @@ def oauth_callback
     authorize :'accounting/quickbooks', :oauth_callback?
     @header_disabled = true
 
-    Division.root.quickbooks_save(access_token: qb_access_token, params: params)
+    Accounting::Quickbooks::Connection.create_from_access_token(access_token: qb_access_token, division: Division.root, params: params)
```
hooverlunch: Good idea to add division

app/models/division.rb
```diff
@@ -65,6 +65,8 @@ class Division < ActiveRecord::Base
   has_many :self_and_descendants, through: :descendant_hierarchies, source: :descendant
   has_many :self_and_ancestors, through: :ancestor_hierarchies, source: :ancestor
 
+  has_one :qb_connection, class_name: 'Accounting::Quickbooks::Connection', dependent: :destroy
+
```
hooverlunch: Yay!

PR #315, opened 2017-04-27
6184 ensure loan health is current

X0nic:


app/jobs/recalculate_all_loan_health_job.rb
```diff
@@ -0,0 +1,11 @@
+class RecalculateAllLoanHealthJob < ActiveJob::Base
+  queue_as :default
+
+  def perform
+    # Create a job for each loan recalculation.
+    # The granularity makes it easier to see what loans have problems when they happen.
+    Loan.pluck(:id).each do |loan_id|
```
X0nic: The question is: Should we recalculate all loans or just the loans that have ever had a recalculation done. In theory, the results should be the same. In the other case, we could have orphaned LoanHealthChecks, so I decided with this way.

app/views/ping/index.text.erb
```diff
@@ -2,5 +2,6 @@
 Madeline
 App is running.
 Delayed Job is <%= @dj ? "" : "NOT " %>running.
+There are <%= @count %> jobs in the queue.
```
X0nic: Not necessary, but I was interested in the number for my testing. I thought it might be nice to have.

X0nic:
I started down the path my and @linesarefuzzy talked about, with adding a column so we don't need a nightly job. Even with that method it still did require one query for the `has_late_steps` so I backed out and did this.

Gemfile
```diff
@@ -151,6 +149,9 @@ group :development do
   gem 'capistrano-rails', '~> 1.1'
   gem 'capistrano-passenger'
 
+  # Improved syntax for creating cron jobs
+  gem 'whenever', '~> 0.9', require: false
```
hooverlunch: This was already in the Gemfile. I fixed the dupe and pegged to a version also. You are right to have it in the dev only block though I think :)

app/views/ping/index.text.erb
hooverlunch: Very nice!

app/jobs/recalculate_all_loan_health_job.rb
hooverlunch: Sounds good. We'll see how it scales.

PR #316, opened 2017-05-08
6350 add additional specs for the quickbooks updater

X0nic:


spec/factories/accounting_transactions.rb
```diff
@@ -4,4 +4,10 @@
     qb_transaction_type { Accounting::Transaction::QB_TRANSACTION_TYPES.sample }
     quickbooks_data { {} }
   end
+
+  factory :journal_entry, class: 'Accounting::Transaction' do
+    sequence(:qb_id)
+    qb_transaction_type { 'JournalEntry' }
+    quickbooks_data { {} }
```
hooverlunch: I don't think these need to be in Procs if they're static values.

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -35,5 +44,73 @@
...
+
+        expect(transaction.qb_transaction_type).to eq 'JournalEntry'
+      end
+
+      it 'is created with quickbooks data' do
+        subject.update
+        transaction = Accounting::Transaction.where(qb_id: qb_id).take
+
+        expect(transaction.quickbooks_data).not_to be_empty
+      end
```
hooverlunch: Could you perhaps condense all the above into one? There is a fair amount of duplication here and I think reducing it would make these easier to read...

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -35,5 +44,73 @@
...
+
+      let(:last_updated_at) { 30.days.ago }
+      let(:qb_id) { 34 }
+      let(:journal_entry) { instance_double(Quickbooks::Model::JournalEntry, id: qb_id) }
+
+      subject { described_class.new(connection) }
+
+      it 'creates a new transaction with the proper qb_id' do
+        expect { subject.update }.to change { Accounting::Transaction.where(qb_id: qb_id).count }.by 1
```
hooverlunch: Since we're using DB cleaner there should only be one transaction in the DB, right? So if you do `take` as below and get something then you know it was created with correct ID

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -35,5 +44,73 @@
...
+
+        expect(transaction.quickbooks_data).not_to be_empty
+      end
+    end
+
+    context 'with updated JournalEntry' do
+      before do
+        allow(subject).to receive(:changes).and_return('JournalEntry' => [journal_entry])
+        create(:journal_entry, qb_id: qb_id)
+      end
```
hooverlunch: Please put `before` blocks after `let`s as they read more naturally that way.

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -35,5 +44,73 @@
...
+      let(:qb_id) { 34 }
+      let(:journal_entry) { instance_double(Quickbooks::Model::JournalEntry, id: qb_id) }
+
+      subject { described_class.new(connection) }
+
+      it 'does not create a new transaction' do
+        expect { subject.update }.not_to change { Accounting::Transaction.where(qb_id: qb_id).count }
+      end
+
+      it 'is created with quickbooks data' do
```
hooverlunch: I think this line is wrong?

spec/factories/accounting_transactions.rb
```diff
@@ -4,4 +4,10 @@
     qb_transaction_type { Accounting::Transaction::QB_TRANSACTION_TYPES.sample }
     quickbooks_data { {} }
   end
+
+  factory :journal_entry, class: 'Accounting::Transaction' do
```
hooverlunch: Can you call this `journal_entry_transaction` perhaps? Otherwise it's easily confused with the QB gem JournalEntry model.

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -35,5 +44,73 @@
...
+        transaction = Accounting::Transaction.where(qb_id: qb_id).take
+
+        expect(transaction.quickbooks_data).not_to be_empty
+      end
+    end
+
+    context 'with updated JournalEntry' do
+      before do
+        allow(subject).to receive(:changes).and_return('JournalEntry' => [journal_entry])
+        create(:journal_entry, qb_id: qb_id)
```
hooverlunch: Please use `let` or `let!` as the standard way to create stuff for a test.

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -35,5 +44,73 @@
...
+      end
+
+      let(:last_updated_at) { 30.days.ago }
+      let(:qb_id) { 34 }
+      let(:journal_entry) { instance_double(Quickbooks::Model::JournalEntry, id: qb_id) }
+
+      subject { described_class.new(connection) }
```
hooverlunch: The above 4 lines are the same everywhere right? Move up a level? Perhaps create a `context` for these three JournalEntry specs?

Please try to always avoid repeated lets, subjects, and befores. That's what contexts are for.

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -35,5 +44,73 @@
...
+
+    context 'with deleted JournalEntry' do
+      before do
+        allow(subject).to receive(:changes).and_return('JournalEntry' => [journal_entry])
+        create(:journal_entry, qb_id: qb_id)
+      end
+
+      let(:last_updated_at) { 30.days.ago }
+      let(:qb_id) { 34 }
+      let(:journal_entry) { instance_double(Quickbooks::Model::ChangeModel, id: qb_id, status: 'Deleted') }
```
hooverlunch: In this case, you have a slight delta on the journal entry, which is fine. You can just leave this one `let` here to override the outer one!

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -35,5 +44,73 @@
...
+
+      subject { described_class.new(connection) }
+
+      it 'destroys transaction with the proper qb_id' do
+        expect { subject.update }.to change { Accounting::Transaction.where(qb_id: qb_id).count }.by -1
```
hooverlunch: I'm not sure about using `change` here. Why not just run `subject.update` in a `before` and then `expect(journal_entry_transaction).to be_destroyed`. Isn't that more clear and direct?

X0nic: I tried to do this, but I have a different instance of the object so `destroyed?` was false @hooverlunch

PR #317, opened 2017-05-10
6280 add disbursment date

X0nic:


app/models/accounting/quickbooks/transaction_creator.rb
```diff
@@ -10,9 +10,10 @@ def initialize(root_division = Division.root)
         @principal_account = root_division.principal_account
       end
 
-      def add_disbursement(amount:, loan_id:, memo:, description:, qb_bank_account_id:, qb_customer_id:)
+      def add_disbursement(amount:, loan_id:, memo:, description:, qb_bank_account_id:, qb_customer_id:, date: nil)
         je = ::Quickbooks::Model::JournalEntry.new
         je.private_note = memo
+        je.txn_date = date if date.present?
```
X0nic: Most of this PR is beefing up the specs, but this is the code that was requested in the ticket.

spec/models/accounting/quickbooks/transaction_creator_spec.rb
```diff
@@ -6,24 +6,126 @@
...
+            loan_id: loan_id,
+            memo: memo,
+            description: description,
+            qb_bank_account_id: 89,
+            qb_customer_id: qb_customer_id,
+            date: date
+          )
+        end
+
+        it 'creates JournalEntry with date' do
```
X0nic: This is the one of two specs for the requested feature, everything else is prep for RM:6281

spec/models/accounting/quickbooks/transaction_creator_spec.rb
```diff
@@ -6,24 +6,126 @@
...
+        amount: amount,
+        loan_id: loan_id,
+        memo: memo,
+        description: description,
+        qb_bank_account_id: 89,
+        qb_customer_id: qb_customer_id
+      )
+    end
+
+    context 'generic_service' do
```
X0nic: A complex model is created for the JournalEntry object needed for QBO. These specs are testing that the values are in the correct place. 

I see two potential issues with the code:
1. This code is tightly coupled to the QBO model. I am abusing mocks too much here? Is there value in these kind of specs?
2. There is also quite a bit of duplication. It possibly could be cleaned up with the right pattern.

app/models/accounting/quickbooks/transaction_creator.rb
hooverlunch: Thanks for this explanation. Nice communication!

spec/models/accounting/quickbooks/transaction_creator_spec.rb
```diff
@@ -6,24 +6,126 @@
   let(:class_service) { instance_double(Quickbooks::Service::Class, find_by: [class_ref]) }
   let(:connection) { instance_double(Accounting::Quickbooks::Connection) }
   let(:account) { instance_double(Accounting::Account, qb_id: 98) }
-  subject { described_class.new(instance_double(Division, qb_connection: connection, principal_account: account)) }
+  let(:creator) { described_class.new(instance_double(Division, qb_connection: connection, principal_account: account)) }
 
   before do
-    allow(subject).to receive(:service).and_return(generic_service)
-    allow(subject).to receive(:class_service).and_return(class_service)
+    allow(generic_service).to receive(:create)
```
hooverlunch: I think you could just put `create: nil` in the double, no?

spec/models/accounting/quickbooks/transaction_creator_spec.rb
```diff
@@ -6,24 +6,126 @@
...
+    allow(creator).to receive(:service).and_return(generic_service)
+    allow(creator).to receive(:class_service).and_return(class_service)
```
hooverlunch: Same with these, just add them to the double?

spec/models/accounting/quickbooks/transaction_creator_spec.rb
```diff
@@ -6,24 +6,126 @@
...
+        qb_customer_id: qb_customer_id
+      )
+    end
+
+    context 'generic_service' do
+      it 'creates JournalEntry with 2 line items' do
+        expect(generic_service).to receive(:create) do |arg|
+          expect(arg.line_items.count).to eq 2
+        end
+
```
hooverlunch: Can we remove these blank lines. They don't really delineate a new 'thought'. Calling `subject` is just a necessary technicality.

spec/models/accounting/quickbooks/transaction_creator_spec.rb
hooverlunch: It's great to beef up specs! I do think these are way too long for what they are testing. Really they are all just checking that the data got assigned properly.

Perhaps a good rule of thumb is that the size of the test code should roughly scale according to the amount of functionality being tested..?

Here is how I would do this:

```
  it 'creates JournalEntry with correct data' do
    expect(generic_service).to receive(:create) do |arg|
      expect(arg.private_note).to eq memo
      expect(arg.txn_date).to be_nil
      lis = arg.line_items
      expect(lis.map { |i| i.journal_entry_line_detail.entity.type }.uniq).to eq %w(Customer)
      expect(lis.map { |i| i.journal_entry_line_detail.entity.entity_ref.value }.uniq).to eq [qb_customer_id]
      expect(lis.map { |i| i.posting_type }).to eq %w(Debit Credit)
      expect(lis.map { |i| i.amount }.uniq).to eq [amount]
      expect(lis.map { |i| i.description }.uniq).to eq [description]
      expect(lis.map { |i| i.qb_account_id }).to eq [XXX, XXX]
      expect(lis.map { |i| i.qb_customer_ref }.uniq).to eq [XXX]
    end
    subject
  end
```

Notice how I've added account_id, posting_type, and customer_ref. Why not!? Account ID is especially important since it varies from one to the other.

Thoughts?

To the extent that you are testing different variations on this, you could move the chunk of expects into a helper method, or just test the things that you expect to vary each time.

hooverlunch:
Oh also obviously you'll need to fill in the XXXs above...

spec/models/accounting/quickbooks/transaction_creator_spec.rb
```diff
@@ -2,28 +2,75 @@
...
+        qb_bank_account_id: qb_bank_account_id,
+        qb_customer_id: qb_customer_id,
+        date: date
+      )
+    end
+
+    context 'generic_service' do
+      it 'creates JournalEntry with correct data' do
+        expect(generic_service).to receive(:create) do |arg|
+          expect(arg.line_items.count).to eq 2
```
X0nic: Much more compact, I like this a lot more. :)

spec/models/accounting/quickbooks/transaction_creator_spec.rb
```diff
@@ -2,28 +2,75 @@
...
+          list = arg.line_items
+          expect(list.map(&:amount).uniq).to eq [amount]
+          expect(list.map(&:description).uniq).to eq [description]
+
+          details = list.map { |i| i.journal_entry_line_detail }
+          expect(details.map { |i| i.posting_type }.uniq).to match_array %w(Debit Credit)
+          expect(details.map { |i| i.entity.type }.uniq).to eq %w(Customer)
+          expect(details.map { |i| i.entity.entity_ref.value }.uniq).to eq [qb_customer_id]
+          expect(details.map { |i| i.class_ref.value }.uniq).to eq [loan_id]
+          expect(details.map { |i| i.account_ref.value }.uniq).to match_array [qb_bank_account_id, qb_principal_account_id]
```
hooverlunch: Quick note: I would use eq here instead of match_array because you know what order they will be in and so you can be sure that the Credit transaction has the qb_bank_account_id, etc.

match_array / contains_exactly are best used when you don't know the order.

PR #318, opened 2017-05-11
6281 create customer for transaction

X0nic:
This is a continuation of #317 use this link to see the differences from that PR. https://github.com/sassafrastech/madeline/compare/6280_add_disbursment_date...6281_create_customer_for_transaction

### DIFF
--- a/app/models/accounting/quickbooks/customer.rb
+++ b/app/models/accounting/quickbooks/customer.rb
@@ -1,0 +1,45 @@
+module Accounting
+  module Quickbooks
+
+    # Represents a QBO Customer object and can create a reference object
+    # for a link to this object in a transaction or other QBO object.
+    class Customer
+      attr_reader :qb_connection, :organization
+
+      def initialize(qb_connection:, organization:)
+        @qb_connection = qb_connection
+        @organization = organization
+      end
+
+      # We may be creating a customer here if needed. We return the qb_id and manualy create a reference.
+      # The gem does not implement a helper method for _id like account or class.
+      def reference
+        qb_customer_id = find_or_create_qb_customer_id
+
+        entity = ::Quickbooks::Model::Entity.new
+        entity.type = 'Customer'
+        entity_ref = ::Quickbooks::Model::BaseReference.new(qb_customer_id)
+        entity.entity_ref = entity_ref
+
+        entity
+      end
+
+      private
+
+      def service
+        @service ||= ::Quickbooks::Model::Customer.new(qb_connection.auth_details)
+      end
+
+      def find_or_create_qb_customer_id
+        return organization.qb_id if organization.qb_id.present?
+
+        qb_customer = ::Quickbooks::Model::Customer.new
+        qb_customer.display_name = organization.name
+
+        new_qb_customer = service.create(qb_customer)
+
+        new_qb_customer.id
+      end
+    end
+  end
+end
--- a/app/models/accounting/quickbooks/transaction_creator.rb
+++ b/app/models/accounting/quickbooks/transaction_creator.rb
@@ -15,7 +15,7 @@ module Accounting
         je.private_note = memo
         je.txn_date = date if date.present?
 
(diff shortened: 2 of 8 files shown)


app/models/accounting/quickbooks/transaction_creator.rb
```diff
@@ -67,9 +72,11 @@ def create_line_item(amount:, loan_id:, posting_type:, description:, qb_account_
         line_item
       end
 
-      # We are not creating a customer here, but a customer reference.
+      # We are may be creating a customer here if needed. We return the qb_id and manualy create a reference.
```
hooverlunch: Grammar

app/models/accounting/quickbooks/transaction_creator.rb
```diff
@@ -78,6 +85,17 @@ def create_customer_reference(qb_customer_id:)
...
+        return organization.qb_id if organization.qb_id.present?
+
+        qb_customer = ::Quickbooks::Model::Customer.new
+        qb_customer.display_name = organization.name
+
+        new_qb_customer = customer_service.create(qb_customer)
+
+        new_qb_customer.id
+      end
+
```
hooverlunch: This is feeling a bit bloat-y to me, i.e. violation of SRP. Can we make a new class `Customer' that is initialized with an Organization and has public method `reference`, which calls private method `find_or_create...`?

Then anywhere you need a customer reference you do `Customer.new(organization).reference`?

... and don't forget the class description comment :)

hooverlunch: And in general, please try to be careful about adding too much responsibility to a class like this. If you're creating and managing customers inside something called TransactionCreator, that is a hint that you need a new class.

spec/models/accounting/quickbooks/transaction_creator_spec.rb
```diff
@@ -2,28 +2,109 @@
...
+    context 'and date is supplied' do
+      let(:date) { 3.days.ago }
+
+      it 'creates JournalEntry with date' do
+        expect(generic_service).to receive(:create) do |arg|
+          expect(arg.txn_date).to eq date
+        end
+        subject
+      end
+    end
```
hooverlunch: The above two contexts should move into a new Customer spec. Then this spec can just stub the customer fetching stuff and assume it works properly. :)

spec/models/accounting/quickbooks/customer_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  context 'when qb customer does exist' do
+    let(:organization) { create(:organization, qb_id: qb_customer_id) }
+    let(:service) { instance_double(Quickbooks::Service::Customer, create: nil) }
+
+    it 'does not create a customer' do
+      expect(service).not_to receive(:create)
+      subject.reference
+    end
+
+    it 'returns proper customer reference' do
```
X0nic: Seems like overkill to extract out a shared example for two instances on one test. Any suggestions?

app/models/accounting/quickbooks/customer.rb
```diff
@@ -0,0 +1,45 @@
...
+        qb_customer = ::Quickbooks::Model::Customer.new
+        qb_customer.display_name = organization.name
+
+        new_qb_customer = service.create(qb_customer)
+
+        new_qb_customer.id
+      end
+    end
+  end
+end
```
hooverlunch: Looks great! So much nicer to have this in its own place. Thanks!

hooverlunch:
Also thanks for the compare link. Very thoughtful!

spec/models/accounting/quickbooks/transaction_creator_spec.rb
```diff
@@ -2,28 +2,89 @@
```
emitche: @X0nic I've made a guess about how best to merge the earlier develop (which includes changes from 6280) with this file. If you look over develop once the release goes out and something isn't right, just let me know.

PR #320, opened 2017-05-12
6433 bugfix dependent steps should change with precedent

emitche:


emitche:
@hooverlunch I will review this.

emitche:
@linesarefuzzy - Please let me know why the spec is failing as well. Thank you!

app/models/project_step.rb
```diff
@@ -384,16 +386,20 @@ def handle_finalized_at
...
+  # Copies date changes to schedule_children as appropriate.
   def handle_schedule_children
     return unless persisted? && schedule_children.present? &&
-      (scheduled_start_date_changed? || scheduled_duration_days_changed?)
+      (scheduled_start_date_changed? || scheduled_duration_days_changed? || actual_end_date_changed?)
 
     schedule_children.each do |step|
-      step.scheduled_start_date = scheduled_end_date
-      step.save
+      step.scheduled_start_date = display_end_date + 1
```
linesarefuzzy: It seems `display_end_date` could be nil here as well, if for example the start date and duration on an incomplete step were both set to blank. 

I'm also wondering if it might be better to do something like `schedule_children.each(&: copy_schedule_parent_date)` instead, so as to keep that logic in only one place. Slower though, I suppose.

app/models/project_step.rb
```diff
@@ -384,16 +386,20 @@ def handle_finalized_at
...
+      step.save!
     end
   end
 
+  # Checks that old_start_date is not present if scheduled_start_date is blank.
+  # Sets scheduled_start_date to actual_end_date if actual_end_date
+  # is present but scheduled_start_date is blank.
```
linesarefuzzy: Nice comment.

spec/models/project_step_spec.rb
```diff
@@ -68,7 +73,7 @@
     expect(step.actual_end_date).to be_nil
 
     step.actual_end_date = actual_end_date
-    step.save
+    step.save!
```
linesarefuzzy: Nice job catching these.

emitche:
Thanks for the awesome teamwork on this issue. :)

PR #321, opened 2017-05-13
Translations

emitche:


config/locales/es/for_use_madeline-system_project_step_move_es.yml
```diff
@@ -1,10 +0,0 @@
-es:
```
emitche: Replaced by timeline step moves.

PR #324, opened 2017-05-23
5802 show status for dashboard projects

emitche:
- Show loan status as an icon
- Show tooltips when hovering over loan status icon
- Show loan status icons and tooltips for all users in dashboard


emitche:
I paired with Fuzzy on this. Regular Bootstrap tooltips were not working. However, popups are in the system (which are an extension of tooltips). So this feature utilizes popovers to display tooltips for the status of a loan.

app/helpers/loans_helper.rb
```diff
@@ -2,4 +2,29 @@ module LoansHelper
...
+
+  def health_status_message(loan)
+    if loan.loan_health_check.healthy?
+      status_message = {
+        status: "healthy",
+        message: I18n.t("health_status.healthy"),
+        warnings: []
+      }
+    else
+      warnings = loan.loan_health_check.health_warnings.collect {|w| I18n.t("health_status.warnings.#{w}")}
```
hooverlunch: Please use `map` instead of `collect`. They are the same but `map` is shorter. 

Also please put spaces inside `{}`s.

app/helpers/loans_helper.rb
```diff
@@ -2,4 +2,29 @@ module LoansHelper
   def back_to_loans
     session[:loans_path] || loans_path
   end
+
+  def health_status_message(loan)
```
hooverlunch: This method only needs a health_check object, not a full loan. Best to depend on the smallest amount of knowledge possible.

app/helpers/loans_helper.rb
```diff
@@ -2,4 +2,29 @@ module LoansHelper
...
+      warnings = loan.loan_health_check.health_warnings.collect {|w| I18n.t("health_status.warnings.#{w}")}
+      status_message = {
+        status: "unhealthy",
+        message: I18n.t('health_status.unhealthy'),
+        warnings: warnings
+      }
+    end
+  end
+
+  def health_status_info(loan)
```
hooverlunch: The arg name of this method is misleading I think, as it can also be getting BasicProjects, right? Change to `project`?

app/models/project.rb
```diff
@@ -122,4 +122,14 @@ def project_events(order_by = "Completed IS NULL OR Completed = '0000-00-00', Co
       !p.completed && p.project_logs.empty? && p.date <= Date.today
     end
   end
+
+  def health_status_available?
+    if type == "Loan"
```
hooverlunch: Checking the type of an object in a conditional like this is usually an anti-pattern. The better way to do this is as follows:

1. In the parent class (Project), just return `false` in this method, very simple.
2. In the child class (Loan), override this method and return `!loan_health_check.nil?`.

Note also that `blank?` isn't needed in this case because blank only makes sense if the object defines `empty?`. Please see this: https://apidock.com/rails/Object/blank%3F.

Using blank? in the wrong place can lead to weird bugs.

app/views/admin/dashboard/_loan_health_message.html.slim
```diff
@@ -0,0 +1,15 @@
...
+        h4 = status_message[:message]
+- else
+  .health-message
+    i.fa.fa-exclamation-triangle.ms-tooltip
+    .message.hidden
+      .tooltip-content
+        h4 = status_message[:message]
+        ul
+          - status_message[:warnings].each do |warning|
+            li = warning
```
hooverlunch: There is a fair amount of repeated code in this partial. Could you assign the appropriate icon to a var at the top and then just do a conditional for displaying the `ul` or not?

app/helpers/loans_helper.rb
emitche: So far we only have status information for loans. But I will change the methods to be project-based, if possible, instead of loan-based.

emitche:
Ready for QA testing prior to merging.

PR #325, opened 2017-05-23
6393: Show logs inside timeline edit step modal

emitche:
- Fix bug that showed a no logs message in edit step modal when logs present


emitche:
This is too tiny for a code review. I will do testing and then merge into develop.

PR #326, opened 2017-05-24
6398 create department with transaction

X0nic:


app/models/accounting/quickbooks/customer.rb
```diff
@@ -21,13 +21,15 @@ def reference
         entity_ref = ::Quickbooks::Model::BaseReference.new(qb_customer_id)
         entity.entity_ref = entity_ref
 
+        organization.update!(qb_id: qb_customer_id)
```
X0nic: Found some bugs in the implementation for organizations

spec/models/accounting/quickbooks/department_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'rails_helper'
```
X0nic: There are _slight_ differences with Company. This one has a `name` not a `display_name`. It also returns a `BaseReference` instead of an entity. Any suggestions on how dry up the differences, or if I should?

app/models/accounting/quickbooks/transaction_creator.rb
```diff
@@ -32,20 +34,24 @@ def add_disbursement(amount:, loan_id:, memo:, description:, qb_bank_account_id:
...
+          qb_customer_ref: qb_customer_ref,
+          qb_department_ref: qb_department_ref
         )
 
         service.create(je)
       end
 
       private
 
-      # Not memoized because organization could vary. Make sure to capture in an ivar,
```
X0nic: Changed to take a param. I don't think the comment make sense. You wouldn't usually memoize a method with a param.

app/models/accounting/quickbooks/department.rb
```diff
@@ -0,0 +1,44 @@
+module Accounting
+  module Quickbooks
+
+    # Represents a QBO Department object and can create a reference object
+    # for a link to this object in a transaction or other QBO object.
```
hooverlunch: Thank you!

Can you add some info to this class and Customer describing what role the Customer and Department objects play in madeline? e.g. Customer == Coop, Department == Division? (Right?)

app/models/accounting/quickbooks/transaction_creator.rb
hooverlunch: 👍

spec/models/accounting/quickbooks/department_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+RSpec.describe Accounting::Quickbooks::Department, type: :model do
+  let(:department_name) { 'A cooperative with a name' }
```
hooverlunch: Change to something more division-like

spec/models/accounting/quickbooks/department_spec.rb
hooverlunch: I had a similar thought when I looked at the Department class, but I think this is fine. There is enough difference and not enough similarity to warrant use of inheritance at this point I think. But let's stay open to that possibility in the future.

PR #328, opened 2017-05-26
6515 overwriting bug

linesarefuzzy:
I feel like @hooverlunch might want to quickly review this when he gets back, but we can go ahead and merge it anyway.


emitche:
Yes, @hooverlunch, can take a look at this next week. I've looked over the code and it looks. Thank you!

PR #331, opened 2017-06-08
5980 add dates to step modal

emitche:


app/views/admin/timeline/_step.html.slim
```diff
@@ -22,7 +22,7 @@
 
   td class="#{classes} step-date step-start-date" data-id=entry.id data-precedent-id=entry.schedule_parent_id
     = entry.display_start_date ? l(entry.display_start_date, format: :short) : ""
-  td class="#{classes} step-date step-end-date" data-id=entry.id data-precedent-id=entry.schedule_parent_id
+  td class="#{classes} step-date step-end-date" data-id=entry.id data-precedent-id=entry.schedule_parent_id data-dependent-step-start-date=entry.dependent_step_start_date
```
emitche: Not sure how to break this line with Slim.

app/assets/javascripts/backbone/views/project_step_modal_view.coffee
```diff
@@ -75,9 +77,55 @@ class MS.Views.ProjectStepModalView extends Backbone.View
...
+    startDateVal = @$('#project_step_scheduled_start_date').val()
+    startDate = new Date(startDateVal)
+
+    durationVal = @$('#project_step_scheduled_duration_days').val()
+    duration = Number(durationVal)
+
+    endDate = new Date(startDateVal)
+    endDate.setDate(endDate.getDate() + duration)
+    endDateMoment = moment(endDate)
+    endDateFormatted = moment(endDateMoment).format(MS.dateFormats.default)
```
hooverlunch: Note what I changed here.

app/views/admin/timeline/_step.html.slim
hooverlunch: Also, can you please change this so that the date is wrapped in `ldate`

config/locales/en/en.yml
```diff
@@ -136,6 +136,7 @@ en:
         old_start_date: "Original Start Date"
         scheduled_duration_days: "Scheduled Duration"
         scheduled_start_date: "Scheduled Start Date"
+        scheduled_end_date: "Scheduled End Date"
```
hooverlunch: Note this was missing. An easy way to catch these is if the label in the form looks like 'Scheduled end date' instead of 'Scheduled End Date' (i.e. not title case).

app/assets/javascripts/backbone/views/project_step_modal_view.coffee
```diff
@@ -75,9 +77,55 @@ class MS.Views.ProjectStepModalView extends Backbone.View
...
+
+    @$(".form-group.project_step_scheduled_end_date").find(".static-text-as-field").
+      html(endDateFormatted)
+
+  setScheduledStartDateOnDependent: ->
+    # Applies to dependent step only
+    # Set start date to the precedent step end date plus 1
+    precedentId = @$('#project_step_schedule_parent_id').val()
+    if precedentId
+      startDate = $(".step-end-date[data-id=#{precedentId}]").data('dependent-step-start-date')
```
hooverlunch: This (`$` instead of `@$`) is generally a no-no but I don't see an easy way around it here and we are over time on this so I guess it has to stay.

emitche:
Thank you for working on the date localization not matching in the simple form custom date field, @hooverlunch. There's an issue for that in Redmine, 6558, which I have assigned to you.

app/views/layouts/base.html.slim
```diff
@@ -15,7 +15,11 @@ html lang="en-US"
 
     javascript:
       I18n.availableLocales = #{json I18n.available_locales};
-      $(function() { new MS.Views.ApplicationView(); });
+      $(function() {
+        new MS.Views.ApplicationView({
+          dateFormats: #{json(I18n.t("date.moment_formats"))}
```
emitche: This does not appear to be working.

app/assets/javascripts/backbone/views/project_step_modal_view.coffee
emitche: If we handed of an object with all of the steps in the timeline in this format, maybe we could get around this. Yes, there is not time in this issue though.

precedentStepId: dependentStartDate

{45: "May 6, 2017", 67: "July 9, 2017"}, etc.

app/views/layouts/base.html.slim
emitche: Fixed.

PR #332, opened 2017-06-08
6574: Tempoarily remove page breaks from loan memo

emitche:
- Temporarily remove page breaks from loan memo
- Temporarily remove headings for page breaks in loan memo
- Keep page breaks on breakeven table and charts
- Fix bug where page breaks were inconsistently being applied in loan memo
- An update for how page breaks appear in loan memo is slated for the next sprint


emitche:
Document heading is still being before certain sections even though page breaks are gone. Will remove those additional headings.

PR #334, opened 2017-06-15
remove erroneous message from dashboard

arisah-ijeoma:


hooverlunch:
Unfortunately the original solution for this issue caused a regression. The correct 'no match' messages were not being displayed on other pages.

I have added some commits to hopefully fix this without a regression. The use of `@no_records_at_all` as a global instance variable was causing the issue because there were two grids on the page and the first one had no records while the other wasn't.

I'm still not super happy with my solution here or with Wice Grid, but I think it works.

PR #336, opened 2017-06-21
6524 active filter on dashboard

emitche:
- Remove active/completed filter for projects in dashboard
- Only show active/open projects


emitche:
@linesarefuzzy We discussed commenting out the code. But the client has already requested us to only show open/active projects. So I removed the filter and its code altogether. If we need to restore it, we can find the code in GitHub. Thanks.

app/models/person.rb
```diff
@@ -100,6 +100,14 @@ def agent_projects
     Project.where("primary_agent_id = ? OR secondary_agent_id = ?", id, id)
   end
 
+  def active_agent_projects
+    # Status values can be found at Loan.status_option_set.options and
+    # BasicProject.status_option_set.options
+    active_statuses = ["active", "changed", "possible", "prospective", "refinanced", "relationship",
+      "relationship_active"]
```
hooverlunch: This should perhaps be a constant on `Project`? Also please switch to `%w(...)` notation, it will fit on one line then.

app/models/person.rb
```diff
@@ -100,6 +100,14 @@ def agent_projects
...
+    agent_projects.where(status_value: active_statuses)
```
hooverlunch: Nice job on this model method!

emitche:
Minor adjustment needed. Then this will be ready for QA testing.

PR #338, opened 2017-06-27
6435 filter out loans not yet finalized

arisah-ijeoma:


app/models/calendar_event.rb
```diff
@@ -85,8 +85,10 @@ def self.project_date_filter(range, scope = Project)
   end
 
   def self.project_step_date_filter(range, scope = ProjectStep)
-    scope.where("actual_end_date BETWEEN :first AND :last OR scheduled_start_date BETWEEN :first AND :last "\
-      "OR old_start_date BETWEEN :first and :last", {first: range.first, last: range.last})
+    scope.
+    where("actual_end_date BETWEEN :first AND :last OR scheduled_start_date BETWEEN :first AND :last "\
+    "OR old_start_date BETWEEN :first and :last", {first: range.first, last: range.last}).
+    where(is_finalized: true)
```
emitche: @arisah-ijeoma Sometimes we will need to see the steps that are not finalized in the calendar. So this needs to be its own filter/method.

In this feature, only finalized steps show up in the general calendar. But steps not finalized would still show up in a loan's or project's calendar. Dashboard will show finalized only. (I'm going to add another issue for the dashboard specifically).

Thank you!

emitche: Let me check in with the client on this first to confirm that unfinalized steps are shown in loan calendar. Thank you!

emitche:
I've confirmed with the client that we do want to show unfinalized steps in a specific loan calendar.

Main general calendar - no unfinalized steps shown, finalized steps only
Dashboard calendar - no unfinalized steps shown, finalized steps only
Loan calendar - show unfinalized steps in addition to finalized steps

This feature is just for the main/general calendar. But the solution needs to be reusable. We should be able to show/hide the unfinalized steps depending on the context.

arisah-ijeoma:
I think some of your changes are in here @emitche

app/models/calendar_event.rb
```diff
@@ -51,10 +51,10 @@ def self.new_project_end(project)
     project.end_date ? new.initialize_project_end(project) : nil
   end
 
-  def self.filtered_events(date_range: nil, project_filter: nil, project_scope: Project, step_scope: ProjectStep)
+  def self.filtered_events(date_range: nil, project_filter: nil, project_scope: Project, step_scope: ProjectStep, project_id: nil)
```
emitche: Try to keep lines no longer than 100 characters long.

There's a few lines where this happens, so just wrap them, as much as is possible.

Usual syntax is below:

````
def some_method
    events = variable(really long params,
        more params)
````

File: app/models/calendar_event.rb [hunk 3]
```diff
@@ -84,9 +84,15 @@ def self.project_date_filter(range, scope = Project)
                 {first: range.first, last: range.last})
   end
 
-  def self.project_step_date_filter(range, scope = ProjectStep)
-    scope.where("actual_end_date BETWEEN :first AND :last OR scheduled_start_date BETWEEN :first AND :last "\
-      "OR old_start_date BETWEEN :first and :last", {first: range.first, last: range.last})
+  def self.project_step_date_filter(range, scope = ProjectStep, project_id)
+    scoped_steps = scope.where("actual_end_date BETWEEN :first AND :last OR scheduled_start_date BETWEEN :first AND :last "\
```
emitche: Wrap these parameters so they don't go past 100 characters. Indent the second line of parameters.

````
scoped_steps = scope.where(.....
    .....)
````

**Comment by emitche** [comment]
@arisah-ijeoma This needs some small tweaks. When you complete the tweaks, change the tag to `complete` and assign it to me. Thank you!

**Comment by emitche** [comment]
I plan to make small syntax tweaks to this issue.

**Comment by emitche** [comment]
Please see Redmine: https://redmine.sassafras.coop/issues/6537#note-22. I encountered a bug during QA, so this needs more work.

**Comment by emitche** [comment]
I updated the wrong pull request.

File: app/models/calendar_event.rb [hunk 4]
```diff
@@ -51,10 +51,11 @@ def self.new_project_end(project)
     project.end_date ? new.initialize_project_end(project) : nil
   end
 
-  def self.filtered_events(date_range: nil, project_filter: nil, project_scope: Project, step_scope: ProjectStep)
+  def self.filtered_events(date_range: nil, project_filter: nil, project_scope: Project,
+      step_scope: ProjectStep, project_id: nil)
```
emitche: Use two spaces instead here for tabbing.

app/models/calendar_event.rb
```diff
@@ -65,8 +66,8 @@ def self.project_events_by_date_project_scope(range, scope = Project)
     project_date_filter(range, scope).map(&:calendar_events).flatten
   end
 
-  def self.step_events_by_date_project_filter(date_range: nil, project_filter: nil, scope: ProjectStep)
-    project_step_date_filter(date_range, scope).
+  def self.step_events_by_date_project_filter(date_range: nil, project_filter: nil, scope: ProjectStep, project_id: nil)
```
emitche: Break this into two lines, maybe at project_id.

app/models/calendar_event.rb
```diff
@@ -84,9 +85,17 @@ def self.project_date_filter(range, scope = Project)
                 {first: range.first, last: range.last})
   end
 
-  def self.project_step_date_filter(range, scope = ProjectStep)
-    scope.where("actual_end_date BETWEEN :first AND :last OR scheduled_start_date BETWEEN :first AND :last "\
-      "OR old_start_date BETWEEN :first and :last", {first: range.first, last: range.last})
+  def self.project_step_date_filter(range, scope = ProjectStep, project_id)
+    scoped_steps = scope.where("actual_end_date BETWEEN :first AND :last OR "\
```
emitche: In terms of syntax, this feels like it could be tweaked. Would something like this work?

```
scoped_steps = scope.
  where("actual_end_date BETWEEN :first AND :last "\
  "OR scheduled_start_date BETWEEN :first AND :last "\
  "OR old_start_date BETWEEN :first and :last",
  {first: range.first, last: range.last})
```

PR #339, opened 2017-06-30
6537 reconfigure project filters

arisah-ijeoma:
6537


app/controllers/admin/dashboard_controller.rb
```diff
@@ -22,15 +22,9 @@ def prep_calendar
...
-    @recent_projects = @person.active_agent_projects.order(created_at: :desc, updated_at: :desc)
-
-    @recent_projects_grid = initialize_grid(
-      @recent_projects,
-      include: [:primary_agent, :secondary_agent],
-      per_page: 15,
-      name: "recent_projects",
-      enable_export_to_csv: false
-    )
+    @recent_projects = @person.active_agent_projects
```
emitche: I really like this refactor approach.

app/controllers/admin/dashboard_controller.rb
```diff
@@ -41,18 +35,38 @@ def prep_projects_grids_for_division_users
...
+
+  private
+
+  def initialize_wice_grid(projects, person, page)
+    initialize_grid(
+        projects,
+        include: [:primary_agent, :secondary_agent],
+        per_page: page,
+        order: 'projects.order_by_agent',
+        custom_order: {
```
emitche: I really like this approach.

app/controllers/admin/dashboard_controller.rb
```diff
@@ -41,18 +35,38 @@ def prep_projects_grids_for_division_users
...
+            'projects.order_by_agent' => "case when projects.primary_agent_id = #{person.id} and projects.status_value = 'active' and projects.type = 'Loan' then 8
```
emitche: The logic is good here. How can this be condensed visually, so it doesn't take up so many characters per line. Can the logic be moved into a variable or method? Can we shorten `projects.primary_agent_id` and `projects.secondary_agent_id` in any way?

Also, we try not to use `=>` syntax and instead use JSON style. From the style guide/dev procedures: "Use newer JSON style hash syntax for new and refactored code, even if existing code uses hash rocket style."

app/models/project.rb
```diff
@@ -48,8 +48,7 @@ class Project < ActiveRecord::Base
 
   # Status values can be found at Loan.status_option_set.options and
   # BasicProject.status_option_set.options
-  ACTIVE_STATUSES = %w(active changed possible prospective refinanced relationship
-    relationship_active)
+  STATUSES = %w(active changed possible prospective)
```
emitche: Maybe `OPEN_STATUSES`. There are other statuses not included here like "complete", "refinanced", and "relationship". I'd choose a name that makes it clear that this is a subset of all the existing status options.

app/views/admin/dashboard/_recent_projects_grid_definition.html.slim
```diff
@@ -1,29 +1,43 @@
 = define_grid(grid, hide_csv_button: true,
     html: {class: 'rowlink table-hover', data: {link: @csv_mode ? nil : 'row'}}) do |g|
 
+    - g.row_attributes do |project|
```
emitche: :) This is really cool. I didn't know about `row_attributes`.

app/views/admin/dashboard/_recent_projects_grid_definition.html.slim
```diff
@@ -1,29 +1,43 @@
...
 
-    - g.column name: t('common.name'), attribute: 'name', filter: false,
-      ordering: false do |project|
-      - project.display_name
-
-    - g.column name: t("activerecord.attributes.project.signing_date"),
-      attribute: 'signing_date', filter: false, ordering: false do |project|
-      - ldate(project.signing_date)
+    - g.column name: t('common.role'), attribute: 'primary_agent_id', filter: false,
+               ordering: false do |project|
```
emitche: Unindent this a bit. Too much indent on this line.

app/views/admin/dashboard/_recent_projects_grid_definition.html.slim
```diff
@@ -1,29 +1,43 @@
...
+      - if project.primary_agent_id == person.id
+        - '1st'
```
emitche: These values need to be translated and come from one of the en.yml files.

app/views/admin/dashboard/_recent_projects_grid_definition.html.slim
```diff
@@ -1,29 +1,43 @@
...
+      - else
+        - '2nd'
 
-    - g.column name: t('activerecord.attributes.project.progress'), attribute: 'status_value',
-      filter: false, ordering: false do |project|
+    - g.column name: t('activerecord.attributes.project.status'), attribute: 'status_value',
+               filter: false, ordering: false do |project|
```
emitche: Check indentation here and in rest of file.

app/views/admin/dashboard/_recent_projects_grid_definition.html.slim
```diff
@@ -1,29 +1,43 @@
...
       - content_tag(:span, project.status,
-        class: "project-status #{project.status_value == 'completed' ? 'completed' : 'active'}")
+                    class: "project-status #{project.status_value == 'completed' ? 'completed' : 'active'}")
```
emitche: No indent needed on this line. Line it up with the c in content. The - serves as the outer most item.

config/locales/es/es.yml
```diff
@@ -116,6 +116,7 @@ es:
         secondary_agent: "Agente Secundario"
         signing_date: "Fecha de Inicio"
         status: "Estado"
+        health_check: "Chequeo de Salud"
```
emitche: You don't need to add Spanish translations. Translations are managed in Transifex. Only adjust English translations, otherwise your changes in the Spanish files may be overridden.

emitche:
After you make the tweaks, please assign to me, remove the current label, and add the `complete` label. Thanks a bunch!

emitche:
The failed test is related to transactions and likely unrelated to this work.

app/controllers/admin/dashboard_controller.rb
arisah-ijeoma: the logic can be moved to the model but I can't shorten the `pai` and `sai`. Well not without typing them out first.

The `=>` here is the syntax being supported by the gem itself and also since I have a dot syntax `projects.order_by_agent` and with the who sql-ish-ness, having the new JSON style won't work here

config/locales/es/es.yml
arisah-ijeoma: ooh....okay

app/models/project.rb
```diff
@@ -131,4 +130,32 @@ def project_events(order_by = "Completed IS NULL OR Completed = '0000-00-00', Co
...
+
+          when projects.primary_agent_id = #{person.id} and projects.status_value = 'active'
+          and projects.type = 'BasicProject' then 7
+
+          when projects.primary_agent_id = #{person.id} and projects.status_value = 'prospective'
+          and projects.type = 'Loan' then 6
+
+          when projects.primary_agent_id = #{person.id} and projects.status_value = 'prospective'
+          and projects.type = 'BasicProject' then 5
+
```
emitche: Next in the list would be be for primary agent id and anything with one of the OPEN_STATUES not 'prospective' and 'active'. Right now all of the remaining statues show up at the bottom of the list, even if the user is the primary agent.

The next two items needed are:
All remaining loans where the user is the primary agent
All remaining basic projects where the user is the primary agent

app/models/project.rb
```diff
@@ -131,4 +130,32 @@ def project_events(order_by = "Completed IS NULL OR Completed = '0000-00-00', Co
...
+          when projects.secondary_agent_id = #{person.id} and projects.status_value = 'active'
+          and projects.type = 'BasicProject' then 3
+
+          when projects.secondary_agent_id = #{person.id} and projects.status_value = 'prospective'
+          and projects.type = 'Loan' then 2
+
+          when projects.secondary_agent_id = #{person.id} and projects.status_value = 'prospective'
+          and projects.type = 'BasicProject' then 1
+
```
emitche: Next in the list is projects where the user is the secondary agent and anything with one of the OPEN_STATUES not 'prospective' and 'active'. This is sorted first by Loan and then by Basic Project.

The next two items are:
All remaining loans in the list
All remaining basic projects in the list

The last filter can, of course, be the else below.

emitche:
This looks really good. One more tweak is needed.

app/controllers/admin/dashboard_controller.rb
hooverlunch: Agreed!

app/models/project.rb
```diff
@@ -131,4 +130,40 @@ def project_events(order_by = "Completed IS NULL OR Completed = '0000-00-00', Co
...
+          when projects.secondary_agent_id = #{person.id} and projects.status_value = 'prospective'
+          and projects.type = 'BasicProject' then 3
+
+          when projects.secondary_agent_id = #{person.id} and projects.type = 'Loan' then 2
+
+          when projects.secondary_agent_id = #{person.id} and projects.type = 'BasicProject' then 1
+
+          else 0 end"
+  end
```
hooverlunch: I think using CASE is generally an OK approach here, so well done! However, I think you can split the three attribs (agent status, type, status) into separate clauses to avoid the combinatorial explosion here.

I believe the logic would be:

* First point person = 1, Second point person = 2, else 3
* Active = 1, Prospective 2, else 3
* Loan = 1, Project = 2

in that order.

Note also that I've switched the order to ASC which I believe would be the default, so you can remove that line from your format grid call.

Happy to pair on this if you like.

hooverlunch: Oh also can you please make any SQL keywords (CASE, WHEN, ELSE, AND, END, THEN, etc.) uppercase? It makes things easier to read in my experience.

config/locales/en/en.yml
```diff
@@ -341,6 +342,7 @@ en:
     division: "Division"
     divisions: "Divisions"
     expand_all: "Expand All"
+    first: "1st"
```
hooverlunch: Nice job putting these in the i18n file!

app/views/admin/dashboard/dashboard.html.slim
```diff
@@ -2,7 +2,7 @@
 
 .block
   section.section.recent-projects.projects-grid
-    = render "recent_projects_grid_definition", grid: @recent_projects_grid
+    = render partial: "recent_projects_grid_definition", locals: { person: @person, grid: @recent_projects_grid }
```
hooverlunch: Can we please do spaces inside {} for procs, but not hashes? I find it's a good way of indicating one vs the other and it's the convention on this project I think.

app/models/project.rb
```diff
@@ -131,4 +130,17 @@ def project_events(order_by = "Completed IS NULL OR Completed = '0000-00-00', Co
...
+    clauses = []
+
+    clauses << "CASE WHEN projects.primary_agent_id = #{person_id} THEN 1"\
+                    "WHEN projects.secondary_agent_id = #{person_id} THEN 2 END"
+
+    clauses << "CASE projects.status_value WHEN 'active' THEN 1 WHEN 'prospective' THEN 2 ELSE 3 END"
+
+    clauses << "CASE projects.type WHEN 'Loan' THEN 1 WHEN 'BasicProject' THEN 2 END"
+
+    clauses.join(', ')
```
hooverlunch: Much better!

emitche:
Thanks a bunch, @hooverlunch, for reviewing.

I'm happy this can go into 1.17. Thanks so much, @arisah-ijeoma.

app/views/admin/dashboard/dashboard.html.slim
```diff
@@ -37,7 +38,8 @@
         .person
           h3 = person.name
           .recent-projects
-            = render "recent_projects_grid_definition", grid: @people_grids[person]
+            = render partial: "recent_projects_grid_definition",
+                     locals: {person: person, grid: @people_grids[person]}
```
emitche: Two spaces here instead of this larger tab. And above on line 6 too.

arisah-ijeoma: oooh...that's the way RubyMine formatted the string when I broke it
I can break it another way tho

emitche:
Thank you!

emitche:
The code looks good.

PR #343, opened 2017-07-06
6666 balance columns

arisah-ijeoma:
WIP and a failing test

I want you to take a look at the changes so far @meltheadorable

### DIFF
--- a/app/models/accounting/transaction.rb
+++ b/app/models/accounting/transaction.rb
@@ -52,6 +52,14 @@ class Accounting::Transaction < ActiveRecord::Base
     transaction.save!(validate: false)
   end
 
+  def self.by_date
+    order(:txn_date, :created_at)
+  end
+
+  def previous
+    self.class.where('txn_date <= ?', txn_date).by_date.last
+  end
+
   def quickbooks_data
     read_attribute(:quickbooks_data).try(:with_indifferent_access)
   end
--- a/app/models/accounting/transaction.rb
+++ b/app/models/accounting/transaction.rb
@@ -4,11 +4,15 @@
 #
 #  accounting_account_id :integer
 #  amount                :decimal(, )
+#  change_in_interest    :decimal(, )      default(0.0)
+#  change_in_principal   :decimal(, )      default(0.0)
 #  created_at            :datetime         not null
 #  currency_id           :integer
 #  description           :string
 #  id                    :integer          not null, primary key
+#  interest_balance      :decimal(, )      default(0.0)
 #  loan_transaction_type :string
+#  principal_balance     :decimal(, )      default(0.0)
 #  private_note          :string
 #  project_id            :integer
 #  qb_id                 :string           not null
@@ -43,6 +47,7 @@ class Accounting::Transaction < ActiveRecord::Base
   belongs_to :currency
 
   before_save :update_fields_from_quickbooks_data
+  after_save :calculate_loan_amounts
 
   validates :loan_transaction_type, :txn_date, :amount, :accounting_account_id, presence: true
 
@@ -53,11 +58,11 @@ class Accounting::Transaction < ActiveRecord::Base
   end
 
   def self.by_date
-    order(:txn_date, :created_at)
+    order(txn_date: :asc)
   end
 
-  def previous
-    self.class.where('txn_date <= ?', txn_date).by_date.last
+  def self.previous(loan_id)
+    where('txn_date <= ? AND project_id = ?', Date.today, loan_id).by_date.last(2).first
   end
 
   def quickbooks_data
@@ -71,6 +76,10 @@ class Accounting::Transaction < ActiveRecord::Base
   def associate_with_qb_obj(qb_obj)
     self.qb_id = qb_obj.id
     self.qb_transaction_type = qb_obj.class.name.demodulize
+  end
+
+  def total_balance
+    interest_balance + principal_balance
   end
 
   private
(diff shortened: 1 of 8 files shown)


app/models/accounting/transaction.rb
```diff
@@ -52,6 +57,14 @@ def self.find_or_create_from_qb_object(transaction_type:, qb_object:)
     transaction.save!(validate: false)
   end
 
+  def self.by_date
+    order(txn_date: :asc)
+  end
+
+  def self.previous(loan_id)
```
meltheadorable: `previous` needs to be an instance method, as it should be pointing to the instance before your current instance, and not just to the second most recent instance overall.

app/models/accounting/transaction.rb
```diff
@@ -95,4 +112,27 @@ def lookup_currency_id
       project.currency_id
     end
   end
+
+  def calculate_loan_amounts
+    loan_id = self.project_id
+
+    if self.loan_transaction_type == 'disbursement'
+      self.update_column(:change_in_principal, amount)
```
meltheadorable: I think you should just be assigning these values and then updating the column once at the end (preferably by just calling `save!` rather than `update_column`, right now this would do the update as a bunch of different database queries.

app/models/accounting/transaction.rb
```diff
@@ -95,4 +112,27 @@ def lookup_currency_id
...
+    end
+
+    if amount <= interest_balance
+      self.change_in_interest =- amount
```
meltheadorable: This should be `change_in_interest = -amount`, the `self` is redundant, and it's important that it be clear that we're assigning "negative amount" -- right now it's easy for somebody reading over the code to confuse `=-` with the `-=` operator which would do `change_in_interest = change_in_interest - amount` which is not what we want.

app/models/accounting/transaction.rb
```diff
@@ -95,4 +112,27 @@ def lookup_currency_id
...
+      self.update_column(:change_in_interest, change_in_interest)
+    elsif amount > interest_balance
+      self.change_in_interest =- interest_balance
```
meltheadorable: Same as above, make sure it's `= -variable_name` rather than `=- variable_name`

app/models/accounting/transaction.rb
meltheadorable: Once it's an instance method, it also shouldn't need to take an argument, since you will have the `project_id` of self.

app/views/admin/accounting/transactions/_transactions_grid_definition.html.slim
```diff
@@ -10,3 +10,20 @@
   - g.column name: t('activerecord.attributes.accounting.transaction.amount'),
     filter: false do |transaction|
     - format_currency transaction.amount, transaction.currency
+  - g.column name: t('activerecord.attributes.accounting.transaction.interest_paid'),
+    filter: false do |transaction|
+    - if transaction.loan_transaction_type == 'repayment'
+      - format_currency transaction.change_in_interest, transaction.currency
```
meltheadorable: I think this will show the opposite value of what you want, because the `change_in_interest` is going to be negative for repayments. I would consider adding methods for `interest_paid` and `capital_returned` that return `-change_in_interest` and `-change_in_principal` respectively.

meltheadorable: These are potentially good candidates for decorators in the future -- might be worth suggesting when the project people get back.

app/models/accounting/transaction.rb
```diff
@@ -95,4 +112,27 @@ def lookup_currency_id
...
   end
+
+  def calculate_loan_amounts
+    loan_id = self.project_id
+
+    if self.loan_transaction_type == 'disbursement'
+      self.update_column(:change_in_principal, amount)
+    end
+
+    if amount <= interest_balance
```
meltheadorable: I think you need to wrap this entire if/else block in a check to see whether it's a repayment. Disbursements still need to update the balances (way down the end) so you can't just return after you're finished checking disbursements, but as it stands you'd execute all of these lines for both and they're specific to repayments.

app/models/accounting/transaction.rb
```diff
@@ -95,4 +112,27 @@ def lookup_currency_id
...
+    elsif amount > interest_balance
+      self.change_in_interest =- interest_balance
+      self.update_column(:change_in_interest, change_in_interest)
+      self.change_in_principal =- (interest_balance - amount)
+      self.update_column(:change_in_principal, change_in_principal)
+    end
+
+    if self.class.previous(loan_id)
+      self.interest_balance = self.class.previous(loan_id).interest_balance + change_in_interest
+      self.update_column(:interest_balance, interest_balance)
```
meltheadorable: Don't forget to update the `principal_balance` as well. (and recall comments from above about using `self` and removing all these excess `update_column` calls. You should just be able to assign all of these values and then `save!`

app/models/loan.rb
```diff
@@ -210,4 +210,12 @@ def healthy?
   def health_status_available?
     return !loan_health_check.nil?
   end
+
+  def principal_balance
+    transactions.by_date.last.principal_balance
```
meltheadorable: You may want to do a `.try(:principal_balance)` as sometimes `.last` will return `nil` -- Same for `interest_balance` below.

spec/factories/loans.rb
```diff
@@ -132,6 +132,12 @@
       end
     end
 
+    trait :with_transaction do
+      after(:create) do |loan|
+        create(:accounting_transaction, project_id: loan.id)
```
meltheadorable: it's probably useful to have several transactions -- I'd consider doing a `create_list` here, with 3 or 4 transactions just to make sure there's always several to work with of different types for each loan. If you do add more than one, you should rename to `:with_transactions`

emitche:
As this was a preliminary pull request, I am closing this. Please open a new pull request when further development is ready to be looked at on this issue.

PR #350, opened 2017-07-26
6447 questions page numbering

hooverlunch:


app/models/loan_question.rb
```diff
@@ -181,6 +183,35 @@ def build_complete_requirements
...
+      nil
+    elsif parent.root?
+      number.to_s
+    else
+      "#{parent.full_number}.#{number}"
+    end
+  end
+
+  def full_number_and_label
+    [full_number, label].compact.join(". ")
```
linesarefuzzy: Oooooh, clever!

app/models/loan_question.rb
```diff
@@ -181,6 +183,35 @@ def build_complete_requirements
...
+
+  def set_numbers
+    # If parent_id is null (which can happen just after creation)
+    # there is no point in doing this.
+    return if parent_id.nil?
+    self.class.connection.execute("UPDATE loan_questions SET number = num FROM (
+      SELECT id, ROW_NUMBER() OVER (ORDER BY POSITION) AS num
+      FROM loan_questions
+      WHERE parent_id = #{parent_id} AND status = 'active'
+    ) AS t WHERE loan_questions.id = t.id")
```
linesarefuzzy: Wow, that's some fancy SQL. I guess efficiency is important here because it needs to run on drag and drop. Thanks for figuring this out!

app/models/loan_question.rb
```diff
@@ -181,6 +183,35 @@ def build_complete_requirements
...
+    self.class.where.not(status: "active").update_all("number = NULL")
```
linesarefuzzy: Is this necessary? What if we just set number to null if we are changing the status to inactive?

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
```diff
@@ -47,12 +47,9 @@
...
             - header_tag = {tag: %i(h3 h4 h5)[question.depth] || :h5}
             *header_tag
-              / Group title
-              => index_str
-              = question.label
-
+              = question.full_number_and_label
           - else
-            .question-label = "#{index_str} #{question.label}"
+            .question-label = question.full_number_and_label
```
linesarefuzzy: Already better 😅

spec/models/loan_question_spec.rb
```diff
@@ -120,4 +118,132 @@
...
+      expect(f31.reload.full_number).to be_nil
+      expect(f32.reload.full_number).to eq "2.1"
+      expect(f33.reload.full_number).to eq "2.2"
+      expect(f331.reload.full_number).to eq "2.2.1"
+      expect(f4.reload.full_number).to eq "3"
+    end
+  end
+
+  # Helper method to shorten keys
+  def create_q(attribs)
```
linesarefuzzy: Nice

spec/models/loan_question_spec.rb
```diff
@@ -120,4 +118,132 @@
...
+    attribs[:loan_question_set] = attribs.delete(:set)
+    attribs[:internal_name] = attribs.delete(:name)
+    attribs[:data_type] = attribs.delete(:type)
+    create(:loan_question, attribs)
+  end
```
linesarefuzzy: Beautiful specs omg thank you

app/models/loan_question.rb
arisah-ijeoma: @linesarefuzzy, it's set this way so it can cover `inactive` and `retired` questions instead of just inactive ones

emitche:
The numbering in the questionnaire aligns with the numbering in manage questions now.

I'm wondering about the number formatting, however. `1.` looks fine. But `1.2.1.` looks a bit strange. This can always adjust this later. So for now, I don't think we need to worry about it. Thanks.

image

This looks good overall from a functionality and QA perspective.

emitche:
@linesarefuzzy: @arisah-ijeoma did everything she could do on this issue. Is there anything else you need or can I go ahead and mark this as complete?

PR #358, opened 2017-08-03
other languages can access system

arisah-ijeoma:


app/controllers/application_controller.rb
```diff
@@ -42,12 +42,14 @@ def default_serializer_options
...
-    I18n.locale = params[:locale] || locale_from_header || I18n.default_locale
+    return I18n.locale = params[:locale] if params[:locale]
+    return I18n.locale = locale_from_header if I18n.available_locales.include?(locale_from_header)
+    I18n.locale = I18n.default_locale
   end
 
   def locale_from_header
     if !Rails.env.test? && (lang_header = request.env['HTTP_ACCEPT_LANGUAGE'])
-      lang_header.scan(/^[a-z]{2}/).first
+      lang_header.scan(/^[a-z]{2}/).first.to_sym
```
hooverlunch: This works great, thanks!

Another way would be to return nil from `locale_from_header` if it's not in the available_locales and then leave `set_locale` alone. Up to you :)

PR #360, opened 2017-08-04
6875: Make question name have more room on page

emitche:


emitche:
I've paired on this and the work is approved. I just want to test one more time before merging.

PR #361, opened 2017-08-07
6905 add line item model with corresponding spec

arisah-ijeoma:
In the transaction model, I noticed there were two line items; one for credit and the other for debit. That being said, I'm not sure what the association is from transactions and accounts (i.e. has_many or has_one)
@hooverlunch

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170726202231) do
+ActiveRecord::Schema.define(version: 20170807122855) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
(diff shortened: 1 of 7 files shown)


linesarefuzzy:
@arisah-ijeoma `has_many` for both.

arisah-ijeoma:
Turned out it was a db issue causing the bug I had initially

emitche:
I've spoken to @hooverlunch  and he plans to review this by 1 PM. This will be the last issue I QA for the release.

app/models/loan_question.rb
```diff
@@ -9,7 +9,6 @@
 #  internal_name         :string
 #  loan_question_set_id  :integer
 #  migration_position    :integer
-#  number                :integer
```
hooverlunch: Not sure why this is getting removed here. Thoughts?

db/schema.rb
```diff
@@ -409,20 +425,20 @@
...
-    t.boolean  "notify_on_new_logs", default: true
+    t.boolean  "notify_on_new_logs",     default: true
     t.integer  "profile_id"
     t.datetime "remember_created_at"
     t.datetime "reset_password_sent_at"
     t.string   "reset_password_token"
-    t.integer  "sign_in_count", default: 0, null: false
-    t.datetime "updated_at", null: false
+    t.integer  "sign_in_count",          default: 0,    null: false
+    t.datetime "updated_at",                            null: false
```
hooverlunch: I'm not sure how all these spaces got added in. We are using the fix-db-schema-conflicts gem which should prevent this. Can you try running your db:schema:dump again? Definitely don't want all these schema changes in the PR in any case please. Just what you're actually adding.

db/schema.rb
```diff
@@ -436,6 +452,8 @@
 
   add_index "users_roles", ["user_id", "role_id"], name: "index_users_roles_on_user_id_and_role_id", unique: true, using: :btree
 
+  add_foreign_key "accounting_line_items", "accounting_accounts"
+  add_foreign_key "accounting_line_items", "accounting_transactions"
```
hooverlunch: These look great.

app/models/loan_question.rb
arisah-ijeoma: I'll run `db:schema:dump`. Perhaps, that's fix this otherwise of course there is another issue

PR #362, opened 2017-08-09
6783 Create a blank interest transaction when a disbursement is created.

X0nic:


app/controllers/admin/accounting/transactions_controller.rb
```diff
@@ -27,15 +27,22 @@ def create
...
       # so that on the next sync, a duplicate is not created.
       @transaction.associate_with_qb_obj(journal_entry)
-
       @transaction.save!
+
+      # Create blank interest transaction. The interest calculator will pick this up and
+      # calculate the value, and sync it to quickbooks.
+      interest_transaction = ::Accounting::Transaction.find_or_initialize_by(transaction_interest_params
+        .merge(qb_transaction_type: ::Accounting::Transaction::LOAN_INTEREST_TYPE))
+      interest_transaction.save!
```
hooverlunch: Use `find_or_create_by!` instead of `save!`?

app/controllers/admin/accounting/transactions_controller.rb
```diff
@@ -27,15 +27,22 @@ def create
...
+
       flash[:notice] = t("admin.loans.transactions.create_success")
       render nothing: true
     rescue => ex
       # We don't need to display the message twice if it's a validation error.
       # But we do want to display the error if the QB API blows up.
       if ex.is_a?(ActiveRecord::RecordInvalid)
-        # Do nothing
+        # Only raise error if we had a problem saving the interest transaction
+        raise ex if @transaction.valid?
```
hooverlunch: You can do this more directly by checking if `ex.record == interest_transaction`, right? Then you wouldn't need the comment. You may have to make interest_transaction an instance var to do that, or initialize it to nil outside the begin block.

app/controllers/admin/accounting/transactions_controller.rb
```diff
@@ -27,15 +27,22 @@ def create
       # It's important we store the ID and type of the QB journal entry we just created
       # so that on the next sync, a duplicate is not created.
       @transaction.associate_with_qb_obj(journal_entry)
-
       @transaction.save!
+
+      # Create blank interest transaction. The interest calculator will pick this up and
+      # calculate the value, and sync it to quickbooks.
+      interest_transaction = ::Accounting::Transaction.find_or_initialize_by(transaction_interest_params
```
hooverlunch: I think this would be clearer as like `transaction_params.except(:amount, ...)`. I was confused at first why there were two _params methods, suggesting two different forms or something...

app/models/accounting/transaction.rb
```diff
@@ -37,14 +37,16 @@
...
 
   belongs_to :account, inverse_of: :transactions, foreign_key: :accounting_account_id
   belongs_to :project, inverse_of: :transactions, foreign_key: :project_id
   belongs_to :currency
 
   before_save :update_fields_from_quickbooks_data
 
-  validates :loan_transaction_type, :txn_date, :amount, :accounting_account_id, presence: true
+  validates :loan_transaction_type, :txn_date, :accounting_account_id, presence: true
+  validates :amount, presence: true, unless: -> { qb_id.blank? && qb_transaction_type == LOAN_INTEREST_TYPE }
```
hooverlunch: Are you sure the `unless` block works? I thought you had to take the object as a Proc param and call the member methods on that. http://guides.rubyonrails.org/active_record_validations.html#using-a-proc-with-if-and-unless

Please test it and let me know one way or the other.

db/migrate/20170809054542_remove_qb_id_null_from_accounting_transaction.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveQbIdNullFromAccountingTransaction < ActiveRecord::Migration
+  def change
+    change_column :accounting_transactions, :qb_id, :string, null: true
```
hooverlunch: `change_column_null :accounting_transactions, :qb_id, true` is a bit simpler.

spec/models/accounting/transaction_spec.rb
```diff
@@ -76,4 +76,70 @@
       expect(subject.project_id).to eq loan.id
     end
   end
+
+  context 'when disbursement transaction' do
+    let(:transaction_type) { 'disbursement' }
+
+    context 'without qb_id' do
+      it 'can save without amount' do
```
hooverlunch: This should read 'requires an amount to save', right?

spec/models/accounting/transaction_spec.rb
```diff
@@ -76,4 +76,70 @@
...
+          txn_date: '2017-10-31',
+          private_note: 'a memo',
+          description: 'desc',
+          project_id: loan.id,
+          qb_transaction_type: transaction_type,
+          qb_id: 123
+        }
+        expect { create(:accounting_transaction, params) }.to raise_error(ActiveRecord::RecordInvalid)
+      end
+    end
```
hooverlunch: Can you please DRY up these specs. Most of them are identical attributes. Those can go in a let block at the top and you can just do `merge` to set the ones you're testing.

This will make it much clearer what you're actually changing from assertion to assertion.

In general if you could please make an effort to keep specs DRY. You're good about this with regular code but often your specs are much longer than they need to be.

Thanks!

app/models/accounting/transaction.rb
X0nic: I could change it if you prefer, but the specs I wrote are to test this validation specifically. So I know it does work.

X0nic:
That is correct @hooverlunch there is no InterestCalculator yet.

app/models/accounting/transaction.rb
hooverlunch: Oh ok, great. Fine to leave as is. Learn something new everyday!

linesarefuzzy:
> I notice that InterestCalculator is not getting called as was requested in the spec. Is that maybe because there is no InterestCalculator yet? Is that what "(if it already exists)" means? @linesarefuzzy?

Yes, exactly.

emitche:
This is what I see when testing:

image

emitche:
Failed spec:

```
Failures:

  1) transaction flow transactions for loan creates new transaction

     Failure/Error: select accounts.sample.name, from: 'Bank Account'

     Capybara::Ambiguous:

       Ambiguous match, found 2 elements matching option "eius"

     # ./spec/features/admin/accounting/transaction_flow_spec.rb:40:in `block (3 levels) in <top (required)>'

Finished in 4 minutes 39.6 seconds (files took 6.18 seconds to load)

759 examples, 1 failure, 28 pending

Failed examples:

rspec ./spec/features/admin/accounting/transaction_flow_spec.rb:35 # transaction flow transactions for loan creates new transaction
```

emitche:
Visually this looks fine.

I'm having problems running tests on my environment. So rspec needs to be run.

Was this part included, @X0nic? This is from the Redmine issue description.
```
Add an optional loan argument to Accounting::Quickbooks::Updater#update (that does nothing for now)
```

emitche:
@X0nic 

I've added multiple transactions and see that these empty transactions are not created consistently. I'm not sure what is the cause of this.

image

emitche:
@xionic

So result of QA:
- A test is failing, but doesn't fail on my local environment.
- I have a question about optional loan argument.
- There appears to be a bug.

linesarefuzzy:
Is the inconsistency related to the feature of not creating a second interest tx on the same day as an existing interest tx?

Regardless, this was not part of the spec, but please make the description of the interest txs "Interest Accrual for Loan #XXX".

emitche:
"Is the inconsistency related to the feature of not creating a second interest tx on the same day as an existing interest tx?"

That's possibly the reason. If so, then this is not a bug.

emitche:
I'm going to clear up the overall questions for this issue and what is needed to move forward.

emitche:
Hi @xionic. There are some more TODOS on this issue before it can be merged.

Please merge develop into this branch.

I had a question about a bug I am seeing. Please look into it. If you know what is happening here, please let me know. https://github.com/sassafrastech/madeline/pull/362#issuecomment-321877400

An additional tweak requested to this feature:
`please make the description of the interest txs "Interest Accrual for Loan #XXX".`
I'll add this to the Redmine description so it is clear that this is needed.

It appears one item from the Redmine description was not done:
````
Add an optional loan argument to Accounting::Quickbooks::Updater#update
(that does nothing for now)
````

Thank you!

**Comment by hooverlunch** [comment]
@X0nic pinging you because Liz's message used the wrong @ handle and she just left for the day :)  Please see above.

File: db/schema.rb [hunk 8]
```diff
@@ -73,7 +74,7 @@
   add_index "accounting_transactions", ["accounting_account_id"], name: "index_accounting_transactions_on_accounting_account_id", using: :btree
   add_index "accounting_transactions", ["currency_id"], name: "index_accounting_transactions_on_currency_id", using: :btree
   add_index "accounting_transactions", ["project_id"], name: "index_accounting_transactions_on_project_id", using: :btree
-  add_index "accounting_transactions", ["qb_id", "qb_transaction_type"], name: "acc_trans_qbid_qbtype_unq_idx", unique: true, using: :btree
+  add_index "accounting_transactions", ["qb_id", "qb_transaction_type"], name: "acc_trans_qbid_qbtype__unq_idx", unique: true, using: :btree
```
hooverlunch: More schema stuff leaking in here. See if you can remove please?

PR #364, opened 2017-08-18
6940 add division to questions

emitche:
- Remove division from question set model
- Add division to question model
- Ensure user can specify a division for each question
- Set root division to division for all existing questions
- Fix tests

### DIFF
--- a/app/views/admin/loan_questions/_form.html.slim
+++ b/app/views/admin/loan_questions/_form.html.slim
@@ -25,6 +25,8 @@
         td = f.input_field :has_embeddable_media, as: :boolean, boolean_style: :inline
         td = f.label :has_embeddable_media
 
+    = f.input "division", label: "Division"
+
     = render "loan_types", f: f
 
     = f.translatable_inputs do |l|
--- a/app/views/admin/loan_questions/_form.html.slim
+++ b/app/views/admin/loan_questions/_form.html.slim
@@ -18,14 +18,15 @@
     = f.input :status, label: t("loan_questions.active"), as: :boolean, checked_value: 'active',
       unchecked_value: 'inactive'
 
+    = f.input "division", label: "Division", collection: current_user.accessible_divisions,
+      selected: @loan_question.division.id
+
     = f.input :data_type do
       = f.input_field :data_type, collection: @data_types
 
       table.table.allow-attachments: tbody: tr
         td = f.input_field :has_embeddable_media, as: :boolean, boolean_style: :inline
         td = f.label :has_embeddable_media
-
-    = f.input "division", label: "Division"
 
     = render "loan_types", f: f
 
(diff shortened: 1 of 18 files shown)


app/views/admin/loan_questions/_form.html.slim
```diff
@@ -18,6 +18,9 @@
     = f.input :status, label: t("loan_questions.active"), as: :boolean, checked_value: 'active',
       unchecked_value: 'inactive'
 
+    = f.input :division_id, collection: current_user.accessible_divisions,
+      selected: @loan_question.division.id
```
hooverlunch: This can be `division_id` instead of `division.id` -- a little more efficient since a database call won't be needed.

db/migrate/20170816150506_change_division_id_in_loan_question.rb
```diff
@@ -0,0 +1,18 @@
...
+    change_column_null :loan_questions, :division_id, false
+  end
+
+  def down
+    # Change column from integer to string
+    change_column :loan_questions, :division_id, :string
+    LoanQuestion.all.each { |q| q.update_attribute(:division_id, 99) }
+
+    change_column_null :loan_questions, :division_id, true
+  end
```
hooverlunch: I'm a bit confused as to why this column was created as a string first. Was it a mistake? If so, that's fine, but it would be better to simply fix the original migration instead of adding a new one.

Or if there is a reason please let me know.

emitche: Yes, it was a mistake in the first migration. Is it best to just adjust all these migrations so that the string to integer conversion does not ever happen? Thanks.

app/views/admin/loan_questions/_form.html.slim
emitche: Ok. I can make that change.

db/migrate/20170816150506_change_division_id_in_loan_question.rb
hooverlunch: Thank you too! Since nobody else has run this migration yet, you should be able to fix these without breaking anyone's db.

emitche: Thanks. I'll fix these migrations.

emitche: I'm having trouble adjusting these migrations. I'll reach out for help.

db/migrate/20170816144924_change_division_on_questions.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeDivisionOnQuestions < ActiveRecord::Migration
+  def up
+    execute("UPDATE loan_questions SET division_id=99")
```
emitche: @linesarefuzzy Is 99 the correct id across environments or should a variable be used here?

linesarefuzzy:
QA done. Looks and works great!

linesarefuzzy:
@emitche Can you resolve conflicts and merge?

emitche:
Yes, I can take care of that.

emitche:
Thank you!

PR #366, opened 2017-08-21
6685 Add simple version of project duplicator

X0nic:


app/models/accounting/transaction.rb
```diff
@@ -20,7 +20,7 @@
 #
 # Indexes
 #
-#  acc_trans_qbid_qbtype_unq_idx                           (qb_id,qb_transaction_type) UNIQUE
```
hooverlunch: Not sure how these schema changes got in here. Can we take it out?

app/models/project.rb
```diff
@@ -64,6 +67,18 @@ class Project < ActiveRecord::Base
 
   alias_method :logs, :project_logs
 
+  # Configure how the class is duplicated
```
hooverlunch: Nice comment! Wouldn't have been clear what 'amoeba' means :)

app/models/project.rb
```diff
@@ -64,6 +67,18 @@ class Project < ActiveRecord::Base
...
+  amoeba do
+    enable
+    propagate
+
+    # The default name is computed, if it hasn't been set it will be blank.
+    # We need to manually copy over the name and set it here for it to work.
+    customize(lambda { |original, new|
+      new.name = "Copy of #{original.display_name}"
```
hooverlunch: Maybe choose a different var name here instead of `original` since that is also used as an association? Maybe just `orig`?

app/models/project_duplicator.rb
```diff
@@ -0,0 +1,15 @@
...
+    @project_to_copy = project_to_copy
+  end
+
+  def duplicate
+    copy = project_to_copy.amoeba_dup
+    copy.original = project_to_copy
+
+    copy.save!
+    copy
+  end
```
hooverlunch: Sweet!

db/migrate/20170821035759_add_original_to_project.rb
```diff
@@ -0,0 +1,8 @@
+class AddOriginalToProject < ActiveRecord::Migration
+  def change
+    add_column :projects, :original_id, :integer, polymorphic: true, foreign_key: { references: :projects}
+    # add_reference :projects, :original
+    # add_foreign_key :projects, :original, column: :original_id
+    # add_index :projects, [:id, :original_id]
```
hooverlunch: Why are these commented?

Here would be my thoughts:
* You are right to have a foreign key on original_id. 
* original_id should also have an index.
* What is the compound index for?
* What does `polymorphic: true` do for `add_column`? I've only ever seen it with `add_reference`? I don't think we need an `original_type` column, do we?

db/schema.rb
```diff
@@ -73,7 +74,7 @@
   add_index "accounting_transactions", ["accounting_account_id"], name: "index_accounting_transactions_on_accounting_account_id", using: :btree
   add_index "accounting_transactions", ["currency_id"], name: "index_accounting_transactions_on_currency_id", using: :btree
   add_index "accounting_transactions", ["project_id"], name: "index_accounting_transactions_on_project_id", using: :btree
-  add_index "accounting_transactions", ["qb_id", "qb_transaction_type"], name: "acc_trans_qbid_qbtype_unq_idx", unique: true, using: :btree
+  add_index "accounting_transactions", ["qb_id", "qb_transaction_type"], name: "acc_trans_qbid_qbtype__unq_idx", unique: true, using: :btree
```
hooverlunch: Hopefully we can nix this.

db/schema.rb
```diff
@@ -11,9 +11,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170810192441) do
+ActiveRecord::Schema.define(version: 20170821035759) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+  enable_extension "pg_stat_statements"
```
hooverlunch: What is this?

spec/features/admin/loan_flow_spec.rb
```diff
@@ -51,6 +51,15 @@
     end
   end
 
+  describe 'details' do
+    scenario 'can duplicate', js: true do
+      visit admin_loan_path(loan)
+
+      click_on('Duplicate')
+      expect(page).to have_content "Copy of #{loan.display_name}"
```
hooverlunch: Yay!

spec/models/project_duplicator_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+RSpec.describe ProjectDuplicator, type: :model do
+  subject(:duplicator) { described_class.new(loan) }
+  let(:loan) { create(:loan) }
+
+  it 'copies proper columns' do
+    new_loan = duplicator.duplicate
+
+    attribs_to_exclude = %w(id created_at updated_at name original_id)
+    expect(new_loan.attributes.except(*attribs_to_exclude))
+      .to eq loan.attributes.except(*attribs_to_exclude)
```
hooverlunch: Nicely written!

spec/models/project_duplicator_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  context 'with default name' do
+    let(:loan) { create(:loan, name: '') }
+
+    it 'name column is prepended with "Copy of"' do
+      new_loan = duplicator.duplicate
+
+      expect(new_loan.name) .to eq "Copy of #{loan.display_name}"
```
hooverlunch: To me this would be clearer if it were hardcoded or maybe `loan.default_name` instead of `display_name`.

db/migrate/20170821035759_add_original_to_project.rb
X0nic: I meant to remove the polymorphic, thanks for catching.
The foreign key is an index, so we don't need to add one. I checked psql, and it is there.

emitche:
@X0nic I ran into an error when working on QA for this issue. Please see https://redmine.sassafras.coop/issues/6685#note-11.

emitche:
QA review: This looks good. One failing spec, but otherwise looking great.

emitche:
Looks good.

PR #367, opened 2017-08-21
6942 manage questions permissions

emitche:
- Show questions as editable and not editable based on permissions


app/assets/javascripts/backbone/views/loan_questions_view.coffee
```diff
@@ -9,16 +9,22 @@ class MS.Views.LoanQuestionsView extends Backbone.View
       data: @tree.data('data')
       dragAndDrop: true
       selectable: false
+      onCanMove: (node) =>
+        if node.can_edit == true
+          return true
+        else
+          return false
```
hooverlunch: I think this could be written as `onCanMove: (node) => node.can_edit`. If it's not clear why, I'm happy to quickly discuss it.

app/assets/javascripts/backbone/views/loan_questions_view.coffee
```diff
@@ -149,3 +155,20 @@ class MS.Views.LoanQuestionsView extends Backbone.View
         "<div class='loan-type-required'>#{loan_type}</div>"
       .join(' ') +
       "</div>"
+
+  permittedActionsHTML: (node) ->
+    if node.can_edit == true
```
hooverlunch: I don't think you need `== true` here. `== true` is very rarely needed. Only if you are checking that something is exactly `true` versus some other "truthy" value. And in that case, you'd want `===` I think. But as I say, that is rare.

app/assets/javascripts/backbone/views/loan_questions_view.coffee
```diff
@@ -149,3 +155,20 @@ class MS.Views.LoanQuestionsView extends Backbone.View
...
+      $('.links-block').html()
+    else
+      $('.actions-disabled-block').html()
+
+  prepTooltips: ->
+    @$('.ms-tooltip').each (index, tip) ->
+      message = I18n.t('loan_questions.not_editable')
+
+      $(tip).addClass('ms-popover').popover
```
hooverlunch: I think if you change the inner `->` to `=>` above, you should be able to use `@$` here, which is preferable to `$`.

In order for `@` (which is the same as `this`) to refer to the right thing, you need the `=>`. Also happy to discuss this further if you like.

app/assets/javascripts/backbone/views/loan_questions_view.coffee
```diff
@@ -149,3 +155,20 @@ class MS.Views.LoanQuestionsView extends Backbone.View
...
+        content: message
+        html: true
+        placement: 'left'
+        toggle: 'popover'
+        trigger: 'manual'
```
hooverlunch: Nice method!

app/controllers/admin/loan_questions_controller.rb
```diff
@@ -7,7 +7,8 @@ def index
     # Hide retired questions for now
     sets = LoanQuestionSet.where(internal_name: %w(loan_criteria loan_post_analysis)).to_a
     @json = ActiveModel::Serializer::CollectionSerializer.new(
-      sets.map { |s| s.root_group_preloaded.children_applicable_to(nil) }.flatten
+      sets.map { |s| s.root_group_preloaded.children_applicable_to(nil) }.flatten,
+      {user: current_user}
```
hooverlunch: Nice use of serializer options here! One thing: you don't need the curly braces since the hash is the last argument to the `new` method.

app/serializers/loan_question_serializer.rb
```diff
@@ -60,4 +62,12 @@ def required_loan_types
   def status
     object.status.presence || "active"
   end
+
+  def can_edit
+    if @user
+      Pundit.policy!(@user, object).edit?
```
hooverlunch: Very cool.

emitche:
@linesarefuzzy If you have capacity, you can QA this. Thanks!

emitche:
After merging 6940 into develop and testing this feature, this branch is not aware that individual questions may be editable. It still has groups listed as editable and individual questions not editable.

emitche:
I've looked into this issue further. The problem is that the policy is not working properly. Any group or question that is a child of one of the main groups is unable to be edited, even though they should be editable.

app/serializers/loan_question_serializer.rb
emitche: This is not working properly. Something needs to be changed here.

emitche:
This issue was already code reviewed. In QA I found a bug that has since been resolved. @linesarefuzzy, can you do a QA review on this issue?

emitche:
@linesarefuzzy I will take care of QA on this issue.

linesarefuzzy:
The basics look great. I made a small change to the popover message. 

But I'm finding a bug: 

1. As a user belonging to a child division, I make any change I am allowed to make: adding a question, editing my own question, dragging my own question, or deleting my own question
2. All the questions, including mine, now show as locked.
3. All the popovers seem to be disabled.
4. When I reload the page, all is well.

emitche:
I noticed this bug too and thought I fixed it. I'll move this out of the current sprint, as fixing it may not be a quick thing.

app/serializers/loan_question_serializer.rb
emitche: This should be working now. The policy was not the problem.

emitche:
@linesarefuzzy I believe I fixed the bug where all items where showing as not-editable when a question was edited.

This is ready for QA again. Thank you!

linesarefuzzy:
Thanks! It was still doing that after creating a question. I fixed it.

PR #370, opened 2017-08-31
6884 update transaction creator

arisah-ijeoma:


app/models/accounting/transaction.rb
```diff
@@ -22,7 +22,7 @@
 #
 # Indexes
 #
-#  acc_trans_qbid_qbtype__unq_idx                          (qb_id,qb_transaction_type) UNIQUE
+#  acc_trans_qbid_qbtype_unq_idx                           (qb_id,qb_transaction_type) UNIQUE
```
emitche: This should be kept as `__` instead of `_`.

spec/models/accounting/quickbooks/transaction_creator_spec.rb
```diff
@@ -51,23 +83,20 @@
...
     expect(generic_service).to receive(:create) do |arg|
-      expect(arg.line_items.count).to eq 2
+      expect(arg.line_items.count).to eq 3
       expect(arg.private_note).to eq memo
       expect(arg.txn_date).to be_nil
 
       list = arg.line_items
-      expect(list.map(&:amount).uniq).to eq [amount]
-      expect(list.map(&:description).uniq).to eq [description]
+      expect(list.map(&:amount)).to eq [line_item_1.amount, line_item_2.amount, line_item_3.amount]
```
hooverlunch: If you define `line_items` (as an array) instead of separate vars, you can do `line_items.map(&:amount)` here. Similar below.

app/models/accounting/transaction.rb
emitche: I would check in with @X0nic or @linesarefuzzy on why we are using `acc_trans_qbid_qbtype__unq_idx` with double underscores. But it's not an error that there are two underscores.

db/migrate/20170831115640_remove_line_item_transaction_constraint.rb
```diff
@@ -0,0 +1,8 @@
+class RemoveLineItemTransactionConstraint < ActiveRecord::Migration
+  def change
+    # transactions being passed to the transaction creator are still new instances
+    # without ids so the specs keep failing when trying to associate a line item
```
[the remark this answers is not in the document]
arisah-ijeoma: Going by the current transaction flow, there didn't seem to be an underlying issue. I am however interested on how we intend to go about the logic here.

emitche:
Wice grid error is going to be resolved in a different pull request.

emitche:
I am not able to save a new transaction with these changes:

image

emitche:
Result of functional and QA review is that this needs changes in order to work correctly.

arisah-ijeoma:
@linesarefuzzy and I hit that yesterday and took a deeper look

It seems to tie to https://redmine.sassafras.coop/issues/6878 where the line items will get created

emitche:
This works as it should, but need to be merged after the 1.21 release.

PR #372, opened 2017-09-01
6764 txn show page

emitche:
- Refactor transaction modal so add, show, and edit will use same modal
- Add show transaction modal


app/views/admin/accounting/transactions/show.html.slim
```diff
@@ -0,0 +1,18 @@
+div
```
emitche: We should reuse the form for new transactions. The transaction form needs to be split out from the modal content.

I will make that change.

app/views/admin/accounting/transactions/_transactions_grid_definition.html.slim
```diff
@@ -1,6 +1,10 @@
 = define_grid(@transactions_grid, hide_csv_button: true,
     html: {class: 'rowlink table-hover', data: {link: 'row'}}) do |g|
 
+  - if action_name == 'show'
+    - g.row_attributes do |transaction|
+      - {onclick: "document.location = '#{admin_loan_transaction_path(loan_id: @loan.try(:id), id: transaction)}';"}
```
emitche: It would be better to use a `link_to` here.

For example, this is from the loans grid.
```
= define_grid(@loans_grid, show_filters: :show, hide_csv_button: true,
    html: {class: 'rowlink table-hover', data: {link: 'row'}}) do |g|

    - g.blank_slate { @loans_grid.no_records_at_all = true; "" }

    - g.column name: t('common.id'), attribute: 'id', filter: true, in_csv: false, detach_with_id: :id_filter do |loan|
      - link_to(loan.id, admin_loan_path(loan.id))
```

app/views/admin/accounting/transactions/_transactions_grid_definition.html.slim
```diff
@@ -1,6 +1,10 @@
 = define_grid(@transactions_grid, hide_csv_button: true,
     html: {class: 'rowlink table-hover', data: {link: 'row'}}) do |g|
 
+  - if action_name == 'show'
```
emitche: Why is `action_name` used here? Are there cases where we would not be showing the link?

I know we sometimes export tables, but a different syntax would be used here.

arisah-ijeoma: This is because of the transactions index link on the sidebar
It breaks since some of those transactions don't have loans

app/views/admin/accounting/transactions/_transactions_grid_definition.html.slim
arisah-ijeoma: gotcha

Will update this

arisah-ijeoma: yea, I remember why I couldn't use the link_to helper

the row_attributes expects a hash to be returned so the helper does not work with it

emitche: Ok. Don't worry about it then. Thanks.

app/views/admin/accounting/transactions/_transactions_grid_definition.html.slim
emitche: Oh. I see what you mean.

emitche:
@arisah-ijeoma Good work on the code.

I'm going to make a tweak to use the form as a partial here and add some view only fields to the existing form. That's all frontend stuff though. :)

emitche:
Using a page seems to be very complicated. So instead, we are going to switch the transaction to open up in a modal, I think.

arisah-ijeoma:
@emitche, are we updating this story to use a modal?

emitche:
@arisah-ijeoma Yes. We are updating the story to use a modal instead. I was going to make that change, but you can also do it, if you are interested. Just let me know. Thanks a bunch.

emitche:
@arisah-ijeoma I've reassigned this issue to back to you. I'll add more info in Redmine.

app/assets/javascripts/backbone/views/show_edit_view.coffee
```diff
@@ -3,10 +3,6 @@ class MS.Views.ShowEditView extends Backbone.View
```
emitche: We currently do not need edit links inside the modal, so changing this file is not needed.

This file is used throughout the system. Keep this file as is. This change here breaks functionality for modals throughout the system.

This is the pattern that should be used for the transaction modal, but not right at this moment. We are only implementing show transaction (not edit and delete transaction) at this time. A good example of modal usage is Add Step and Edit Step in the Timeline. Or inside the loan calendar, clicking on a step shows the modal in show view for a step. You can click edit and see the edit view. That's what type of functionality we are aiming for with this feature. But we only need the show view (and the new transaction view as well). Edit transaction is a separate feature.

A good example of show and edit views being used is Loan Details. When you first get to the page you see the show view. Then edit is shown.

app/views/admin/accounting/transactions/_show_modal_content.slim
```diff
@@ -0,0 +1,24 @@
+div.modal-header
```
emitche: The modal content would use: `app/views/admin/accounting/transactions/_form.slim`. If you need help with this part, please let me know.

The best example I can see of a flow for this would be the project logs modal. The way the code is set up there would be what we would use similarly for this feature.

A show modal file would not be needed, if we can follow the same pattern.

app/views/admin/accounting/transactions/_show_modal_content.slim
```diff
@@ -0,0 +1,24 @@
...
+  .actions.txn-actions
```
emitche: No transaction actions should be in the modal at this time. Edit and delete are a separate feature.

app/views/admin/accounting/transactions/show.js.erb
```diff
@@ -0,0 +1,2 @@
+$('#transaction_show').html("<%= j render 'show_modal' %>");
```
emitche: This is not standard. We usually use a Backbone view. We would build upon `transactions_view.coffee` and `transactions_modal_view.coffee`.

Examples of a JavaScript, controller, and template flow (where these work together) are project logs and project groups.

app/views/admin/loans/_transactions.html.slim
```diff
@@ -6,6 +6,7 @@ section.transactions
   = render "admin/accounting/transactions/transactions_grid_definition"
   = render_index_grid_with_redirect_check(@transactions_grid)
   = render "admin/accounting/transactions/modal"
+  #transaction_show
```
emitche: This is non-standard. The modal in line 8 can be reused for show transaction.

spec/features/admin/accounting/transaction_flow_spec.rb
```diff
@@ -47,7 +47,21 @@
...
+
+  describe 'show', js: true do
+    let!(:loan) { create(:loan) }
+    let!(:txn) { create(:accounting_transaction, project_id: loan.id, description: 'I love icecream') }
+
+    scenario 'can show transactions' do
+      visit admin_loan_tab_path(loan, tab: 'transactions')
+      click_on txn.txn_date.strftime('%B %d, %Y')
+      sleep 5
+      expect(page).to have_content('Edit')
```
emitche: We won't need this line.

spec/features/admin/accounting/transaction_flow_spec.rb
```diff
@@ -47,7 +47,21 @@
       fill_in 'Memo', with: 'Chunky monkey'
```
emitche: Thanks for adding a spec.

spec/features/admin/accounting/transaction_flow_spec.rb
```diff
@@ -47,7 +47,21 @@
...
-      expect(page).to have_content("Foo bar")
+      expect(page).to have_content('Foo bar')
     end
   end
+
+  describe 'show', js: true do
+    let!(:loan) { create(:loan) }
+    let!(:txn) { create(:accounting_transaction, project_id: loan.id, description: 'I love icecream') }
+
+    scenario 'can show transactions' do
```
emitche: This will need to be adjusted once we make adjustments to the show modal.

emitche:
@arisah-ijeoma This is a lot of front-end work. Let me know if you want to continue to work on this and learn about front-end patterns in Madeline. If you'd rather I make these changes, that is fine as well. Thanks for your work on this.

app/views/admin/accounting/transactions/_form.slim
```diff
@@ -0,0 +1,45 @@
...
+
+  = f.input :txn_date
+    .view-element = ldate(@transaction.txn_date)
+    = f.input_field :txn_date, as: :date_picker
+
+  = f.input :account
+    .view-element
+      - if @transaction.account
+        = @transaction.account.name
+    = f.input_field :account, collection: @accounts
```
emitche: The Rails view expects `account` instead of `account_id`, but the parameters for create transaction expect `account_id`. Usually `f.association` would work, but because of a show/edit configuration here, `f.association` has some styling issues. What do you recommend?

hooverlunch: Let's discuss when we pair.

app/views/admin/accounting/transactions/_modal_content.slim
```diff
@@ -1,20 +1,19 @@
...
-    = f.input :amount
-    = f.input :description
-    = f.input :private_note, as: :text
+  = render "admin/accounting/transactions/form"
 
 div.modal-footer
-  button.btn.btn-default.cancel type="button" data-dismiss="modal" = t(:cancel)
-  button.btn.btn-primary = t("common.add")
+  .view-element
+    button.btn.btn-default.cancel type="button" data-dismiss="modal" = t(:close)
```
emitche: This content needs to be replaced with `a` tags instead. Use logs modal as an example.

emitche:
I need help fixing tests.

emitche:
The failing tests are just on this branch and not on develop.

app/models/accounting/transaction.rb
```diff
@@ -56,7 +56,9 @@ class Accounting::Transaction < ActiveRecord::Base
   validates :loan_transaction_type_value, :txn_date, :accounting_account_id, presence: true
   validates :amount, presence: true, unless: :uninitialized_interest?
 
-  delegate :division, to: :project
+  # since we have a new route that has to do with loans
+  # transactions index breaks since not all transactions have loans
```
hooverlunch: It's an open question to me if it makes sense to have transactions without loans. Do you have any insight on this? If possible, I'd rather add a not-null constraint to transactions.project_id.

app/models/accounting/transaction.rb
```diff
@@ -89,11 +91,13 @@ def associate_with_qb_obj(qb_obj)
...
   def change_in_principal
-    @change_in_principal ||= sum_for_account(division.principal_account_id)
+    # not all transactions have loans
+    @change_in_principal ||= sum_for_account(division.try(:principal_account_id))
   end
 
   def change_in_interest
-    @change_in_interest ||= sum_for_account(division.interest_receivable_account_id)
+    # not all transactions have loans
+    @change_in_interest ||= sum_for_account(division.try(:interest_receivable_account_id))
```
hooverlunch: This may clear the error but passing nil to `sum_for_account` seems odd.

app/views/admin/accounting/transactions/_modal_content.slim
```diff
@@ -1,20 +1,19 @@
...
 
 div.modal-footer
-  button.btn.btn-default.cancel type="button" data-dismiss="modal" = t(:cancel)
-  button.btn.btn-primary = t("common.add")
+  .view-element
+    a.btn.btn-default.cancel data-dismiss="modal" = t(:close)
+  .form-element
+    a.btn.btn-default.cancel data-dismiss="modal" = t(:cancel)
+    a.btn.btn-primary data-action="submit"
+      = @transaction.new_record? ? t("common.add") : t("admin.loans.transactions.update")
```
hooverlunch: Templates look solid.

spec/features/admin/accounting/transaction_flow_spec.rb
```diff
@@ -43,11 +43,24 @@
...
+      visit admin_loan_tab_path(loan, tab: 'transactions')
+      click_on txn.txn_date.strftime('%B %-d, %Y')
+      sleep 5
```
hooverlunch: Hooray for feature specs!

This `sleep` should not be necessary though. Capybara should know to wait for the content to appear (does it?). If it isn't, we'd need to troubleshoot that. But using `sleep` in Capybara is an anti-pattern.

hooverlunch:
Specs should be fixed now.

emitche:
Thank you!

app/models/accounting/transaction.rb
```diff
@@ -22,7 +22,7 @@
 #
 # Indexes
 #
-#  acc_trans_qbid_qbtype__unq_idx                          (qb_id,qb_transaction_type) UNIQUE
+#  acc_trans_qbid_qbtype_unq_idx                           (qb_id,qb_transaction_type) UNIQUE
```
emitche: Should this change? We've been keeping the double underscore.

hooverlunch: Crap, sorry. I have no idea. Up to you all.

PR #373, opened 2017-09-05
6686 loan dup first pass

X0nic:


spec/models/project_duplicator_spec.rb
```diff
@@ -35,14 +35,30 @@
...
+    # Using #reload does not do it.
+    let(:new_loan) { Loan.find(duplicator.duplicate.id) }
 
     it 'ignores media' do
       expect(new_loan.media.count).to eq 0
     end
 
-    it 'ignores timeline_entries' do
-      expect(new_loan.timeline_entries.count).to eq 0
+    it 'copies timeline_entries' do
```
X0nic: On the second pass, I will add an expectation to ensure the fields are being copied. No point in adding it before hand.

app/models/project_duplicator.rb
```diff
@@ -1,15 +1,36 @@
...
+    @copy = orig.amoeba_dup
+    copy.original = orig
+
+    duplicate_timeline_entry(orig.root_timeline_entry)
 
     copy.save!
     copy
   end
+
+  # make sure to set project_id on copy
```
hooverlunch: This comment is not really accurate anymore. Perhaps "Recursively copies timeline entries in pre-order traversal, ensuring correct parentage."

spec/models/project_duplicator_spec.rb
hooverlunch: Ok. When you are doing so, please be sure to add a test that the translations are being copied properly. I don't think we covered that this time and it's a big question mark for me at this point.

config/locales/en/loans.en.yml
```diff
@@ -22,6 +22,7 @@ en:
     details: "Details"
     duplicate: "Duplicate"
     duplicated_message: "The loan was successfully duplicated."
+    confirm_duplication: "This will duplicate the %{object_type} and all associated timeline entries, logs, and questionnaire responses. Do you want to continue?"
```
hooverlunch: I can see why you'd do it this way but using parameters in this way doesn't always work in languages other than english due to genders, agreements, etc. e.g. in spanish, step might be masculine while group might be feminine, etc. Can you please add one full translation per object type, perhaps nested under `confirm_duplication`?

spec/models/project_duplicator_spec.rb
```diff
@@ -56,5 +72,24 @@
...
+      it 'copies project logs' do
+        log = loan.project_logs.first
+        new_log = new_loan.project_logs.first
+
+        expect(new_log.id).not_to eq log.id
+        expect(new_log.date).to eq log.date
+        expect(new_loan.project_logs.count).to be > 0
+        expect(new_loan.project_logs.count).to eq loan.project_logs.count
+      end
+    end
```
hooverlunch: Very nice! Nailed this one!

app/models/project_duplicator.rb
```diff
@@ -1,15 +1,36 @@
...
+    end
+
+    # closure_tree doesn't seem to fire properly unless we save here.
+    copy_entry.parent = copy_parent_entry
+    copy_entry.save!
+
+    orig_entry.children.each do |child|
+      duplicate_timeline_entry(child, copy_entry)
+    end
+  end
```
hooverlunch: This class is looking so good!

PR #374, opened 2017-09-11
6946 date picker translation

hooverlunch:


spec/features/admin/date_picker_i18n_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      find("a", text: "Editar Préstamo").click
+      expect(page).to have_field("Fecha de Firma", with: "2017-04-25")
+      pick_aug_19(april_name: "Abril")
+      expect(page).to have_field("Fecha de Firma", with: "2017-08-19")
+      click_on("Actualizar Préstamo")
+      expect(page).to have_css(".loan_signing_date", text: "ago 19, 2017")
+    end
+  end
+
+  def pick_aug_19(april_name:)
```
linesarefuzzy: This is fine I think, but maybe just add a comment here and/or near the top of this file noting that this method and/or these tests only work in specific circumstances (a loan page in Edit mode with a signing date in April?).

emitche: Per @hooverlunch, he discussed with @linesarefuzzy. This is good to go and no changes need to be made here.

emitche:
6946-bug

PR #375, opened 2017-09-11
6941 filter questions by division

linesarefuzzy:


app/models/loan_filtered_question.rb
```diff
@@ -31,7 +35,7 @@ def required?
     if override_associations || depth == 1
       loan_types.include?(loan.loan_type_option)
     else
-      parent&.required?
+      parent.required?
```
linesarefuzzy: Why remove this?

app/models/loan_filtered_question.rb
```diff
@@ -2,6 +2,10 @@
 class LoanFilteredQuestion < FilteredQuestion
   attr_accessor :loan
 
+  # the super initialize is breaking this in the tests since it requires a division
+  # I'm thinking two things. At the moment, preferring 1
+    # create a separate class that inherits from filtered_questions for division
+    # update this initialize method, the tests for it and all other usages with division arg
```
linesarefuzzy: Actually, this subclass will need to pass in a division anyway. We can leave it failing for the moment.

spec/models/filtered_question_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  it 'filters correctly' do
+    expect(FilteredQuestion.new(q0, division: d0).visible?).to be_truthy
+    expect(FilteredQuestion.new(q1, division: d1).visible?).to be_truthy
+    expect(FilteredQuestion.new(q1, division: d11).visible?).to be_falsey
+    expect(FilteredQuestion.new(q1, division: d2).visible?).to be_falsey
+    expect(FilteredQuestion.new(q1, division: d21).visible?).to be_falsey
+    expect(FilteredQuestion.new(q1, division: d0).visible?).to be_truthy
+    expect(FilteredQuestion.new(q2, division: d1).visible?).to be_falsey
+    expect(FilteredQuestion.new(q2, division: d0).visible?).to be_truthy
```
linesarefuzzy: Any reason to use "truthy" and "falsey" instead of "true" and "false"? We expect the method to return a boolean, right?

spec/models/filtered_question_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+describe FilteredQuestion, type: :model do
+  let(:d0) { create(:division, name: 'Root Division') }
+  let(:d1) { create(:division, name: 'First Division') }
+  let(:d11) { create(:division, name: 'Child - First Division') }
+  let(:d2) { create(:division, name: 'Second Division') }
+  let(:d21) { create(:division, name: 'Child - Second Division') }
+
+  let(:q0) { create(:loan_question, division: d0) }
+  let(:q1) { create(:loan_question, division: d1) }
+  let(:q2) { create(:loan_question, division: d2) }
```
linesarefuzzy: I think we need to create questions for each of the divisions, including children.

spec/models/filtered_question_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+
+  # create division generations
+  before do
+    d0.add_child(d1)
+    d0.add_child(d2)
+    d1.add_child(d11)
+    d2.add_child(d21)
+  end
```
linesarefuzzy: This could also be accomplished by setting the `parent` attribute when creating the child divisions.

app/models/loan_filtered_question.rb
arisah-ijeoma: oops...was debugging failing tests and forgot to undo that change

out of curiosity though, what does that do?

spec/models/filtered_question_spec.rb
arisah-ijeoma: this also tests for booleans

app/models/loan_filtered_question.rb
arisah-ijeoma: alrighty

app/models/loan_filtered_question.rb
emitche: This ended up having an impact on @arisah-ijeoma's code. See 
https://github.com/sassafrastech/madeline/pull/380#pullrequestreview-62372571. So this code change actually needs to be discussed further.

emitche:
@arisah-ijeoma Is the only thing blocking this issue getting the specs to pass at this point? Is the rest of the work done?

PR #376, opened 2017-09-11
6539 media attachments loan memo

emitche:
- Show other images in the loan memo
- Show 8 images per page
- Include page breaks and loan memo headings every 8 images


app/assets/stylesheets/admin/loans/_print.scss
```diff
@@ -65,4 +67,26 @@
       min-height: inherit;
     }
   }
+
+  .images {
+    .image-list {
+      display: flex;
+      flex-direction: row;
+      align-items: flex-start;
+      flex-wrap: wrap;
```
hooverlunch: Hooray for flex!

app/assets/stylesheets/admin/loans/_print.scss
```diff
@@ -65,4 +67,26 @@
...
+    }
+
+    .img-container {
+      height: 2in;
+      width: 3.5in;
+      margin: .5em;
```
hooverlunch: Would it make sense to express the margin in inches too?

app/assets/stylesheets/admin/loans/_print.scss
```diff
@@ -65,4 +67,26 @@
...
+      border: 1px solid $gray3;
+    }
+
+    img {
+      max-height: 100%;
+      max-width: 100%;
```
hooverlunch: This was confusing at first but now I get it! This basically forces the image to grow to take up as much of the container as it can? Cool! Might be nice to put this as a comment...

app/controllers/admin/loans_controller.rb
```diff
@@ -117,7 +117,8 @@ def print
     authorize @loan, :show?
     @print_view = true
     @mode = params[:mode]
-    @first_image = @loan.media.find {|item| item.kind_value == 'image'}
+    @images = @loan.media.where(kind_value: "image")
+    @image_list = @images.drop(1)
```
hooverlunch: Cool! I never knew about `drop`...

hooverlunch: One other thing: if you do `@images.drop(1).each_slice(8).to_a` you'll get an array of arrays of 8 images each. Then you can use that to clean up the view. See below.

app/views/admin/loans/print.html.slim
```diff
@@ -9,8 +9,25 @@ section data-container="loan"
...
+    - if @images.first
+      section.featured-image
+        = image_tag(@images.first.item)
+
+    - unless @image_list.empty?
+      - @image_list.each_with_index do |image, index|
+        / HTML instead of Slim is used in this block to have control over how items are nested
+        / A new page and heading appears every 8 images
+        - if index % 8 == 0
+          <section class="images">
```
hooverlunch: I see what you're doing here. I appreciate the comment and bravo for figuring this out!

It's best to try to avoid conditionally rendering closing tags if at all possible because it's hard to read and error prone. We are doing something similar to this in `_questionnare_group.html.slim` and it's been a real pain. It's better to set up your model/data structure so that you can iterate over it in a normal way.

If you do as I suggest with `each_slice` above, you'll be able to just iterate over the slices first and then the individual images and you won't have to use HTML.

Please let me know if you try this and it doesn't work or if you need help.

Thanks! And again, nice job working through this and getting a working solution! It's not an easy problem!

app/assets/stylesheets/admin/loans/_print.scss
emitche: Yes. That's what is happening.

PR #377, opened 2017-09-12
6687 loan dup second pass

X0nic:
Use this for comparing changes with 6686

https://github.com/sassafrastech/madeline/compare/6686_loan_dup_first_pass...6687_loan_dup_second_pass?expand=1


X0nic:
This does work, and has specs covering the new feature. A strange thing happened though when the copy schedule feature was implemented. When the schedule copy code gets run, the children that get copied start returning in a non-deterministic order... Everything works, if you run tests one at a time, but not all at once. I don't think it is test leakage from previous tests, stuff just seems to be jumbled.

emitche:
6686 has been merged into develop. If you merge develop into this branch, that will show 6687 changes only.

PR #378, opened 2017-09-12
6687 loan dup second pass - 6687 changes only

hooverlunch:


emitche:
6686 is merged into develop.

PR #379, opened 2017-09-12
6687 loan dup second pass

hooverlunch:


app/models/project_group.rb
```diff
@@ -103,6 +108,9 @@ def filtered_children
     end
   end
 
+  # For use in specs
+  alias_method :c, :filtered_children
+
```
linesarefuzzy: Would it be better to do this in a monkey patch in the spec file?

spec/models/project_duplicator_spec.rb
```diff
@@ -92,4 +91,64 @@
...
+
+  context 'with scheduled children' do
+    # Creates a timeline and returns nodes stored in a hash.
+    let!(:nodes) { ProjectGroupFactoryHelper.create_full_timeline }
+    let(:loan) { nodes[:root].project }
+
+    # Break each of the nodes out into a let so that we can examine them individually.
+    ProjectGroupFactoryHelper::NODE_NAMES.each do |name|
+      let(name) { nodes[name] }
+    end
```
linesarefuzzy: Could you do this instead:
```
nodes.each { |k, v| let(k) { v } }
```
and then eliminate the manually synced `NODE_NAMES` constant?

spec/models/project_duplicator_spec.rb
```diff
@@ -92,4 +91,64 @@
...
+
+    before do
+      # See ProjectGroupFactoryHelper for the layout of nodes in the timeline.
+      # We introduce some schedule dependencies in such a way that one of the dependencies
+      # actually goes downward instead of upward.
+      # g3_s3 is sorted above s1 because g3 has an early step. But we're going to make g3_s3 depend
+      # on s1, which won't change the sort order, but is a downward depenendency.
+      g3_s3.update_attributes!(schedule_parent_id: s1.id)
+      g5_s1.update_attributes!(schedule_parent_id: g3_s3.id)
```
linesarefuzzy: Why use the more verbose version of `update`?

spec/models/project_duplicator_spec.rb
```diff
@@ -92,4 +91,64 @@
...
+        expect(g3_s3.schedule_parent).to eq s1
+
+        expect(g5_s1.scheduled_start_date).to eq Date.parse('2017-04-06')
+        expect(g5_s1.scheduled_duration_days).to eq 3
+        expect(g5_s1.schedule_parent).to eq g3_s3
+      end
+    end
+
+    context 'original loan' do
+      let(:loan_to_test) { loan }
```
linesarefuzzy: Could you use `subject` instead?

spec/models/project_duplicator_spec.rb
linesarefuzzy: Apparently not—unless you're willing to accept `@` signs in front of the variable names, which doesn't seem all that onerous to me... 

Inside `before`:
```
 nodes.each { |k, v| instance_variable_set("@#{k}", v) }
```

spec/models/project_duplicator_spec.rb
linesarefuzzy: Yes. Done.

hooverlunch:
Outside the spec world: fake data script.

spec/models/project_duplicator_spec.rb
hooverlunch: No reason. Old habits die hard? I'll try to remember next time.

spec/models/project_duplicator_spec.rb
hooverlunch: @'s are bad. The whole point of lets is that they "let" you move stuff around, up and down the scope tree.

app/models/project_group.rb
hooverlunch: Not a bad idea.

spec/factories/project_groups.rb
```diff
@@ -11,56 +11,98 @@
...
+      nodes[:s1] = create_step(nodes[:root], "2017-02-28", 30)
+    end
+  end
 
-    # Randomly sometimes don't include dates.
-    attribs[:scheduled_start_date] = nil if rand(5) == 0
+  def self.create_group(parent)
+    FactoryGirl.create(:project_group, project: parent.project, parent: parent).tap do |group|
+      parent.children << group
+    end
```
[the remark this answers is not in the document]
hooverlunch: Good point. The reason it doesn't "take" I think is because that would require saving the parent simply as a result of assigning it as the parent of some other object, which I think could be seen as heavy handed. Not sure though.

spec/factories/project_groups.rb
```diff
@@ -11,56 +11,98 @@
...
+      nodes[:g4] = create_group(nodes[:root])
+      nodes[:g4_s1] = create_step(nodes[:g4], "2017-03-01", 0)
+      nodes[:g2] = create_group(nodes[:root])
+      nodes[:g2_g2] = create_group(nodes[:g2])
+      nodes[:g2_g2_s1] = create_step(nodes[:g2_g2], "2017-01-10", 5)
+      nodes[:g2_g1] = create_group(nodes[:g2])
+      nodes[:g2_g1_s1] = create_step(nodes[:g2_g1], "2017-01-01", 5)
+      nodes[:s2] = create_step(nodes[:root], nil, 0)
+      nodes[:s1] = create_step(nodes[:root], "2017-02-28", 30)
+    end
```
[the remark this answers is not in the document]
hooverlunch: Yes you are 100% correct.

app/models/project_duplicator.rb
```diff
@@ -29,8 +31,27 @@ def duplicate_timeline_entry(orig_entry, copy_parent_entry = nil)
...
+      if old_step.schedule_parent_id
+        new_step.schedule_parent_id = step_map[old_step.schedule_parent].id
+      else
+        new_step.scheduled_start_date = old_step.scheduled_start_date
+      end
+      new_step.scheduled_duration_days = old_step.scheduled_duration_days
+
+      new_step.save!
+    end
+  end
```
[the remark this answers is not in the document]
hooverlunch: Credit to @X0nic on this one.

linesarefuzzy:
Do you think the fake data script possibility is worth the manually synced constant?

hooverlunch:
It's not a possibility. It exists! Take a look: :p

PR #382, opened 2017-09-15
6941 filter qtn by dvsn

emitche:


app/models/loan_response.rb
```diff
@@ -54,9 +54,10 @@ def field_attributes
     end
   end
 
+  # Checks if response is blank, including any descenants if this is a group.
   def blank?
     if group?
-      loan_question.descendants.all? { |i| loan_response_set.response(i.id).blank? }
+      question.children.all?(&:blank?)
```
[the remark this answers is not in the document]
emitche: Using `question.children.all?(&:blank?)` causes a regression. The questions show up in the loan memo. If a question or group is unanswered, it should not appear in the loan memo. This code change makes these empty groups appear in the printed loan memo, which should not be the case. They show up both in the table of contents and in the following pages, but should not appear at all.

Because `descendants` is a banned method, this would need to be revised in order for this to work with the permitted methods while not causing a regression.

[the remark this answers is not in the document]
emitche: I was unable to make a fix for this regression.

app/views/admin/loans/questionnaires/_questionnaire_toc.html.slim
```diff
@@ -1,8 +1,7 @@
 ul
   - group.child_groups.each do |child|
-    - response = response_set.response(child.id)
+    - response = response_set.response(child)
```
arisah-ijeoma: 👍

app/models/concerns/progress_calculable.rb
```diff
@@ -52,7 +47,17 @@ def progress_denominator
...
+  # If the current response is required, only count children that are also required.
   def progress_applicable_children
-    required? ? active_children.select(&:required?) : active_children
+    children.select do |c|
+      if self.required?
+        c.active? && c.required?
+      else
+        c.active?
+      end
+    end
```
arisah-ijeoma: noice

app/models/loan_filtered_question.rb
```diff
@@ -29,7 +29,7 @@ def required?
     @required ||= if override_associations || depth == 1
       loan_types.include?(loan.loan_type_option)
     else
-      parent&.required?
+      root? ? true : parent.required?
```
arisah-ijeoma: what's the difference between `parent&.required?` and `parent.required?`

app/controllers/application_controller.rb
```diff
@@ -48,7 +48,7 @@ def set_locale
   end
 
   def locale_from_header
-    if !Rails.env.test? && (lang_header = request.env['HTTP_ACCEPT_LANGUAGE'])
+    if !Rails.env.test? && (lang_header = request.env['HTTP_ACCEPT_LANGUAGE']).present?
```
arisah-ijeoma: wouldn't this set `lang_header` to a boolean?

app/models/filtered_question.rb
```diff
@@ -1,14 +1,17 @@
 # Wraps LoanQuestion and delegates most methods, but enables filtering by loan and division via subclasses.
 class FilteredQuestion < SimpleDelegator
-  def self.decorate_collection(collection, **args)
-    collection.map { |q| self.new(q, **args) }
-  end
-
   def initialize(question, **args)
     super(question)
+    @division = args[:division]
```
arisah-ijeoma: yaay! finally found how to use args

app/models/filtered_question.rb
```diff
@@ -46,6 +58,7 @@ def self_and_descendants
   protected
 
   def decorated_children
+    p @args
```
arisah-ijeoma: forgotten?

app/models/loan_filtered_question.rb
```diff
@@ -2,16 +2,14 @@
 class LoanFilteredQuestion < FilteredQuestion
   attr_accessor :loan
 
-  def initialize(question, loan:)
-    super(question, loan: loan)
-    @loan = loan
+  def initialize(question, **args)
+    @loan = args[:loan]
+    args[:division] = @loan.division
+    super(question, **args)
```
arisah-ijeoma: 👍

arisah-ijeoma:
the filter is not working (at least on the manage loan question page)

emitche:
This issue will be ready to be reviewed again once the outstanding pull request, #388, is merged into this one.

emitche:
I'm using a user who is the admin of the root division. This is what I see on the `develop` branch when I view Loan Questions:
image

This is what I see on this branch when I view loan questions:
image

It looks like some of the loan questions inside Manage Questions are being hidden. However, I thought a root division admin would be able to see all questions.

hooverlunch:
@emitche Great question. Very astute of you to have notice this. I'm impressed! Great QA!

I wonder if group 1.1 belongs to a child division? Under this story, you only see questions that are in the currently selected division _or its ancestors_. This is so that lower divisions can add lots of custom questions without cluttering up the top-level list.

emitche:
@hooverlunch Interesting. Yes, CONAMI is a child of La Base Argentina. So now only the questions for the currently selected division and its ancestors will show up. That makes sense. This is working as expected then. Thank you!

emitche:
Some of the spec for 6941 negates work in 6942. I'm sifting through that and will make a plan for fixing those items.

emitche:
I've decided to QA just what was in the spec for 6941 and correct the concerns I raised today in 7235.

emitche:
Great work here. Sorry I was confused about many things when doing the QA for this. This is good to go.

PR #383, opened 2017-09-19
6849 calendar events font size

emitche:
- Make step and loan events have same font-size in calendar
- Truncate loan names in calendar


emitche:
@linesarefuzzy, these are UX tweaks.

emitche:
This small CSS change does not warrant a code review. However, a UX review is needed because pairing did not happen on this tweak.

linesarefuzzy:
Looks great, thanks!

PR #384, opened 2017-09-19
6851 adjust step events design

emitche:
- Make steps in calendar take one line instead of 2
- Refine icons in step events

### DIFF
--- a/app/assets/stylesheets/admin/calendar/_calendar.scss
+++ b/app/assets/stylesheets/admin/calendar/_calendar.scss
@@ -27,7 +27,6 @@
 
       &.fa-link {
         color: $gray3;
-        padding-right: .25em;
       }
     }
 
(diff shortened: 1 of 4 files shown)


emitche:
Fuzzy, these are UX tweaks.

emitche:
@hooverlunch please review this one. I can have @linesarefuzzy do a QA and UX review afterwards. Thank you.

app/assets/stylesheets/admin/calendar/_calendar.scss
```diff
@@ -25,9 +25,8 @@
       font-size: 1em;
       line-height: 1.2em;
 
-      &.fa-link {
-        color: $gray3;
-        padding-right: .25em;
+      span {
+        font-family: "Open Sans",sans-serif;
```
hooverlunch: I think this should be variable, shouldn't it? If these are the main fonts for the admin panel and there is not already of variable, we should make one. Or if this is some special style,We could have a variable for that too.

What do you think?

emitche: I'll look into it. It sounds sensible to make the fonts a variable.

linesarefuzzy:
OMG, it looks SO much better! Thank you. I noticed a very minor bug where the start date is sometimes duplicated in the modal, but it exists on develop too, so I'll make a new issue for it.

PR #385, opened 2017-09-20
6927 table headings

emitche:
Add visual distinction between dashboard table header and rows

### DIFF
--- a/app/assets/stylesheets/admin/dashboard/_dashboard.scss
+++ b/app/assets/stylesheets/admin/dashboard/_dashboard.scss
@@ -98,4 +98,12 @@
       font-size: 1em;
     }
   }
+
+  table {
+    // Specificuty added to override defaults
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/admin/dashboard/_dashboard.scss
```diff
@@ -98,4 +98,12 @@
       font-size: 1em;
     }
   }
+
+  table {
+    // Specificity added to override defaults
```
hooverlunch: Nice comment

linesarefuzzy:
Looks fine. Some of these changes—at least this one, in my opinion—you could just push right to develop. Here's the question I ask myself: Am I pretty sure it's a minor, uncontroversial improvement? Then if it turns out to be controversial after all, undoing a minor change is very easy. 

In any event, thank you for spotting and fixing all these little things!

PR #386, opened 2017-09-21
7120: Fixed signing_date nil error and added spec

hooverlunch:


hooverlunch:
I didn't pair with fuzzy on this because it was such a simple fix but I'm putting him as the reviewer since is the assigned pair. Just a suggestion though.

hooverlunch:
One request: in the future when we are creating issues from error emails, can we please include all the information. It took a while to figure out which request was triggering this error Because I only had the stack trace. Thanks!

spec/models/loan_health_check_spec.rb
```diff
@@ -154,6 +155,12 @@
 
               it { is_expected.to_not include :sporadic_loan_updates }
             end
+
+            context 'with no signing_date' do
+              let(:signing_date) { nil }
+
+              it { is_expected.to include :sporadic_loan_updates }
+            end
```
linesarefuzzy: Does this mean a loan with no signing date will always trigger the "Less than one step every month" warning?

spec/models/loan_health_check_spec.rb
```diff
@@ -121,7 +121,8 @@
           end
 
           context 'ending in 90 days' do
-            let(:loan) { create(:loan, :prospective, end_date: 90.days.from_now) }
+            let(:loan) { create(:loan, :prospective, signing_date: signing_date, end_date: 90.days.from_now) }
+            let(:signing_date) { Date.today }
```
linesarefuzzy: I'm a little confused about this. Are you using a variable before it's assigned?

hooverlunch: Lazy execution of the let's allows this

spec/models/loan_health_check_spec.rb
hooverlunch: Hmm yeah maybe thats bad. Can you fix please? I am swamped.

emitche:
@linesarefuzzy, do you have capacity to finish up this issue?

spec/models/loan_health_check_spec.rb
hooverlunch: Actually no, it doesn't mean that. It is only triggering that warning because it has no future steps, and if there is no signing date, Then it defaults To the current date. If I move this Context block it will be clearer.

PR #387, opened 2017-09-22
7134: Show project name in show view of modal

emitche:
- Show project name in the show view of the project step modal
- Ensures general calendar and dashboard calendar step events have project context

The show view of the project step modal appears in all calendars, but is not displayed in the timeline.

### DIFF
--- a/app/controllers/admin/project_steps_controller.rb
+++ b/app/controllers/admin/project_steps_controller.rb
@@ -192,6 +192,7 @@ class Admin::ProjectStepsController < Admin::AdminController
 
   def render_modal_content(status = 200)
     @mode = params[:action] == "show" ? :show_and_form : :form_only
+    @project = Project.find(@step.project_id)
     @parents = @step.project.timeline_groups_preordered
     @precedents = @step.project.timeline_entries.where("type = 'ProjectStep' AND id != ?", @step.id || 0).by_date
     render partial: "/admin/project_steps/modal_content", status: status, locals: {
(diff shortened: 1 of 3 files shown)


emitche:
I did not see a quick way to not show project name inside the project calendar. We have a context variable, but it is set to "calendar" for all 3 types of calendars.

app/controllers/admin/project_steps_controller.rb
```diff
@@ -192,6 +192,7 @@ def render_step_partial(mode)
 
   def render_modal_content(status = 200)
     @mode = params[:action] == "show" ? :show_and_form : :form_only
+    @project = Project.find(@step.project_id)
```
hooverlunch: Would `@project = @step.project` work?

emitche: I will check. Thanks!

linesarefuzzy:
Looks great, just one suggestion: I think it would be nice to have the project name link to the project. I'm not sure the best way to do that without knowing what type of project it is, though, since loans and basic projects have different paths. Which makes me wonder why that is. @hooverlunch is there any good reason not to change the loan and basic project paths both to `/projects`?

linesarefuzzy:
Also, it's not that big a deal. This could totally be merged and released as is, without the link.

emitche:
I think it would be pretty simple to link to the project. I'll make that change.

emitche:
Linking was not as straight-forward as I thought it would be. Will the approach I took have an impact on performance?

We can consider refactoring all projects to use the same path (admin/projects/:id). But I think that refactor should happen in a different issue. That way we can make sure to catch all the places where both `admin/loans/:id` and `admin/basic-projects/:id` are used.

emitche:
@linesarefuzzy, you and I will need to decide on this, as hooverlunch has a lot going on right now.

emitche:
This has received QA and only needs a small tweak.

emitche:
We already had in use a simple way to load the project path, regardless of type of project. That has been implemented in this branch.

We don't need to make basic projects and loans use the same path at this point, given that we have a good solution for that. But if you want to consider it in the future, please do.

PR #390, opened 2017-09-26
6241 media captions

emitche:
- If caption available, show caption underneath media


app/helpers/media_helper.rb
```diff
@@ -27,6 +27,24 @@ def media_title(media_item, shorten: true)
     end
   end
 
+  def media_caption(media_item, shorten: true)
+    caption = media_item.send("caption_#{I18n.locale}")
+
+    if caption && caption.text && !caption.text.empty?
```
arisah-ijeoma: I think `caption.text` is redundant here
`caption && !caption.text.empty?` should be sufficient

emitche: I'll make that change. Thanks!

PR #391, opened 2017-09-26
7209 media alt text

emitche:
- Ensure media image alt text does not take up more room than expected
- Use the caption as alt text, if available, instead of auto-generated alt text

### DIFF
--- a/app/helpers/media_helper.rb
+++ b/app/helpers/media_helper.rb
@@ -1,7 +1,8 @@
 module MediaHelper
   def media_thumbnail(media_item)
     if media_item.thumbnail?
-      image_tag(media_item.item.thumb.url)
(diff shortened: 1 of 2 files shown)


app/helpers/media_helper.rb
```diff
@@ -1,7 +1,13 @@
...
   def media_thumbnail(media_item)
     if media_item.thumbnail?
-      image_tag(media_item.item.thumb.url)
+      if media_item.caption.text.present?
+        alt_text = truncate(media_item.caption.text, length: 36)
+        return image_tag(media_item.item.thumb.url, alt: alt_text)
+      else
+        # Use auto-generated alt text
+        return image_tag(media_item.item.thumb.url)
+      end
```
arisah-ijeoma: 👍

PR #392, opened 2017-10-02
4834 link icons date dependency

emitche:
- Show link icons to represent step dependency in timeline


emitche:
This is a tiny code change that does not need code review.

PR #394, opened 2017-10-04
6852 unfinalized steps styling in loan calendar

emitche:
- Style draft steps inside the loan calendar
- Ensure colors chosen are color contrast accessible
- Adjust the calendar legend

### DIFF
--- a/app/assets/javascripts/backbone/views/calendar_view.coffee
+++ b/app/assets/javascripts/backbone/views/calendar_view.coffee
@@ -18,6 +18,7 @@ class MS.Views.CalendarView extends Backbone.View
     options = $.extend defaultCalendarSettings, settings
     @$calendar.fullCalendar(options)
 
+
   stepClick: (e) ->
     @stepModal.show(@$(e.currentTarget).data('step-id'), @refresh.bind(@))
 
@@ -29,6 +30,12 @@ class MS.Views.CalendarView extends Backbone.View
   # Load custom content inside the fullCalendar rendered event rather than title only
   eventAfterRender: (calEvent, initialElement) ->
     $(initialElement).find('.fc-content').append(calEvent.html)
+
+    if calEvent.is_finalized
+      $(initialElement).addClass('finalized-step')
+    else
+      $(initialElement).addClass('draft-step')
+
     $(initialElement).addClass(calEvent.event_classes)
     $(initialElement).data('step-id', calEvent.model_id)
     $(initialElement).find('.fc-title').remove()
--- a/app/assets/javascripts/backbone/views/calendar_view.coffee
+++ b/app/assets/javascripts/backbone/views/calendar_view.coffee
@@ -32,9 +32,9 @@ class MS.Views.CalendarView extends Backbone.View
     $(initialElement).find('.fc-content').append(calEvent.html)
 
     if calEvent.is_finalized
-      $(initialElement).addClass('finalized-step')
+      $(initialElement).addClass('finalized')
     else
-      $(initialElement).addClass('draft-step')
+      $(initialElement).addClass('draft')
 
     $(initialElement).addClass(calEvent.event_classes)
     $(initialElement).data('step-id', calEvent.model_id)
--- a/app/assets/javascripts/backbone/views/calendar_view.coffee
+++ b/app/assets/javascripts/backbone/views/calendar_view.coffee
@@ -29,12 +29,6 @@ class MS.Views.CalendarView extends Backbone.View
   # Load custom content inside the fullCalendar rendered event rather than title only
   eventAfterRender: (calEvent, initialElement) ->
     $(initialElement).find('.fc-content').append(calEvent.html)
-
-    if calEvent.is_finalized
-      $(initialElement).addClass('finalized')
-    else
-      $(initialElement).addClass('draft')
-
     $(initialElement).addClass(calEvent.event_classes)
     $(initialElement).data('step-id', calEvent.model_id)
     $(initialElement).find('.fc-title').remove()
--- a/app/assets/javascripts/backbone/views/calendar_view.coffee
+++ b/app/assets/javascripts/backbone/views/calendar_view.coffee
@@ -18,7 +18,6 @@ class MS.Views.CalendarView extends Backbone.View
     options = $.extend defaultCalendarSettings, settings
     @$calendar.fullCalendar(options)
 
-
   stepClick: (e) ->
     @stepModal.show(@$(e.currentTarget).data('step-id'), @refresh.bind(@))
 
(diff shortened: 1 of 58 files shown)


app/assets/stylesheets/admin/_colors.scss
```diff
@@ -26,17 +26,19 @@ $background-gray: rgb(243, 243, 243);
...
+$gray8: #545454;  // Meets accessibility standard on $table-gray
+$gray9: #595959; // Meets accessibility standards on white background
 $body: #333;
 $black: black;
 
 // Color values from project step model
-$onTime: hsl(120, 73%, 57%);
-$superEarly: hsl(120, 41%, 47%);
-$barelyLate: hsl(56, 100%, 66%);
-$superLate: hsl(0, 74%, 54%);
```
emitche: hsl is a bit non-standard in terms of using web development tools. Having the rgb values available makes things easier for future styling.

app/assets/stylesheets/admin/_variables.scss
```diff
@@ -3,3 +3,9 @@
 $content-wrapper-padding: 20px;
 $sidebar-width: 190px;
 $app-fonts: 'Open Sans', sans-serif;
+
+// Common styles
```
emitche: Is there a better place to put things if using `@extend`?

app/assets/stylesheets/admin/_variables.scss
```diff
@@ -3,3 +3,9 @@
...
+.draft {
```
emitche: Intended to dry up step draft styles. We use something similar in both the timeline in the calendar. If we change either in the future, the change can touch both places at once.

app/assets/stylesheets/admin/loans/timeline/_table.scss
```diff
@@ -1,4 +1,5 @@
 @import "../../colors";
+@import "../../variables";
```
emitche: I'd ideally like to get away from importing variables and colors. Fixing the underlying bug that requires this as a fix is part of our backlog.

app/views/admin/calendar/_legend.html.slim
```diff
@@ -4,20 +4,23 @@ div#legend-content
       div.column.col-lg-6
         table.legend-block
           tr
+            td
```
emitche: Draft steps do not exist in the main calendar. Do we need to have a revised list of legend items for the general calendar or is that not needed?

app/assets/javascripts/backbone/views/calendar_view.coffee
```diff
@@ -29,6 +29,12 @@ class MS.Views.CalendarView extends Backbone.View
   # Load custom content inside the fullCalendar rendered event rather than title only
   eventAfterRender: (calEvent, initialElement) ->
     $(initialElement).find('.fc-content').append(calEvent.html)
+
+    if calEvent.is_finalized
+      $(initialElement).addClass('finalized')
+    else
+      $(initialElement).addClass('draft')
```
hooverlunch: Given that CalendarEventSerializer already has the event_classes method, perhaps better to put this logic in there?

app/assets/stylesheets/admin/_colors.scss
hooverlunch: Is there any reason not to just convert to RGB? I would be fine with that.

app/assets/stylesheets/admin/_variables.scss
hooverlunch: Great idea! If you intend for this style to just be applicable to project steps, It should maybe go in the project steps folder? Or, if it's going to be an app wide style for draft-type things, Then maybe it should go in _general.scss?

app/assets/stylesheets/admin/_variables.scss
```diff
@@ -1,5 +1,13 @@
+@import "colors";
+
 // Common variables
 
 $content-wrapper-padding: 20px;
 $sidebar-width: 190px;
 $app-fonts: 'Open Sans', sans-serif;
+
```
hooverlunch: Whoops, I accidentally deleted your comment here. See below.

app/assets/stylesheets/admin/calendar/_calendar.scss
```diff
@@ -79,6 +79,12 @@
     }
   }
 
+  .cal-step {
+    &.draft {
+      @extend .draft;
+    }
```
hooverlunch: I'm not sure that you need to use @extend here or below. If you define a global class, it should Already apply to any Element with that class, right? Try it without?

app/assets/stylesheets/admin/loans/timeline/_table.scss
hooverlunch: Amen!

app/views/admin/calendar/_legend.html.slim
hooverlunch: I don't think it's a big deal.

app/assets/javascripts/backbone/views/calendar_view.coffee
emitche: I tried, but was having trouble with it working, so this was the best solution I could come up with. I'm not sure why it was not working.

app/assets/stylesheets/admin/calendar/_calendar.scss
emitche: That makes sense. I overcomplicated things I think.

app/views/admin/calendar/_legend.html.slim
emitche: Ok.

app/assets/javascripts/backbone/views/calendar_view.coffee
emitche: I figured out what I needed to change. Fixed. Thanks!

emitche: Update: The problem is that the event serializer does not successfully distinguish between finalized and draft steps. I'm not sure why that happens. But finalized steps get the `.draft` class though they are not drafts.

emitche:
This has received code review and just needs a QA review. Thank you, @linesarefuzzy!

app/assets/stylesheets/admin/_colors.scss
linesarefuzzy: The reason I used HSL here is that the project step model does calculations that require colors in HSL format. The colors are defined in that model as well as here. I think it's useful to have them defined in the same way in both places.

emitche:
I'll take a look. Thanks!

emitche:
@linesarefuzzy: The issue you brought up during QA has been addressed. Is anything else needed?

linesarefuzzy:
Yes, thank you! However, I just did another QA review and noticed that incomplete steps are not being bolded, yet the legend says they are. It's that way in develop too. I'll make an issue.

emitche:
Yes, I'll make a ticket for that. Thanks for letting me know.

emitche:
I'll merge this into develop.

app/assets/stylesheets/admin/_colors.scss
emitche: I'll keep the values in hsl but keep rgb in the comments.

PR #396, opened 2017-10-11
7277 loan health error

hooverlunch:


spec/models/filtered_question_spec.rb
```diff
@@ -15,66 +15,85 @@
...
-      let!(:user) { create(:user, :admin, division: d1) }
+      let(:user) { create(:user, :admin, division: d1) }
 
       it 'should still be able to see questions from ancestor division' do
+        # This is usually not permitted for most objects. Loan questions are different.
         expect(filtered_question(q0, d0)).to be_visible
       end
+
+      it 'should not be able to see questions in divisions outside own acestors and descendants' do
+        expect(filtered_question(q2, d2)).not_to be_visible
```
linesarefuzzy: I think this is an illegal option—it's saying this user (who belongs to d1) selected d2 in the division select menu. They shouldn't be able to see d2 in that menu. For that matter, I think the test above has the same problem. Both should probably pass in nil for division, or if that doesn't work, the user's division. In fact maybe the filtered_question method should have a default of nil for that argument. A comment making it clear that the division argument represents the division selected in the menu might also be helpful.

spec/rails_helper.rb
```diff
@@ -65,6 +65,11 @@
     DatabaseCleaner.start
   end
 
+  config.before(:each) do
+    # Create root division
+    Division.create!(name: '-')
+  end
+
```
linesarefuzzy: What's the reasoning behind this change? Doesn't it mean an extra db create statement in cases where it's not needed?

spec/models/filtered_question_spec.rb
hooverlunch: Yes, you are right, but it is still the correct behavior for the class were it to receive that argument. The class knows nothing about the division drop-down.

The reason I put this in was that I needed some kind of negative test for the addition of the policy. It turns out that checking the loan question policy in the filtered question decorator Does not reduce the available set of questions at all, Assuming the arguments to the decorator are "legal" as you define it. So there is no way to test that The policy is actually being checked without Doing what I did here.

I do think it is still a good idea to check the loan question policy, just in case it changes in the future.

spec/rails_helper.rb
hooverlunch: Yes, however it is a very fast Database query, and not having it results in some really weird Test failures that are hard to debug. So while this might add 1-2 seconds to the test suite run, I think it is worth it.

PR #398, opened 2017-10-24
6907 push transaction changes

X0nic:


app/controllers/admin/accounting/transactions_controller.rb
```diff
@@ -20,8 +20,8 @@ def create
       if Rails.env.test?
         journal_entry = Quickbooks::Model::JournalEntry.new(id: 123)
       else
-        creator = ::Accounting::Quickbooks::TransactionCreator.new
-        journal_entry = creator.create_in_qb @transaction
+        reconciler = ::Accounting::Quickbooks::TransactionReconciler.new
```
X0nic: There is a bug in the `TransactionCreator` that will fail when trying to create objects in QBO. The `TransactionReconciler` still uses this class internally, so it still has the bug. I assume there is some relation to the `InterestCalculator` that needs to be sorted out here. I will need help from @hooverlunch or @linesarefuzzy for this.

X0nic:
This does what is outlined in the ticket, and the test pass. There is still an issue that needs to be addressed in order for this to work with the actual QBO API.

emitche:
Marking this as `ready-for-review` in addition to `help-wanted` so we can get a start on code review for this.

Thanks, @X0nic.

linesarefuzzy:
I did not mean to close this branch, it closed automatically because I deleted the branch it was based on.

linesarefuzzy:
All tests are passing. I believe QA can't be done on this issue in isolation, so I'm going to merge it and we'll QA the whole system at once.

PR #400, opened 2017-10-25
6713 only show one flash message at a time on transaction page

X0nic:


hooverlunch:
I think we can perhaps ignore the failing spec here for now until everything is integrated. There are just so many moving pieces it's not clear what's causing this failure.

emitche:
@X0nic: Please merge develop into this branch and resolve any merge conflicts. Thank you!

X0nic:
This should match what you need @linesarefuzzy . Ping me if something jumps out at you. I had real problems trying to test this with live QBO data.

app/models/division.rb
```diff
@@ -138,6 +138,7 @@ def accounts
   end
 
   def qb_accounts_connected?
+    return true if Rails.env.test?
```
X0nic: If anyone has a better way of mocking this for flow specs?

hooverlunch:
This should be good to go if the specs pass. I'll be watching.

@linesarefuzzy There will probably be some ugly merge conflicts with the integration work and this work (once merged to develop).

emitche:
The add transaction button should not show up when QuickBooks is connected.

image

emitche:
When transactions do not exist and QuickBooks is connected, should we be seeing this? Or should there be a message that says there are no transactions?

image

emitche:
Redmine#7489 was created to address these questions.

PR #401, opened 2017-10-30
a fix?

arisah-ijeoma:
I was not able to recreate this bug but from the error reported, it was a nil value so I added an extra check for nil


linesarefuzzy:
Yup, it works! Thanks Jay.

PR #402, opened 2017-11-15
7354 wide timeline table

emitche:
- Ensure timeline table has a gray background
- Ensure timeline dropdown and dropup menus fully display

### DIFF
--- a/app/assets/javascripts/backbone/views/timeline_table_view.coffee
+++ b/app/assets/javascripts/backbone/views/timeline_table_view.coffee
@@ -162,8 +162,9 @@ class MS.Views.TimelineTableView extends Backbone.View
     @logListModalView.show(stepId, @refresh.bind(@))
 
   styleDropdowns: ->
+    # Make top 4 rows of timeline have dropdown menus instead of dropup menus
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/backbone/views/timeline_table_view.coffee
```diff
@@ -158,3 +160,11 @@ class MS.Views.TimelineTableView extends Backbone.View
     unless @logListModalView
       @logListModalView = new MS.Views.LogListModalView(el: @$('#log-list-modal'))
     @logListModalView.show(stepId, @refresh.bind(@))
+
+  styleDropdowns: ->
+    # Make top 4 rows of timeline have dropdown menus instead of dropup menus
+    $topRows = @$el.find('tbody tr:nth-child(-n+4)')
```
hooverlunch: Neat nth-child work. Didn't know you could do that.

emitche:
The table should line up with the buttons. There is too much padding around the table:
image

Below is the way this looks in develop. Spacing and padding should be similar to the below.
image

emitche:
The blank space without steps does look empty. But it's minor. Given what this solves (menus that can be cut off), it is better to keep the large blank space and have working menus.

image

PR #404, opened 2017-11-17
6847 diff pry and sec agents

arisah-ijeoma:


app/models/loan.rb
```diff
@@ -211,4 +212,12 @@ def healthy?
   def health_status_available?
     return !health_check.nil?
   end
+
+  def agents_the_same?
+    (primary_agent.present? || secondary_agent.present?) && primary_agent == secondary_agent
+  end
+
+  def check_agents
+    errors.add(:primary_agent, "can't be the same as secondary agent") if agents_the_same?
```
emitche: We'll replace this error with an error from the yml file. Let's use this error message for now:

"The point person for this project cannot be the same as the second point person."

app/models/loan.rb
```diff
@@ -211,4 +212,12 @@ def healthy?
   def health_status_available?
     return !health_check.nil?
   end
+
+  def agents_the_same?
+    (primary_agent.present? || secondary_agent.present?) && primary_agent == secondary_agent
+  end
+
+  def check_agents
+    errors.add(:primary_agent, 'for this project cannot be the same as the second point person') if agents_the_same?
```
emitche: Thanks. Please move the error message into translations so that this can also be translated into Spanish. Let me know if you need help with this or if this is not clear. Thank you so much!

arisah-ijeoma: sorry! forgot about that since a portion of it is translated
will do so now

emitche: Thank you!

app/views/admin/loans/new.html.slim
```diff
@@ -4,3 +4,8 @@ section.loan.new-view
 
   section.block
     = render 'admin/loans/form'
+
+javascript:
+  $(function(){
+    new MS.Views.LoanView({});
+  });
```
hooverlunch: If we agree to take the frontend stuff out, this would go too.

app/assets/javascripts/backbone/views/loan_view.coffee
```diff
@@ -0,0 +1,21 @@
...
+    selected = e.options[e.selectedIndex].value
+
+    $('#loan_secondary_agent_id').find('option[value=' + selected + ']').remove()
+
+  removeFromPryAgentList: ->
+    e = document.getElementById('loan_secondary_agent_id')
+
+    selected = e.options[e.selectedIndex].value
+
+    $('#loan_primary_agent_id').find('option[value=' + selected + ']').remove()
```
hooverlunch: This is a good first attempt but there is a problem here. If you accidentally select "Sally" from the primary agent list but she's really a secondary agent, you can never fix your mistake and select her from the secondary list. You'd have to refresh the page and start again.

While I appreciate the effort here, I would suggest removing this front-end functionality. I think the backend validation will be enough to keep data clean and provide a decent user experience. It's up to you @emitche and @linesarefuzzy as planners, but I think to get this working in JS is going to take quite a bit more work.

emitche: Thanks so much for all of this work, @arisah-ijeoma.

Let's remove the front end portion of this change and keep the back-end portions. Thanks again.

PR #405, opened 2017-11-17
update codebase with new date format

arisah-ijeoma:


arisah-ijeoma:
I spoke to Tom about it and he said it was fine to remove it

emitche:
Cool. Thanks, @arisah-ijeoma!

app/models/loan.rb
```diff
@@ -97,7 +97,7 @@ def recalculate_loan_health
   def default_name
     if organization
       date = signing_date || created_at.to_date
-      "#{organization.name} - #{I18n.l(date)}"
+      "#{organization.name} - #{ApplicationController.helpers.ldate(date)}"
```
hooverlunch: Nice job getting this to work but we definitely want to avoid coupling the model to a view helper. This is not a good dependency.

Ideally we would not be doing this kind of work in the model at all, but in a decorator. However, since we're not doing decorators I think we should do the following:

1. If we know for sure that the date value can't be nil, then leave the code as it was.
1. If we don't know for sure, then you could use `I18n.l(nil, default: "")`. This will achieve the same effect without adding a new dependency.

(By the way, how I figured out `default: ""` was by looking at the `l` method source code: https://github.com/svenfuchs/i18n/blob/master/lib/i18n/backend/base.rb#L64

app/views/admin/notes/_note.html.slim
```diff
@@ -10,6 +10,6 @@
   javascript:
     $(function() {
       new MS.Views.NoteView({
-        el: '.note[data-id=#{note_id}]',
+        el: '.note[data-id=#{note_id}]'
```
hooverlunch: 👍

app/views/admin/timeline/_step.html.slim
```diff
@@ -29,15 +29,15 @@
...
+    = ldate(entry.display_start_date, format: :short) if entry.display_start_date
     - if entry.schedule_parent_id
       i.fa.fa-link
   td[
     class="#{classes} step-date step-end-date"
     data-id=entry.id
     data-precedent-id=entry.schedule_parent_id
     data-dependent-step-start-date=ldate(entry.display_end_date)]
-    = entry.display_end_date ? ldate(entry.display_end_date, format: :short) : ""
+    = ldate(entry.display_end_date, format: :short) if entry.display_end_date
```
hooverlunch: The above `if`s are not necessary -- that's what `ldate` does. If the value is nil, it returns empty string.

When you're refactoring, it's good to think through the logic of the preexisting code. Don't assume it's correct or necessary! You might be able to clean it up even more :)

app/views/admin/project_steps/_show.html.slim
```diff
@@ -77,15 +77,15 @@
...
+                  => ldate(step.old_start_date)
                   = "(#{step.old_start_date_statement})"
 
           - if step.completed?
             tr
               th = t("project_step.labels.completed")
               td style="background-color: #{step.background_color}"
                 p.tight-block
-                  => step.actual_end_date ? l(step.actual_end_date) : ''
+                  => ldate(step.actual_end_date) if step.actual_end_date
```
hooverlunch: `if` not necessary here, see below.

app/views/admin/project_steps/_show.html.slim
```diff
@@ -65,7 +65,7 @@
             th
               = t("project_step.labels.start_date")
             td style="background: #{step.scheduled_bg}"
-              = step.display_start_date ? l(step.display_start_date) : ''
+              = ldate(step.display_start_date) if step.display_start_date
```
hooverlunch: `if` not necessary here, see below.

app/views/loans/index.html.slim
```diff
@@ -42,7 +42,7 @@
                 = loan.location
               .sign_date
                 i.fa.fa-fw.fa-calendar>
-                = loan.SigningDate ? l(loan.SigningDate, format: :long) : ''
+                = ldate(loan.SigningDate, format: :long) if loan.SigningDate
```
hooverlunch: No need for `if` here either.

app/models/loan.rb
arisah-ijeoma: 👍

emitche:
I started doing QA on this accidentally and merged in `develop`. I'll do QA later today.

PR #412, opened 2017-11-28
6853 bug fix events missing in dashboard calendar

emitche:
- Bug fix for a Full Calendar bug
- Load all calendar events in dashboard calendar
- Fixes events not showing if they began before the displayed week


emitche:
Thanks, @hooverlunch.

emitche:
@hooverlunch You are correct about events that span entire months. I noticed something similar and logged a separate bug for that: https://redmine.sassafras.coop/issues/7547. But they might be the same problem. Thank you!

I'll pair with either you or Fuzzy on this. Thanks!

PR #413, opened 2017-11-29
7466 add fields to basic project

emitche:
- Add summary and details fields to Basic Project form
- Use same form partial for Loans and Basic Projects


spec/models/basic_project_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+describe BasicProject, type: :model do
```
emitche: Do I need to add any additional tests for BasicProject?

app/controllers/admin/basic_projects_controller.rb
```diff
@@ -95,7 +95,7 @@ def prep_form_vars
   end
 
   def basic_project_params
-    params.require(:basic_project).permit(:division_id, :length_months, :name, :primary_agent_id,
-      :secondary_agent_id, :signing_date, :status_value)
+    params.require(:basic_project).permit([:division_id, :length_months, :name, :primary_agent_id,
+      :secondary_agent_id, :signing_date, :status_value] + translation_params(:summary, :details))
```
hooverlunch: Nice modification here.

app/views/admin/projects/_translatable.html.slim
```diff
@@ -17,5 +17,7 @@
 
 javascript:
   $(function() {
-    new MS.Views.TranslationsView({el: $('[data-content-translatable="loan"]')});
+    new MS.Views.TranslationsView({
+      el: $("[data-content-translatable='#{project.class.name.underscore}']")
```
hooverlunch: Very nice code reuse!

spec/models/basic_project_spec.rb
hooverlunch: No this is fine, no major logic additions in this issue. Nice to have a basic projects spec file now!

linesarefuzzy:
Looks good. We may want to add `simple_format` to the details display on both loans and basic projects.

emitche:
@linesarefuzzy Interesting. That does look useful.

PR #414, opened 2017-11-29
6881 fix story scope

arisah-ijeoma:
I need assistance with the specs


emitche:
@hooverlunch, you would be best to help with fixing the tests.

arisah-ijeoma:
I intended to write a feature spec but that didn't work out due to some js thingies

emitche:
The attachments checkbox is nested inside loan type, but it should not be. The result is when the item validation shows, it shows in a weird place. image

emitche:
Failed test is unrelated to this pull request.

PR #417, opened 2017-12-06
7293 update rails

arisah-ijeoma:


app/views/admin/basic_projects/_details.slim
```diff
@@ -11,12 +11,7 @@ section.details class=(@basic_project.valid? ? "show-view" : " edit-view")
         a href=admin_basic_project_path(@person) = t("cancel_edit")
     - if policy(@basic_project).destroy?
       = link_to [:admin, @basic_project], method: :delete,
-          data: { \
-            confirm: t("basic_project.confirm_deletion"),
-            "confirm-title": t("basic_project.delete"),
```
emitche: What happens to the modal titles and modal custom button text with these changes? It looks like those customizations are lost. Cancel and confirm button text are not needed necessarily, as across files, those look to be the same messages so they don't need to be defined in multiple places. But what about "confirm-title"?

emitche:
Thanks for this work, @arisah-ijeoma.

app/views/admin/basic_projects/_details.slim
arisah-ijeoma: the responsible gem `twitter-bootstrap-rails-confirm` is not compatible with rails 5 yet. This  also brought up the question of whether we needed fancy confirm boxes

emitche: Okay. Thanks!

emitche: It's more important to be secure than to have fancy confirm boxes. So these changes are fine.

emitche:
I've merged develop into this branch and run migrations. There are two failing tests on my local development setup when I run tests. I'll send them your way, @arisah-ijeoma. If you could look into fixing those tests, that would be very helpful. Thank you!

emitche:
The failure also happens in continuous integration tests.

arisah-ijeoma:
@emitche, I had just one spec and it is intermittently failing test. JS

@hooverlunch, seems our hack didn't work on this.

db/migrate/20171129145057_remove_default_from_duration.rb
```diff
@@ -1,4 +1,4 @@
-class RemoveDefaultFromDuration < ActiveRecord::Migration
+class RemoveDefaultFromDuration < ActiveRecord::Migration[4.2]
```
emitche: I don't think this was the cause for the failing test.

app/views/admin/basic_projects/_details.slim
emitche: This did end up causing a bug. So we'll look into it.

PR #418, opened 2017-12-07
6961 duplicate name exists error

arisah-ijeoma:
@linesarefuzzy, I fixed the query and transactions are being created on the Madeline system

There is a merge conflict with develop but I don't think I can handle it by myself


linesarefuzzy:
Conflicts with develop are a separate issue that @hooverlunch is working on. I just merged this into the release-2.0 branch.

PR #421, opened 2017-12-09
7617: Show scheduled start date in show step modal

emitche:


emitche:
TODO: Style date field

PR #422, opened 2017-12-11
7615 fix bug

arisah-ijeoma:


db/seeds.rb
```diff
@@ -26,8 +26,8 @@
 LoanQuestionSet.connection.schema_cache.clear!
 LoanQuestionSet.reset_column_information
 
-LoanQuestionSet.find_or_create_by(id: 2, division: Division.root,
```
emitche: Why is division removed here? Automated change?

arisah-ijeoma: I dropped and repopulated my db and got an error from the seeds file

After checking, there was no association between LoanQuestionSet and Division. Also, the seeds file is not always run which would make it harder to catch this

emitche: Ok. Thanks.

emitche:
Thanks for this work. I tested it. The delete step confirm box now respects the "cancel" from the timeline context. If you try to delete the step from the calendar, this does not work.

Also, I noticed that "Add Step" no longer works in both the timeline and the calendar. I can look into this further to find out when this was introduced.

arisah-ijeoma:
Ok...I'm not sure about the add part but the confirm part was caused by removing the plugin

emitche:
I'll file a different bug for the "Add Step"part.

So this just needs the delete cancel button to work in the calendar context when deleting a step. Then this can be reviewed again.

Thank you!

PR #424, opened 2017-12-11
7620 fix bug

arisah-ijeoma:


arisah-ijeoma:
In `config/locales/en/loan_questions.en.yml`, there are some values that have html tags but these are not being escaped

After going through `https://github.com/fnando/i18n-js/issues/310`, it was mentioned that this would not work in i18n-js

emitche:
For the bug mentioned, I have logged that bug as 7647.

emitche:
This is the bug that was mentioned earlier.

image

PR #427, opened 2017-12-11
7626 fix bug

arisah-ijeoma:


app/views/admin/notes/_show.html.slim
```diff
@@ -15,9 +15,8 @@ header.small
     - if policy(note).destroy?
       .item
         => link_to admin_note_path(note), method: :delete, class: 'delete-action',
-            data: { confirm: t('notes.confirm_deletion') }
-            remote: true
-          i.fa.fa-trash>
+            remote: true, data: { confirm: t('notes.confirm_deletion') }
+          i.fa.fa-trash
```
emitche: The > at the end of trash might be a useful thing. I'll check.

PR #429, opened 2017-12-13
7647 remove html tag from loan questions en.yml

arisah-ijeoma:


arisah-ijeoma:
See note in `https://redmine.sassafras.coop/issues/7647`

PR #431, opened 2017-12-13
7610 make field show as read-only

arisah-ijeoma:


app/views/admin/project_steps/_modal_content.html.slim
```diff
@@ -104,13 +104,9 @@ section.step-fields class=(@mode == :show_and_form ? 'show-view' : 'edit-view')
 
       - if @step.old_duration_days && @step.old_duration_days > 0 && @step.is_finalized?
         = f.input :old_duration_days
-          - if policy(@step).edit_finalized_dates?
```
emitche: @linesarefuzzy Are certain individuals supposed to be able to edit the step's original duration?

emitche: We need to keep this part in.

PR #434, opened 2017-12-15
7505 txn created with error from qb

hooverlunch:


emitche:
I came across a bug while testing this in my environment. Please see: https://redmine.sassafras.coop/issues/7505#note-12.

app/models/accounting/quickbooks/updater.rb
```diff
@@ -1,23 +1,22 @@
...
-    class FullSyncRequiredError < StandardError; end
-    class NotConnectedError < StandardError; end
 
     # Reponsible for grabbing only the updates that have happened in quickbooks
     # since the last time this class was run. If no quickbooks data exists in the sytem
     # yet, FullFetcher will need to be run first.
     class Updater
       attr_reader :qb_connection
 
+      MIN_TIME_BETWEEN_UPDATES = 5.seconds
```
arisah-ijeoma: I thought this would be 15 seconds

hooverlunch: Yes, I reconsidered. There is a comment about this elsewhere in the file I think.

emitche:
Even with a fresh database copy, I still am unable to load transactions. Going to the settings page and running a full sync gives me the same error. More info in Redmine. Thanks.

emitche:
I believe the code is ready for this branch. The problem is a Rails bug that blocks this branch. See https://redmine.sassafras.coop/issues/7684.

emitche:
I can't really QA this, so I think this is ready, given that tests are passing.

PR #437, opened 2017-12-18
6715: Update sync messages

meltheadorable:


config/locales/en/en.yml
```diff
@@ -329,7 +329,7 @@ en:
   quickbooks:
     authorization_failure: "You must be connected to QuickBooks before you can access transactions. Please go to %{settings} to connect."
     class_location_notice_html: "<b>NOTE:</b> Before you can add transactions, you need to make sure <b>Classes</b> and <b>Locations</b> (a.k.a. Divisions) are enabled in your QuickBooks company settings."
-    full_sync_required: "A full sync is required before you can access transactions. Please go to %{settings} to sync."
+    full_reset_required: "A full reset is required before you can access transactions. Please go to %{settings} to sync."
```
hooverlunch: How about `A Quickbooks data reset is required before you can access transactions. Please go to %{settings} to perform a reset.`

app/models/accounting/quickbooks/updater.rb
```diff
@@ -102,7 +102,7 @@ def extract_qb_data(txn)
 
       def changes
         unless last_updated_at && last_updated_at > max_updated_at
-          raise FullSyncRequiredError, "Last update was more than 30 days ago, please do a full sync"
+          raise FullSyncRequiredError, "Last update was more than 30 days ago, please do a full reset"
```
hooverlunch: Could you please move this into I18n? And perhaps reword to `please reset your Quickbooks data`.

emitche:
Please use `QuickBooks` for the spelling, if there are text changes. Thank you!

hooverlunch:
@emitche Right on. I would love to standardize that. I never know which to use! Maybe `Quickbooks` in code but "QuickBooks" in user-facing text?

emitche:
All user facing text, including error messages, and all documentation should use the QuickBooks spelling. It's not a big deal though, as a search and replace is all I need to do to get things in sync.

linesarefuzzy:
QuickBooks is the proper spelling for user-facing stuff, but the Modules from quickbooks-ruby and the ones we created are spelled Quickbooks. So a global replace operation would have to be careful to only change instances in user-facing strings.

emitche:
Yes, @linesarefuzzy, that's what I meant. When I go to do translations, I would check for those strings inside the yml files and replace as needed. I did that on one of our previous releases. I meant client facing text only.

emitche:
Ready to be merged into release-2.0. If going to be merged into develop, develop just needs to be merged into this first.

PR #438, opened 2017-12-19
7673 double transactions

linesarefuzzy:


linesarefuzzy:
I don't understand this Travis failure but all tests pass on my machine.

PR #439, opened 2017-12-19
7504 wipe on full sync

meltheadorable:


app/models/accounting/quickbooks/full_fetcher.rb
```diff
@@ -10,20 +10,56 @@ def initialize(qb_connection = Division.root.qb_connection)
...
+        restore_all_accounts!(divisions_accounts)
       end
+
+      private
+
+      # Set all divisions' accounts to nil and return a hash of the QB ids of the removed accounts
+      # by division
+      def clear_all_accounts!
+        divisions_accounts = {}
+        Division.all.each do |d|
```
hooverlunch: I think we need to keep moving toward doing all QB stuff separately for different divisions. I think this class should accept a Division, not a qb connection, as its only param, and then when we are clearing out accounts like this, it should only do it for the current division. Existing instantiations of this class should explicitly pass the root division for now. I'd rather not have it as a default parameter, that seems odd to me.

Mel, could you make this change without too much trouble?

@linesarefuzzy would you agree?

app/models/accounting/quickbooks/full_fetcher.rb
```diff
@@ -10,20 +10,56 @@ def initialize(qb_connection = Division.root.qb_connection)
...
+          accounts_qb_ids = {
+            principal_qb_id: d.principal_account&.qb_id,
+            interest_receivable_qb_id: d.interest_receivable_account&.qb_id,
+            interest_income_qb_id: d.interest_income_account&.qb_id,
+          }
+          d.update(
+            principal_account_id: nil,
+            interest_receivable_account_id: nil,
+            interest_income_account_id: nil,
```
hooverlunch: I'm wondering why we're not deleting the account objects instead of just nullifying the foreign keys... Is there any reason to keep them around? @linesarefuzzy thoughts?

app/models/accounting/quickbooks/full_fetcher.rb
```diff
@@ -10,20 +10,56 @@ def initialize(qb_connection = Division.root.qb_connection)
...
+      def restore_all_accounts!(divisions_accounts)
+        divisions_accounts.each do |did, qb_ids|
+          Division.find(did).update(
+            principal_account: Accounting::Account.find_by(qb_id: qb_ids[:principal_qb_id]),
+            interest_receivable_account: Accounting::Account.find_by(qb_id: qb_ids[:interest_receivable_qb_id]),
+            interest_income_account: Accounting::Account.find_by(qb_id: qb_ids[:interest_income_qb_id]),
+          )
+        end
+      end
+
```
hooverlunch: Shouldn't we be deleting transactions here too? I realize the spec didn't say one way or the other but I think transactions and accounts should both be deleted. @linesarefuzzy can you confirm this?

spec/models/accounting/quickbooks/full_fetcher_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      instance_double(Quickbooks::Model::Account,
+        id: interest_receivable_account.qb_id,
+        name: interest_receivable_account.name,
+        classification: interest_receivable_account.qb_account_classification),
+      instance_double(Quickbooks::Model::Account,
+        id: interest_income_account.qb_id,
+        name: interest_income_account.name,
+        classification: interest_income_account.qb_account_classification)
+    ])
+  end
```
hooverlunch: Wow, a very lovely double!

spec/models/accounting/quickbooks/full_fetcher_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  let(:qb_transaction_service) { instance_double(Quickbooks::Service::JournalEntry, all: []) }
+  let(:account_fetcher) { Accounting::Quickbooks::AccountFetcher.new(qb_connection) }
+  let!(:account_fetcher_class) { class_double(Accounting::Quickbooks::AccountFetcher, new: account_fetcher).as_stubbed_const }
```
hooverlunch: Cool, never had to use this before!

spec/models/accounting/quickbooks/full_fetcher_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      expect(transaction_fetcher).to receive(:fetch).and_call_original
+
+      expect(qb_connection).to receive :update_attribute
+
+      subject.fetch_all
+      division = Division.root
+
+      new_account_ids = division.accounts.map(&:id)
+
+      # Accounts should be restored with the same QB ids but they should have different DB ids
```
hooverlunch: Nice comment!

spec/models/division_spec.rb
```diff
@@ -47,4 +47,8 @@
     root_division
     expect { create(:division, parent: nil) }.to raise_error(ActiveRecord::RecordInvalid)
   end
+
+  describe '#' do
+
+  end
```
hooverlunch: Delete?

meltheadorable:
I should note that fuzzy did the implementation itself, I picked things up just to implement the test suite, so I can't  really  answer your questions about the feature. I do have to assume that Fuzzy coded it as it was intended, but you and he should discuss the details and let me/him know what changes to make (I'm available to do the work, but don't need to be the one who owns it)

app/models/accounting/quickbooks/full_fetcher.rb
linesarefuzzy: Oh gosh yes I agree. Good catch. This was my branch, so I can fix it.

app/models/accounting/quickbooks/full_fetcher.rb
linesarefuzzy: They’re all about to get deleted anyway. Why delete them in two places?

app/models/accounting/quickbooks/full_fetcher.rb
linesarefuzzy: They all get deleted in the fetch_all method above. (I think I accidentally committed that to the release branch so it’s not shown as a change here.)

app/models/accounting/quickbooks/full_fetcher.rb
hooverlunch: Ok I didn't realize that, we cool.

PR #440, opened 2017-12-19
6714: Reload Settings when QB Connected

meltheadorable:


app/assets/javascripts/backbone/views/settings_view.coffee
```diff
@@ -0,0 +1,11 @@
...
+  initialize: ->
+    new MS.Views.AutoLoadingIndicatorView()
+    @setUpEventListener();
+
+  setUpEventListener: ->
+    $(window).one("focus", @checkQuickbooksConnected);
+
+  checkQuickbooksConnected: (e) ->
+    $.get '/admin/accounting/quickbooks/connected', (connected) =>
+      location.reload() if(connected)
```
hooverlunch: Parens not needed after `if`

app/assets/javascripts/backbone/views/settings_view.coffee
```diff
@@ -0,0 +1,11 @@
+class MS.Views.SettingsView extends Backbone.View
+  initialize: ->
+    new MS.Views.AutoLoadingIndicatorView()
+    @setUpEventListener();
+
+  setUpEventListener: ->
+    $(window).one("focus", @checkQuickbooksConnected);
```
hooverlunch: Cool, didn't know about `one`

hooverlunch: So I'm assuming the `focus` event here is being used to detect the user switching back to the Madeline browser window after closing the oauth flow?

But what if the user loads the settings page, checks their Gmail, and comes back. The page will reload needlessly won't it?

Couldn't we only setup the event handler once the connect to quickbooks button is clicked? It's still not perfect but perhaps a bit better?

hooverlunch: Also can we get a comment here explaining what is going on? It took me awhile to understand it. I had almost finished typing this to you Mel:

"In case you’re not busy in your meeting (ignore if not) -- I can’t figure out what the point of https://github.com/sassafrastech/madeline/pull/440/files is. The way I read it, the `connected` request is made immediately on page load? If so, why is it necessary? Perhaps I’m not understanding the use of window focus here" <--- AHA moment

meltheadorable: I thought about trying to hook into that click event in order to set up the listener, but I didn't want to overcomplicate things by getting too clever. What I think we can do is make setting up the listener in that initializer conditional on a parameter passed into the view instantiation that says whether or not QB is connected. 

It amounts to basically the same thing (if somebody opens the settings page when QB is already connected, there won't be a reload when it comes back) but we don't have to worry about responding to other event triggers or relying on the QB button not to change.

hooverlunch: I guess I don't see what is especially clever about watching for clicks on the QB button. The solution you propose would also not help in the situation where QB is not connected and the user does what I described earlier (checking their email), right?

Is there anything other than just listening for click on `.intuitPlatformConnectButton` that would need to happen? Can we try it?

meltheadorable: If QB is not connected, and they don't click the button, refocus will call the HTTP method to check, but the page will not reload -- that's what the `if` statement is checking for, `connected` will be false. The page only reloads if the connection status changes from disconnected to connected.

hooverlunch: Ah of course. This is fine with me then. Thanks.

hooverlunch: Would still like those comments though PS! :)

PR #442, opened 2018-01-08
5696 add test for division validation

arisah-ijeoma:


arisah-ijeoma:
Turns out the 5696 story is already done

PR #443, opened 2018-01-08
add simple format

arisah-ijeoma:


app/models/translation.rb
```diff
@@ -28,4 +28,12 @@ def blank?
   def to_s
     text
   end
+
+  def empty?
```
emitche: What are these two methods for?

arisah-ijeoma: they were required for simple format to work in the translatable modal pretty much like how the other 2 methods are used

emitche: Ok. Thanks.

PR #446, opened 2018-01-10
6459 disable link if logs not available on steps

arisah-ijeoma:


app/assets/javascripts/backbone/views/timeline_table_view.coffee
```diff
@@ -111,6 +111,13 @@ class MS.Views.TimelineTableView extends Backbone.View
     @openMenu(e, 'group')
 
   openStepMenu: (e) ->
+    $logsLink = @$('#project-step-menu a[data-action="view-logs"]').closest('li')
+
+    if !(@$(e.currentTarget).closest('.step-menu-col').hasClass('logs-available'))
```
emitche: Instead of disabling the logs link, could we do hide/show instead.

I know the issue description was not clear. But yes, we want to completely hide or show instead. I think show/hide for jQuery will do it. There are other examples where we use show/hide. Thank you!

arisah-ijeoma: I'll fix that right away

emitche: Thank you!

PR #447, opened 2018-01-10
6392 batch action bug

arisah-ijeoma:


app/assets/javascripts/backbone/views/timeline_select_steps_view.coffee
```diff
@@ -1,5 +1,9 @@
 class MS.Views.TimelineSelectStepsView extends Backbone.View
 
+  initialize: ->
+    # on page load, disable the batch actions since no step would have been checked
+    $('.dropdown-menu.batch-actions > li').addClass("disabled");
```
emitche: Smart.

app/assets/javascripts/backbone/views/timeline_select_steps_view.coffee
```diff
@@ -57,3 +61,9 @@ class MS.Views.TimelineSelectStepsView extends Backbone.View
         ids.push($(@).data('id'))
 
     @$('.step-ids').attr('value', ids)
+
+    # disable/enable the batch actions depending on if any step is checked
+    if ids.length == 0
```
emitche: :) I like this approach.

app/controllers/admin/project_steps_controller.rb
```diff
@@ -123,6 +123,9 @@ def duplicate
   end
 
   def batch_destroy
+    # don't initialize batch destroy if there are no step ids
```
emitche: The UI should catch any cases when there are no steps selected. Are the changes in this file just a safeguard?

emitche:
Looks good. I just have one question.

app/controllers/admin/project_steps_controller.rb
arisah-ijeoma: Yep
The UI would catch those due to the frontend manipulation but this safeguards our backend and makes sure no dirty data gets in, especially the wrong way

PR #448, opened 2018-01-10
6853: Fix calendar events not showing up in calendar

emitche:


app/models/calendar_event.rb
```diff
@@ -56,9 +56,6 @@ def self.filtered_events(date_range: nil, project_filter: nil, project_scope: Pr
     events = project_events_by_date_project_scope(date_range, project_scope.where(project_filter))
     events += step_events_by_date_project_filter(date_range: date_range, project_filter: project_filter,
       scope: step_scope, project_id: project_id)
-
```
emitche: This was hiding events that spanned across more than the range.

app/models/calendar_event.rb
```diff
@@ -87,11 +84,11 @@ def self.project_date_filter(range, scope = Project)
   end
 
   def self.project_step_date_filter(range, scope = ProjectStep, project_id)
-    scoped_steps = scope.where("actual_end_date BETWEEN :first AND :last "\
```
emitche: We changed which events we want to show up. Any events that occur over the range, even if they start before or end after the range, should be displayed.

emitche:
Thank you!

PR #452, opened 2018-01-24
7703 invalid interest bug

arisah-ijeoma:


emitche:
The error message works as expected when adding a transaction. However, after not being able to add the transaction and then reloading the transactions page, the error message remains on transactions page. Here's an example:

image

emitche:
Additionally, a page where I've not added new transactions, but that the system detects something is wrong with the balances will also display the same error message on the top of the page.

emitche:
Needs a QA re-review, not a code re-review.

emitche:
"The error message works as expected when adding a transaction. However, after not being able to add the transaction and then reloading the transactions page, the error message remains on transactions page."

Fixed

emitche:
"a page where I've not added new transactions, but that the system detects something is wrong with the balances will also display the same error message on the top of the page."

This still is not fixed, but I think this is a separate bug introduced by this feature. I will log a separate ticket.

image

emitche:
We are unlikely to encounter the above mentioned bug, as Madeline Accounting is not yet being used in production. The staging data is just test data. So fixing this bug does not need to block the overall bug that 7703 fixes. I'll add this information into the backlog so we have a record.

Thank you!

PR #454, opened 2018-01-24
5837 update show txn modal

arisah-ijeoma:


app/views/admin/accounting/transactions/_form.slim
```diff
@@ -44,3 +46,23 @@
   = f.input :private_note
     .view-element = simple_format(@transaction.private_note)
     = f.input_field :private_note, as: :text
+
+  .view-element.form-group
```
emitche: We need to use the f.input :some_field formatting here. The reason is that the form is using the label names as the field names, which is not what we want.

```
= f.input :change_in_interest
     .view-element = format_currency(@transaction.change_in_interest, @transaction.currency)
```

I notice that the labels are not capitalized as they are in the en.yml file, so maybe the translations for the labels are not being properly picked up. This occurs both with the way you've set up the form below, @arisah-ijeoma, and with the way I suggest setting up the form in this comment.

Please find a way to add the labels with capitalization. I believe it means that in addition to the items being inside activerecord in the en.yml file, the labels also need to be in the simple_form>labels>accounting_transaction section as well.

emitche: @arisah-ijeoma These fields are not editable and are not inputs when using the .view-element class as in the example above. Also, each of these items (change in interest, change in principal, interest balance, principal balance, and total balance) are all methods on the transaction. Therefore no custom code is needed.

Additionally, the corresponding labels inside activerecord.attributes.accounting.transaction need to be copied over to simple_form.labels.accounting_transaction.

If this is not clear, please let me know.

arisah-ijeoma:
screen shot 2018-01-25 at 2 27 48 pm

This was the issue I had as well

PR #458, opened 2018-01-26
7899 currency not showing

arisah-ijeoma:


emitche:
Polish:
Remove the extra space from the transactions.

Testing this inside the browser, there is a space between the currency and the amount. For US $, I am expecting `US $0.00`.

For other currencies often used in the system, Argentina is displayed as `AR $0.00` in the loan details. Nicaragua is displayed as `NI $0.00` at the moment in loan details. So we should do something similar in the transactions table.

This is what is currently in transactions:
image

emitche:
I think #459 fixes the above space issue.

emitche:
Merging develop into this branch will fix the spacing concern noted above.

app/models/accounting/quickbooks/updater.rb
```diff
@@ -111,6 +113,16 @@ def extract_qb_data(txn)
         txn.save!
       end
 
+      def lookup_currency(txn)
+        project = txn.project
+
+        if txn.quickbooks_data && txn.quickbooks_data[:currency_ref]
```
linesarefuzzy: I think you can do `.try(:[], :currency_ref)` here

app/models/accounting/quickbooks/updater.rb
```diff
@@ -111,6 +113,16 @@ def extract_qb_data(txn)
...
+          Currency.find_by(code: quickbooks_data[:currency_ref][:value]).try(:id)
+        elsif project
+          Currency.find_by(id: project.currency_id)
```
linesarefuzzy: Can't you just use `project.currency` here?

app/models/accounting/quickbooks/updater.rb
```diff
@@ -111,6 +113,16 @@ def extract_qb_data(txn)
         txn.save!
       end
 
+      def lookup_currency(txn)
+        project = txn.project
```
linesarefuzzy: This variable seems unnecessary to me

app/models/accounting/quickbooks/updater.rb
arisah-ijeoma: I could :)
thanks

app/models/accounting/quickbooks/updater.rb
arisah-ijeoma: I just felt weird calling `txn.project` in two different places

PR #460, opened 2018-01-29
5938 fix filtering:

arisah-ijeoma:
Wice grid filters by db column. Having either primary_agent or secondary_agent will cause the agents tovbe filtered by that one column so the filter is incomplete
According to https://github.com/leikind/wice_grid/issues/216, the solution would be to create another column which I think is not necessary


emitche:
I checked in with @linesarefuzzy about this feature. We've decided to de-prioritize this.

Thanks for your work so far, @arisah-ijeoma.

emitche:
Please keep this branch alive fro when we return to this work.

PR #462, opened 2018-01-30
7916 fix log bug

arisah-ijeoma:


app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -19,7 +19,7 @@
           i.fa.fa-fw.fa-calendar
         = @step.summary
 
-  .logs-index.post-list data-expandable="step-logs-#{@step.try(:id)}"
+  .logs-index.post-list data-expandable=(@context == 'calendar' ? "step-logs-#{@step.id}" : "")
```
emitche: Why did you add a context here? Could we do something like this instead?

`data-expandable=([Does @step.id exist] ? "step-logs-#{@step.id}" : "")`

The hope is the embedded logs list will function the same from inside a calendar and from a step, so adding contexts doesn't feel like the right approach.

Thanks.

emitche: The functionality is all working correctly.

I did not understand from a code perspective why we ask if the context is calendar because the timeline context can also have an expandable logs list. But the code is working correctly.

The reason this works correctly is that the step modal from the timeline context has a context of "calendar" when it loads its logs list. So that's why the functionality works.

emitche:
Functionality and QA review completed.

PR #463, opened 2018-01-30
7910 fix Nicaraguan currency symbol

arisah-ijeoma:
https://github.com/sassafrastech/madeline/pull/458/files needs to get merged in first for this to be fully tested


app/helpers/admin_helper.rb
```diff
@@ -54,7 +54,16 @@ def admin_custom_colors
   end
 
   def format_currency(amount, currency, tooltip: true)
-    display_symbol = currency ? currency.symbol.sub('$', ' $') : '' # add space before $ (pretty)
+    # add space before $ and C$ (pretty)
```
meltheadorable: I think I would probably remove the changes to this function and just add the space in the DB, since that reduces the amount of special-case formatting we have to do in this function.

meltheadorable: Otherwise this looks good!

arisah-ijeoma:
I was just about to get on it
I'll fix up the seeds file

Thanks for the find @meltheadorable

PR #469, opened 2018-02-12
3690 fix deploy url

emitche:
- Update Capistrano to the latest version
- Update repo_url to use https syntax for deployment

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -145,7 +145,7 @@ group :development do
 
   # Deployment
   gem 'capistrano3-delayed-job', '~> 1.0'
-  gem 'capistrano',  '~> 3.1'
+  gem 'capistrano',  '~> 3.10.1'
   gem 'capistrano-rails', '~> 1.1'
   gem 'capistrano-passenger'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -85,7 +85,7 @@ GEM
       activesupport (>= 3.0.0)
       uniform_notifier (~> 1.10.0)
     byebug (9.1.0)
-    capistrano (3.10.0)
+    capistrano (3.10.1)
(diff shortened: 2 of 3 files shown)


emitche:
In each user's own `.git/config` file, the url can be set to:
`url = git@github.com:sassafrastech/madeline.git`

This allows easier updates to the repo.

But for deployment, it appears the https style works best.

Gemfile
```diff
@@ -145,7 +145,7 @@ group :development do
 
   # Deployment
   gem 'capistrano3-delayed-job', '~> 1.0'
-  gem 'capistrano',  '~> 3.1'
+  gem 'capistrano',  '~> 3.10.1'
```
meltheadorable: This change shouldn't have been necessary to update to the latest version. Using `~>` with a two-digit version will always update to the latest minor version, in practice that means we would update to a `3.11` or `3.12` when they land, instead of being stuck on `3.10.x` until we manually change this version string again.

linesarefuzzy: I agree, I think `~> 3.10` is better

emitche:
"Perhaps the time on this fits better in 3733 than 3690. Or maybe it should have its own ticket."

Yes, 3733 is better, but I will give this it's own ticket. Thank you!

emitche:
@meltheadorable wrote: "One small comment, which isn't a really big deal. Using the https url is a good change in general, but I would still like us to figure out why key authentication is no longer working for you in your environment, since this really shouldn't have broken."

I agree.

I'm leaning towards holding off on trying to get these changes into sprint 2.3 in place of figuring out what happened altogether for this change to be needed.

emitche:
This is [ticket 7998](https://redmine.sassafras.coop/issues/7998) in Redmine

PR #470, opened 2018-02-12
7796 refactor layouts

hooverlunch:


hooverlunch:
Hi folks. A couple notes:

* I ended up going with separate JS and SCSS manifests for the admin and public areas since folks who are using only the public site shouldn't need to download the large amount of JS and SCSS needed for the admin site. I was still able to make SCSS variables work properly whilst maintaining this approach. 
* JS and SCSS that is needed by both is contained in application.js/css, and in the `shared` directories.
* We renamed 'embedded' to 'public' throughout
* The public stuff has not been thoroughly tested since none of it is working enough to be loaded in the browser. So I did my best. As it is all being rehabilitated we can fix any errors introduced due to this refactor.
* I added comments to the key files in `app/views/layouts/` to explain the kind of confusing 'nested layout' pattern that is getting used there.

emitche:
Thanks for this work, @hooverlunch.

app/assets/stylesheets/application.scss
```diff
@@ -1,15 +1,7 @@
...
- * or vendor/assets/stylesheets of plugins, if any, can be referenced here using a relative path.
- *
- * You're free to add application-wide styles to this file and they'll appear at the top of the
- * compiled file, but it's generally better to create a new file per style scope.
- *
- */
+// These are styles required by all parts of the app, including admin and public.
+// This manifest does NOT pull in the admin and public directories.
 
+// TODO: Not sure if this variable will exist in the public or admin stylesheets.
```
hooverlunch: To clarify: I don't think this ever worked. Just adding a flag here.

app/assets/stylesheets/public/_general2.scss
```diff
@@ -1,3 +1,5 @@
+// TODO: This is an old file and should probably be removed, with styles moved to appropriate places.
+
```
hooverlunch: The vast majority of this file is commented out. I figure it should get cleaned up during the public site work.

config/initializers/devise.rb
```diff
@@ -265,9 +265,9 @@
...
-  Devise::SessionsController.layout "devise"
-  Devise::RegistrationsController.layout proc{ |controller| user_signed_in? ? "application" : "devise" }
-  Devise::ConfirmationsController.layout "devise"
-  Devise::UnlocksController.layout "devise"
-  Devise::PasswordsController.layout "devise"
+  Devise::SessionsController.layout "admin/devise"
+  Devise::RegistrationsController.layout proc{ |controller| user_signed_in? ? "admin/signed_in" : "admin/devise" }
+  Devise::ConfirmationsController.layout "admin/devise"
+  Devise::UnlocksController.layout "admin/devise"
+  Devise::PasswordsController.layout "admin/devise"
```
hooverlunch: Devise is coupled to the admin system currently which is why I put the devise template stuff under 'admin'. If we end up needing to use devise for the crowd funding platform, we can refactor.

hooverlunch:
Note: I suggest Liz and Fuzzy as reviewers since you both will want to see this I think. Not sure if Mel needs to see it (in the interest of time) since it's just getting rid of old stuff. But up to you all.

config/initializers/devise.rb
emitche: This sounds sensible. Thank you.

PR #471, opened 2018-02-13
6130 monthly interest accrual

arisah-ijeoma:


app/models/accounting/quickbooks/updater.rb
```diff
@@ -47,9 +47,16 @@ def update(loan = nil)
         # but that condition is measured in days, not seconds, so this small a difference shouldn't matter.
         qb_connection.update_attribute(:last_updated_at, Time.now)
 
-        if loan
-          update_ledger(loan)
-          InterestCalculator.new(loan).recalculate
+        if loans
+          # check if loan is one object or multiple
+          [loans] if loans.is_a? Loan
```
linesarefuzzy: Look closely at this line, @arisah-ijeoma . It's not doing anything. You need to add `loans = ` at the beginning. 

Please also add to the comment at the top of this method noting that its argument can be either a single loan or a collection or array of loans.

app/models/loan.rb
```diff
@@ -63,6 +63,7 @@ class Loan < Project
   }
   scope :status, ->(status) { where(status: status) }
   scope :visible, -> { where.not(publicity_status: 'hidden') }
+  scope :active, -> { where(status_value: 'active') }
```
linesarefuzzy: Nice! :)

config/schedule.rb
```diff
@@ -1,3 +1,8 @@
 every 1.day, at: '3:00' do
-  runner "RecalculateAllLoanHealthJob.perform_later"
+  runner 'RecalculateAllLoanHealthJob.perform_later'
+end
+
+# runs job at midnight
+every '0 0 1 * *' do
```
linesarefuzzy: Isn't this equivalent to `every 1.month`? That would be much easier to read.

emitche:
Assigned to @meltheadorable for code review and @arisah-ijeoma for edits.

app/jobs/monthly_interest_accrual_job.rb
```diff
@@ -0,0 +1,5 @@
+class MonthlyInterestAccrualJob < ApplicationJob
+  def perform
+    Accounting::Updater.new.update(Loan.active)
```
meltheadorable: Nice.

app/models/accounting/interest_calculator.rb
```diff
@@ -60,75 +60,101 @@ def initialize(loan)
...
-        # set the needs_qb_push flag. We ignore changes to the transaction's balances since these
-        # don't need to get pushed. Therefore we don't need to check changes on the transaction
-        # object itself, just the line items.
-        #
-        # Note that if the flag is already set we leave it as true even if no changes
-        # occurred. The transaction may have changed earlier (e.g. via the UI) and may need a push
-        # even if we don't change anything here.
-        tx.needs_qb_push = tx.needs_qb_push || tx.line_items.any?(&:type_or_amt_changed?)
+      txn_dates = txns_by_date.keys
+      month_boundaries = month_boundaries(first_date, last_date)
```
meltheadorable: I might rename this variable, I try to avoid using temporary variable names that are the same as method names because it's harder when you start trying to refactor things (find/replace will catch false positives)

app/models/accounting/quickbooks/updater.rb
meltheadorable: I'd also check why the test suite didn't catch this, if we're stubbing too much of this updater or not testing both types of input it would be nice to have a `TODO` to find some way to stub less of this workflow/add more unit tests.

spec/models/accounting/interest_calculator_spec.rb
```diff
@@ -3,7 +3,7 @@
 # See docs/example_calculation.xlsx for ground truth used to build this spec.
 describe Accounting::InterestCalculator do
   let!(:division) { create(:division, :with_accounts) }
-  let(:loan) { create(:loan, division: division, rate: 8.0) }
+  let(:loan) { create(:loan, division: division, rate: 8.0, status_value: 'active') }
```
meltheadorable: Is there a way to handle this and any other related properties that might need to be set as a trait on the factory? It seems like this is something we should be using in other tests too.

spec/models/accounting/interest_calculator_spec.rb
```diff
@@ -75,43 +75,43 @@
         expect(t3.line_items.size).to eq(2)
 
         # account details
-        expect(t3.line_item_for(int_rcv_acct).amount).to equal_money(2.24)
+        expect(t3.line_item_for(int_rcv_acct).amount).to equal_money (0.08/365 * 27 * 117.50).round(2)
```
meltheadorable: I'm generally suspicious of using too many "magic numbers"  by themselves -- would there be a way to define these as constants so that it's easier for somebody coming into this later to see what the components going into this calculation are and why they are being done this way?

app/models/accounting/quickbooks/updater.rb
arisah-ijeoma: ah yes

spec/models/accounting/interest_calculator_spec.rb
arisah-ijeoma: left you a note on slack

emitche:
@arisah-ijeoma Is this ready to be QAed? Do you need to do more work on this?

emitche:
@meltheadorable Does this need additional code review or is it ready for QA? Thank you!

emitche:
Thank you, @meltheadorable!

PR #472, opened 2018-02-15
7999 tidy logic in timeline view

arisah-ijeoma:


app/views/admin/timeline/_step.html.slim
```diff
@@ -28,40 +28,46 @@
...
       = project_step_status(entry)
   - else
-    td class=classes data-id=entry.id
+    td class=classes data-id=entry.id data-action="show"
       = project_step_status(entry)
 
   - latest_logs = entry.latest_logs
   - logs_available = latest_logs.length
-  td.recent-logs class="#{classes} #{logs_available > 0 ? 'has-logs' : '' }" data-id=entry.id
+  td.recent-logs class="#{classes} #{logs_available > 0 ? 'has-logs' : ''}" data-id=entry.id data-action="show"
```
emitche: I like the approach of using data-action="show"

app/assets/javascripts/backbone/views/timeline_table_view.coffee
```diff
@@ -22,10 +22,10 @@ class MS.Views.TimelineTableView extends Backbone.View
     'click .timeline-action[data-action="new-step"]': 'newStep'
     'click #project-group-menu [data-action="add-child-group"]': 'newChildGroup'
     'click #project-group-menu [data-action="add-child-step"]': 'newChildStep'
-    'click #project-group-menu [data-action="edit"]': 'editGroup'
+    'click .project-groups[data-action="edit"]': 'editGroup'
     'confirm:complete #project-group-menu [data-action="delete"]': 'deleteGroup'
     'click .project-step': 'showStep'
```
emitche: This would no longer work given the new class name. It is also no longer needed.

app/views/admin/timeline/_step.html.slim
emitche: I noticed when there are logs you can't open the modal by clicking on the cell only. I will add this to the backlog for a fix.

PR #473, opened 2018-02-20
8014 step duration calendar bug

arisah-ijeoma:


app/models/calendar_event.rb
```diff
@@ -84,8 +84,10 @@ def self.project_date_filter(range, scope = Project)
   end
 
   def self.project_step_date_filter(range, scope = ProjectStep, project_id)
-    scoped_steps = scope.where("scheduled_start_date <= :last AND "\
-      "COALESCE(actual_end_date, scheduled_start_date + scheduled_duration_days) >= :first",
+    scoped_steps = scope.where(
+      "COALESCE(actual_end_date, scheduled_start_date + scheduled_duration_days) >= :first AND "\
+      "scheduled_start_date <= :last OR "\
+      "scheduled_start_date BETWEEN :first AND :last",
```
emitche: If we just used line 90 and not line 89, would it have the same effect?

arisah-ijeoma: I could test that out
before we moved around the scopes, one couldn't do without the other

emitche:
Functional review: Yes, the bug is fixed with these changes.

emitche:
I merged the most recent develop into this branch.

app/models/calendar_event.rb
emitche: Okay. Thanks.

arisah-ijeoma: @emitche, I've tested it and this OR statement is a necessity

PR #474, opened 2018-02-20
7701 fix blank currencies for fake data

arisah-ijeoma:


spec/factories/organizations.rb
```diff
@@ -45,7 +45,6 @@
 FactoryBot.define do
   factory :organization do
     division { root_division }
-    country
```
emitche: Why did you remove country from the factory?

arisah-ijeoma: it was responsible for adding new currencies since there's an association

emitche: Ok. Thanks.

PR #476, opened 2018-02-21
7933 update loan policy

arisah-ijeoma:


meltheadorable:
Changed to "needs revision" because of the open question in #479

emitche:
@meltheadorable Assigning to you to move this forward, if possible, and to make a decision in #479.

PR #478, opened 2018-02-26
7982 currency issue on txn

arisah-ijeoma:


arisah-ijeoma:
@linesarefuzzy and I are supposed to discuss that this afternoon if he's able to make it

linesarefuzzy:
Yeah, this is a little weird. Why do we need to set the currency again right after we just set it and saved in the extract method? If it needs to be done, please add a comment explaining why. Thanks!

arisah-ijeoma:
I moved `txn.currency = lookup_currency(txn)` from `extract_qb_data` to `update_ledger`

This is because `lookup_currency` is doing a check for whether the transaction is from QB or not. If left in `extract_qb_data`, it wouldn't utilize that condition to the maximum so I moved it out where it would append the currency whether it's from QB or not

@linesarefuzzy @meltheadorable

emitche:
@arisah-ijeoma Because this was marked as needs-minor-revision, it does not need to be reviewed again. Do you want someone to look it over again from the code standpoint or is this good to go? Thank you.

arisah-ijeoma:
oh..my bad @emitche 

I re-wrote some code and added some explanation on why I did that. That's why I used the `ready for review` label. Which would you suggest?

emitche:
Thank you!

emitche:
@linesarefuzzy @meltheadorable : Please see the earlier note from @arisah-ijeoma. This could use a look over for the most recent change she made before going to QA. Thank you!

PR #479, opened 2018-02-26
7802 add public flag to divisions

arisah-ijeoma:


emitche:
@meltheadorable Assigning this to you to make a decision and to move this forward. If you do need input from @linesarefuzzy, which may be the case, just assign it back to him. Thanks so much.

PR #481, opened 2018-02-27
8025 add class

arisah-ijeoma:
@emitche

I tried the clicks in an incognito browser and everything was fine. On the regular browser, for some reason it'd render `font` html tag inconsistently. This seems to be what is causing the breakage

I'll await your testing but this PR might just be closed

screen shot 2018-02-27 at 12 04 18 pm


emitche:
That's peculiar and unexpected in terms of it rendering `<font>` tags. Definitely a bug. That should not be happening.

Most people are using regular browsing and not incognito browsing.

emitche:
I'll look this over in the browser and get back with you.

emitche:
We weren't able to duplicate the bug during pairing, but I left a note in the Redmine ticket to look out for the <font> bug during the QA process.

PR #490, opened 2018-03-09
8119: Remove QuickBooks connection error message on production

emitche:
We are getting hundreds of emails because Madeline production does not have an active QuickBooks account connected to it. If QuickBooks is not connected on production, this does not indicate a failure. It could be an intentional choice (which it is in this instance). Please see Redmine for more details.

### DIFF
--- a/app/models/accounting/quickbooks/updater.rb
+++ b/app/models/accounting/quickbooks/updater.rb
@@ -27,6 +27,10 @@ module Accounting
       #
       # This argument can either be a single loan or an array of loans
       def update(loans = nil)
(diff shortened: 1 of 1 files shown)


linesarefuzzy:
I don't think it's a good idea to modify the Updater for this purpose, since it's used in other places. Instead, I think we should edit the monthly job so it only runs when there is a connection. I made a PR for a hotfix to do that: 
https://github.com/sassafrastech/madeline/pull/491

I also noticed that the cron job on the production server was set to run every minute for some reason rather than once a month. That's why we were getting so many error emails. I manually edited the crontab to put it back to 2am on the first of the month. So we're not getting the emails anymore.

PR #491, opened 2018-03-10
#8119 Bug fix: job attempting to run even when not connected to QB

linesarefuzzy:


emitche:
This can go into the release for 2.5.

PR #492, opened 2018-03-13
8115 fix currency in spanish

arisah-ijeoma:


app/helpers/admin_helper.rb
```diff
@@ -71,7 +71,9 @@ def format_currency(amount, currency, tooltip: true)
         class="currency_symbol" title="#{currency.try(:plural_name)}">#{display_symbol}</a>).html_safe
     end
 
-    return number_to_currency(amount, unit: display_symbol, negative_format: "(%u%n)")
+    # since we want to display all amounts with the symbols in front and dot notation
+    # we're adding locale: :en to set it to the English format by default
+    number_to_currency(amount, unit: display_symbol, negative_format: "(%u%n)", locale: :en)
```
emitche: Could we set the currency formatting inside the yml instead for both en and es locales? They would be the same.

arisah-ijeoma: I was thinking instead of adding some lines of code to the `en.yml` file and transifex would pick it up and create them in the `es.yml` file, I could use the default `en` notation here

Is there a reason for the preference on using the yaml files instead? @emitche

emitche: I think your preference would work. My preference was based on how we do yaml translations in general in the project.

PR #494, opened 2018-03-15
8000: fix wp template fetching

meltheadorable:


emitche:
This code looks good so far. Thank you!

lib/wordpress_template.rb
```diff
@@ -27,6 +27,7 @@ def self.process_template(division:, html:)
       '<%= yield :\k<section> %>'
     )
     additional_substitutions.each { |sub| html.gsub! *sub }
+    FileUtils.mkdir_p(file_path)
```
linesarefuzzy: Nice catch!

PR #495, opened 2018-03-15
7996 pull old txns from qb

arisah-ijeoma:


emitche:
@meltheadorable: @arisah-ijeoma wrote me that this can be moved to `needs-qa`, as it is good to go, but to check in with you first. Good to go to QA? Thank you!

emitche:
When trying to load transactions for a loan using fake test data and a sandbox QB account, I get this error:

````
NoMethodError at /admin/loans/7/transactions
undefined method `start_with?' for nil:NilClass
````

The error is in reference to line 90 of app/models/accounting/quickbooks/updater.rb.

`txn.managed = (txn.quickbooks_data['doc_number'].start_with? 'MS') ? true : false`

emitche:
After loading up a completely different branch, but with the same data, I get this page. So maybe there's something weird causing the bug on the 7996 branch.

image

emitche:
@arisah-ijeoma 

I am still seeing this bug when I load a loan. Do you think it's because I need to delete transactions in QuickBooks?

image

emitche:
This bug has been fixed:

`NoMethodError at /admin/loans/7/transactions
undefined method `start_with?' for nil:NilClass`

emitche:
Negative balance bug is on my end and not caused by work in 7996.

emitche:
I know for certain that this is working because it is pulling in old transactions.

PR #496, opened 2018-03-20
5431 member deletion bug

arisah-ijeoma:
the test I was going for was the text in the dialog box but it was slippery
the PR is ready for review tho


spec/models/person_spec.rb
```diff
@@ -41,14 +41,16 @@
 require 'rails_helper'
 
 describe Person, type: :model do
+  let(:person) { create(:person, :with_member_access, :with_password) }
+  let(:log) { create(:project_log) }
 
+  before { person.project_logs << log }
```
meltheadorable: You should be able to use `let(:log) { build(:project_log) }` (hopefully?) before you create the person, and then in the person, do `create(:person, :with_member_access, :with_password, project_logs: [log])` and avoid this `before` call (they work really funky and i make an effort to avoid them when possible.

emitche:
During QA, I found this bug. After deleting the member that had logs and trying to visit the logs page for a specific loan (the loan related to the log), the page errors out. See the error below. Additionally, the related step's modal cannot be opened from the timeline.

image

app/models/project_step.rb
```diff
@@ -385,7 +385,7 @@ def handle_old_duration_days_logic
 
     # By default, old_duration_days is set to 0.
     # Only remember old duration if a duration has been set other than the default and then changed.
-    unless old_duration_days > 0
+    unless old_duration_days && old_duration_days > 0
```
arisah-ijeoma: found another bug and fixed

emitche: Thanks!

emitche:
Okay. This can use some UX tweaks. But the functionality is here. Thank you! Thank you also for fixing the bug and the bug you found, @arisah-ijeoma.

emitche:
I found another bug while QAing this, which is a completely new bug. I'll log a ticket for it.

PR #497, opened 2018-03-21
6541 show project summary

arisah-ijeoma:


emitche:
If there is no summary present, the icon should not display at all. There would be no tooltip for those projects that have no summary.

emitche:
I'll make tweaks. They are tiny changes.

PR #500, opened 2018-03-27
8217 move extract_qb_data to its own class

arisah-ijeoma:
https://github.com/sassafrastech/madeline/pull/478 needs to be merged in first

PR #501, opened 2018-03-27
8004 determine txn type

arisah-ijeoma:


app/models/accounting/quickbooks/data_extractor.rb
```diff
@@ -1,13 +1,15 @@
...
-      attr_reader :txn
+      attr_reader :txn, :loan
 
       def initialize(txn)
         @txn = txn
+        @loan = txn.project
       end
 
       def extract!
+        # TODO: if 'other', managed: false
```
arisah-ijeoma: @meltheadorable, this is breaking at the moment because it requires the txn to have an account in the transaction model

app/models/accounting/quickbooks/data_extractor.rb
```diff
@@ -44,18 +45,53 @@ def extract!
...
+        line_items.each do |li|
+          line_item = txn.line_item_with_id(li['id'].to_i)
+          li_details[line_item.posting_type] = line_item.account
+        end
+
+        set_type(li_details['Debit'], li_details['Credit'], line_items)
+      end
+
+      def set_type(debit, credit, lis)
+        if (debit && debit.id == @int_rcv.id) && (credit && credit.id == @int_inc.id) && lis.count == 2
```
meltheadorable: I think our ruby version is new enough that you can use `debit&.id` and `credit&.id` to shortern this up a bit (it's called the "safe navigation operator" if you wanna look it up and read more about it, introduced in ruby 2.3)

app/models/accounting/quickbooks/data_extractor.rb
```diff
@@ -44,18 +45,53 @@ def extract!
...
+        @int_rcv = loan.division.interest_receivable_account
+        @int_inc = loan.division.interest_income_account
+        @prin_acct = loan.division.principal_account
+        li_details = {}
+
+        return 'other' if line_items.count > 3
+
+        line_items.each do |li|
+          line_item = txn.line_item_with_id(li['id'].to_i)
+          li_details[line_item.posting_type] = line_item.account
```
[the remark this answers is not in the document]
meltheadorable: I know that technically it shouldn't change the outcome of the result, because if something has extra line items it'll get classified as "other" anyway, but if we decide to change that later and reclassify transactions as the type they are behaving like (even if they include extra line items) we want the behavior to be correct.

app/models/accounting/quickbooks/data_extractor.rb
```diff
@@ -44,18 +45,53 @@ def extract!
...
-        txn.save!
+        # set transaction type
+        txn.loan_transaction_type_value = txn_type
+
+        txn.managed = false if txn.loan_transaction_type_value == 'other'
+
+        # TODO: set txn account
+        # the validation for account will hit here from the spec when testing managed
+        # so skipping for test env and doing validate: false in spec
+        txn.save! unless Rails.env.test?
```
meltheadorable: This shouldn't be needed, and we should figure out why it is/what to do about it. We should fix this in the test, or we should fix it in the code, or we should fix the validation on that condition so that it doesn't run for `other` type transactions, but I don't want to commit a hack like this if it can be avoided.

app/models/accounting/quickbooks/data_extractor.rb
```diff
@@ -44,19 +45,69 @@ def extract!
...
 
-        txn.save!
+        # set transaction type
+        txn.loan_transaction_type_value = txn_type
+
+        txn.managed = false if txn.loan_transaction_type_value == 'other'
+
+        # TODO: set txn account
+        # the validation for account will hit here from the spec when testing managed
+        # so skipping for test env and doing validate: false in spec
```
meltheadorable: This comment is a bit inaccurate now.

app/models/accounting/quickbooks/updater.rb
```diff
@@ -88,16 +87,7 @@ def extract_qb_data(txn)
 
         extractor = Accounting::Quickbooks::DataExtractor.new(txn)
         extractor.extract!
```
meltheadorable: Do the changes actually stick when you do this? I wouldn't expect the txn object to have actually been mutated with all the changes without saving it inside the extractor.

app/models/accounting/quickbooks/data_extractor.rb
arisah-ijeoma: will update

app/models/accounting/quickbooks/updater.rb
arisah-ijeoma: I'll write a spec to cover that

config/locales/en/en.yml
```diff
@@ -580,6 +580,10 @@ en:
     send_instructions: "Send instructions"
     waiting_confirmation: "Currently waiting confirmation for: %{email}"
 
+  member:
+    system_access: "System access allows a user to log into Madeline and view the loans and projects they have access to based on their division."
+    access_role: "Administrators have the ability to edit specific content not available to a regular member, such as managing divisions and editing loan questions."
```
linesarefuzzy: YAAASSSSSSS!!!

spec/models/accounting/quickbooks/data_extractor_spec.rb
```diff
@@ -158,5 +158,334 @@ def expect_line_item_amounts(amounts)
...
+          end
+        end
+
+        describe 'random with multiple debits and credits' do
+          it do
+            expect(txn.loan_transaction_type_value).to eq('other')
+            expect(txn.managed).to be false
+          end
+        end
+      end
```
linesarefuzzy: I think I get the spirit of what you're doing here, but you're actually just running the exact same test over and over again with different names. This only needs to be done once. Or if we need to test all these edge cases, they should be tested with different sets of line items.

spec/models/accounting/quickbooks/data_extractor_spec.rb
```diff
@@ -158,5 +158,334 @@ def expect_line_item_amounts(amounts)
...
+
+      describe 'too few line items' do
```
linesarefuzzy: This is actually an impossible scenario, since QB won't allow a txn to be saved if its credits and debits are not equal. So it's not possible to have a journal entry with only one line item. I would add another line item that debits a random account.

spec/models/accounting/quickbooks/data_extractor_spec.rb
```diff
@@ -158,5 +158,334 @@ def expect_line_item_amounts(amounts)
...
+            'total' => '12.30',
+            'private_note' => 'Random stuff' }
+        end
+
+        it do
+          expect(txn.loan_transaction_type_value).to eq('other')
+          expect(txn.managed).to be false
+        end
+      end
+    end
```
linesarefuzzy: These tests are good. At some point we might want to refactor this to have a `quickbooks_data_generator` method that takes arguments like `line_items: [{type: :credit, account: txn_acct, amount: 1.23}]`. That would clean this up a lot. But it doesn't have to be done now.

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -238,4 +238,13 @@
       end
     end
   end
+
+  describe 'extract_qb_data' do
+    let(:txn) { Accounting::Transaction.new(project: loan, managed: true, quickbooks_data: quickbooks_data,
```
linesarefuzzy: Why not use the factory here?

spec/models/accounting/quickbooks/updater_spec.rb
```diff
@@ -238,4 +238,13 @@
...
+      loan_transaction_type_value: nil, account: txn_acct) }
+
+    it 'data persists from the extractor to the updater' do
+      expect(subject.send(:extract_qb_data, txn).loan_transaction_type_value).to eq('repayment')
```
linesarefuzzy: Why repayment? Perhaps add a comment?

app/models/accounting/quickbooks/data_extractor.rb
```diff
@@ -44,19 +45,67 @@ def extract!
...
+        end
+      end
+
+      def any_debit_txns?(debits, acct)
+        debits.any? { |debit| debit&.id == acct.id }
+      end
+
+      def any_credit_txns?(credits, acct)
+        credits.any? { |credit| credit&.id == acct.id }
       end
```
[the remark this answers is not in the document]
arisah-ijeoma: thanks!

haven't used `in?` before

spec/models/accounting/quickbooks/data_extractor_spec.rb
arisah-ijeoma: I would like to pair on this at that time

emitche:
When 8217 (#500) is merged into develop, develop will become the base branch.

PR #504, opened 2018-03-29
4581 fix route

arisah-ijeoma:


app/controllers/admin/basic_projects_controller.rb
```diff
@@ -1,5 +1,4 @@
 class Admin::BasicProjectsController < Admin::ProjectsController
-  include TranslationSaveable
```
arisah-ijeoma: removed because it's already included in the parent class

app/controllers/admin/loans_controller.rb
```diff
@@ -1,5 +1,5 @@
 class Admin::LoansController < Admin::ProjectsController
-  include TransactionControllable, TranslationSaveable, QuestionnaireRenderable
```
arisah-ijeoma: removed because it's already included in the parent class

PR #505, opened 2018-03-31
8223 seeds refactor and loan public options

emitche:
- Update translations for loan public level options


emitche:
Let me know if the migration looks ok, if either of you have time on Monday. I tested this. Thank you!

db/migrate/20180331230509_add_loan_public_level_options.rb
```diff
@@ -0,0 +1,18 @@
+class AddLoanPublicLevelOptions < ActiveRecord::Migration[5.1]
+  def up
+    loan_public_level = OptionSet.find_or_create_by(division: Division.root, model_type: Loan.name,
+      model_attribute: 'public_level')
+    loan_public_level.options.destroy_all
```
meltheadorable: Why destroy instead of just updating the translations?

db/migrate/20180331230509_add_loan_public_level_options.rb
```diff
@@ -0,0 +1,18 @@
...
+      label_translations: {en: 'Featured', es: 'Destacado'})
+    loan_public_level.options.create(value: 'hidden', position: 2,
+      label_translations: {en: 'Hidden', es: 'Oculto'})
+  end
+
+  def down
+    loan_public_level = OptionSet.find_or_create_by(division: Division.root, model_type: Loan.name,
+      model_attribute: 'public_level')
+    loan_public_level.options.destroy_all
+    loan_public_level.destroy
```
meltheadorable: Since the options already exist, this reverse operation doesn't actually undo the above migration, it would also need to recreate them with the old TODO values. If you redo the `up` migration so that it changes the values only (instead of destroying them first), you should be able to just have an empty `down` migration, since there's no situation where you would really want to roll back changing the translations to be more correct.

emitche: Okay. I'll change it to an empty `down` function.

db/migrate/20180331230509_add_loan_public_level_options.rb
emitche: I followed other migrations for option set seeds from the past.

How would you approach the updates?

Maybe I can check in with you today about this, if you have capacity.

emitche:
Thanks for reviewing this, @meltheadorable!

emitche:
@meltheadorable: Does this work for the db/migrate file? Thank you!

emitche:
I tested these changes and they fail. I need to make updates.

emitche:
Okay. I'll need to pair on this approach. I'll likely move this to 2.8. Thank you!

emitche:
There are broken translations on the loans public option that this ticket fixes. I'd like to prioritize getting those translations fixed, even if just a temporary fix as we figure out the architecture pieces of this ticket.

meltheadorable:
An update on the current status of this would be great, as I'm not sure from this comment thread where things stand.

lib/option_set_creator.rb
```diff
@@ -14,200 +14,306 @@ def create_all
...
-    progress_metric.options.create(migration_id: 1, label_translations: {
-        en: I18n.t('database.option_sets.progress_metric.on_time', locale: 'en'),
-        es: I18n.t('database.option_sets.progress_metric.on_time', locale: 'es')
-      })
-    progress_metric.options.create(migration_id: 2, label_translations: {
-        en: I18n.t('database.option_sets.progress_metric.ahead', locale: 'en'),
-        es: I18n.t('database.option_sets.progress_metric.ahead', locale: 'es')
-      })
+
+    change_plan = progress_metric.options.find_or_create_by(migration_id: -3)
```
emitche: Are migration ids needed here? Could we search by value instead?

lib/option_set_creator.rb
```diff
@@ -14,200 +14,306 @@ def create_all
...
-    loan_transaction_type.options.create(value: 'repayment', position: 3, label_translations: {
-        en: I18n.t('database.option_sets.loan_transaction_type.repayment', locale: 'en'),
-        es: I18n.t('database.option_sets.loan_transaction_type.repayment', locale: 'es')
-      })
-    loan_transaction_type.options.create(value: 'other', position: 4, label_translations: {
-        en: 'Other',
-        es: 'Otro'
-      })
+
+    interest = loan_transaction_type.options.find_or_create_by(value: 'interest', position: 1)
```
emitche: Do you know why we use position in this method?

db/migrate/20180331230509_add_loan_public_level_options.rb
```diff
@@ -0,0 +1,24 @@
+class AddLoanPublicLevelOptions < ActiveRecord::Migration[5.1]
```
meltheadorable: I don't think we need these anymore, the options were added a while ago and are already on prod, unless I'm mistaken.

lib/option_set_creator.rb
```diff
@@ -14,200 +14,306 @@ def create_all
...
+    community_solar.save
   end
 
   def create_public_level
     public_level = OptionSet.find_or_create_by(division: Division.root, model_type: Loan.name,
       model_attribute: 'public_level')
-    public_level.options.destroy_all
-    public_level.options.create(value: 'public', label_translations: {
+
+    public = public_level.options.find_or_create_by(value: 'public')
```
meltheadorable: I think `public` is a reserved word, if this works I wouldn't guarantee it would continue to in the future. I'd feel better about naming this variable `public_option` or something.

lib/option_set_creator.rb
meltheadorable: I think because the options are intended to have an order other than alphabetical, so it's needed for sorting purposes, but I'd probably rather set those values only in the `find_or_create` and update the position with the translations...

emitche:
I can make these changes. Thanks, @meltheadorable.

db/migrate/20180331230509_add_loan_public_level_options.rb
emitche: Yes, they should already be on production.

emitche:
For QA: Just check that these items do actually update with some test strings.

PR #507, opened 2018-04-02
8227: Update LoanQuestion to Question in Legacy migration

emitche:


emitche:
This was merged but then changed back to the original.

PR #508, opened 2018-04-03
7799: update public frontend

meltheadorable:


app/controllers/public/loans_controller.rb
```diff
@@ -1,28 +1,34 @@
...
     # Set last loan list URL for 'Back to Loan List' link
     session[:loans_path] = request.fullpath
   end
 
   def show
-    @loan = Loan.status('all').find(params[:id])
+    @loan = Loan.find(params[:id])
     @pictures = @loan.featured_pictures(5) # for slideshow
-    @other_loans = @loan.cooperative.loans.status('all').order("SigningDate DESC") if @loan.cooperative
-    @repayments = @loan.repayments.order('DateDue')
```
linesarefuzzy: Repayments will probably come back in the future so I'd prefer to keep this line here, commented out—I think it's referenced in other commented out code

app/controllers/public/loans_controller.rb
```diff
@@ -1,28 +1,34 @@
...
+    
+    authorize @loan
   end
 
   def gallery
-    @loan = Loan.status('all').find(params[:id])
+    @loan = Loan.find(params[:id])
     @coop_media = @loan.coop_media(100, true).in_groups_of(4, false)
     @loan_media = (@loan.loan_media(100, true) + @loan.log_media(100, true)).in_groups_of(4, false)
+    authorize @loan
```
linesarefuzzy: Why put the authorize statements at the end?

app/models/currency.rb
```diff
@@ -13,6 +13,8 @@
 #
 
 class Currency < ActiveRecord::Base
+  validates :name, uniqueness: { scope: [:code, :country_code, :short_symbol, :symbol] }
```
linesarefuzzy: What's this for? Comment maybe?

app/models/loan.rb
```diff
@@ -58,11 +58,10 @@ class Loan < Project
...
 
-  scope :country, ->(country) {
-    joins(division: :super_division).where('super_divisions_Divisions.Country' => country) unless country == 'all'
-  }
-  scope :status, ->(status) { where(status: status) }
-  scope :active, -> { where(status_value: 'active') }
+  scope :status, ->(status) { where(status_value: status) }
+  scope :visible, -> { where.not(public_level_value: 'hidden') }
+  scope :active, -> { status('active') }
+  scope :related_loans, -> (loan) { loan.organization.loans.where.not(id: loan.id) }
```
linesarefuzzy: Awesome

app/models/loan.rb
```diff
@@ -149,64 +137,50 @@ def signing_date_long
...
   end
 
-  # def short_description
-  #   self.translation('ShortDescription')
-  # end
-  # def description
-  #   self.translation('Description')
-  # end
-
   def coop_media(limit=100, images_only=false)
```
linesarefuzzy: I was thinking we might want to replace these with keyword args, but not a big deal for now

app/models/loan.rb
```diff
@@ -149,64 +137,50 @@ def signing_date_long
...
     media = []
-    begin
-      self.logs("Date").each do |log|
-        media += log.media(limit - media.count, images_only)
-        return media unless limit > media.count
-      end
-    rescue Mysql2::Error # some logs have invalid dates
+    self.project_logs.find_each do |log|
+      media += log.get_media(limit: limit - media.count, images_only: images_only)
+      return media unless limit > media.count
```
linesarefuzzy: Hooray!

app/models/loan.rb
```diff
@@ -149,64 +137,50 @@ def signing_date_long
...
     # then loan pics
-    pics += get_media('Loans', self.id, limit - pics.count, images_only=true)
+    pics += loan_media(limit - pics.count, images_only=true)
     return pics unless limit > pics.count
     # then log pics
-    pics += self.log_media(limit - pics.count, images_only=true)
+    pics += log_media(limit - pics.count, images_only=true)
     return pics unless limit > pics.count
-    # then remaining coop pics
+  # then remaining coop pics
```
linesarefuzzy: Accidental dedent?

app/models/loan.rb
```diff
@@ -149,64 +137,50 @@ def signing_date_long
...
   def thumb_path
     if !self.featured_pictures.empty?
-      self.featured_pictures.first.paths[:thumb]
+      self.featured_pictures.first.item.thumb.url
     else "/assets/ww-avatar-watermark.png" end
   end
 
   def ensure_currency
     currency || Currency.find_by(code: 'USD')
   end
```
linesarefuzzy: Is this still needed?

config/environments/development.rb
```diff
@@ -38,13 +38,16 @@
   # For wordpress template
   config.x.wordpress_template = {
     division_urls: { # get division from request url
-      '/' => :us,
+      'theworkingworld.test' => :us,
+      'labase.test' => :argentina
```
linesarefuzzy: Does this (or somewhere—maybe readme) deserve a comment about editing hosts file?

lib/tasks/fake_data.rake
```diff
@@ -18,18 +18,21 @@ if Rails.env.development?
...
-      user.add_role :admin, Division.root
+      user = User.find_by(email: "admin@example.com")
+      if user.empty?
+        user = FactoryBot.create(:user, :admin,
+          email: "admin@example.com",
+          password: "xxxxxxxx",
+          password_confirmation: "xxxxxxxx"
+        )
+        user.add_role :admin, Division.root
+      end
```
linesarefuzzy: Nice... now it can be run multiple times

app/models/currency.rb
meltheadorable: In the specs it was creating like 50+ copies of the same currency, the other changes to the currency on related factories also prevent that, but the validation helps ensure no literally identical currencies get inserted.

config/environments/development.rb
meltheadorable: There's a lot of ways to accomplish this that vary by platform. It might be worth mentioning something in the dev setup docs somewhere but I don't want us duplicating a lot of info on setting this sort of thing up. 

In the meantime, it defaults to the US division if none of them match, so you can still get a working test environment for the features other than division detection without that.

app/models/loan.rb
meltheadorable: the `ensure_currency` method is used admin-side in the breakeven tables, so presumably it is still needed

app/models/loan.rb
meltheadorable: Yeah, I might do that anyway when I make my revision pass on this PR, I had left them alone at first because I was trying to get the minimal set of changes to get stuff working again, figuring we could worry about refactors later if we wanted them.

app/controllers/public/loans_controller.rb
meltheadorable: No particular reason, they could go anywhere after the @loan assignment, I can move them to be directly below it if you'd prefer.

app/controllers/public/loans_controller.rb
```diff
@@ -1,28 +1,34 @@
...
   include WordpressEmbeddable
+  after_action :verify_authorized
 
   def index
     params[:division] = get_division_from_url
-    @loans = Loan.filter_by_params(params).visible.
-      includes(:cooperative, division: :super_division).
-      paginate(:page => params[:pg], :per_page => 20).
-      order('signing_date DESC')
+    @params = { status: params[:status], pg: params[:pg], country: params[:country] }
```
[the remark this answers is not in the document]
meltheadorable: It's needed on the frontend side for the status-based filtering/active tab display stuff. I'd like to see a version of this that was a little less messy about that, but I didn't want to completely overhaul how all the frontend pages were built until we had a working core and knew we wanted to invest more time in the code.

app/controllers/public/loans_controller.rb
meltheadorable: It's not a difficult line to rewrite and the order clause won't work, if repayments are coming back we should track that in redmine rather than leaving dead code lying around that will need to be rewritten to function anyway.

lib/tasks/fake_data.rake
```diff
@@ -18,18 +18,21 @@ if Rails.env.development?
...
 
       # Create admin user
-      user = FactoryBot.create(:user, :admin,
-        email: "admin@example.com",
-        password: "xxxxxxxx",
-        password_confirmation: "xxxxxxxx"
-      )
-      user.add_role :admin, Division.root
+      user = User.find_by(email: "admin@example.com")
+      if user.empty?
```
emitche: When I try to run `rake dev:fake_data`, I get `NoMethodError: undefined method `empty?' for nil:NilClass` for this line.

I found replacing this with `user.nil?` does work.

PR #510, opened 2018-04-03
8227: Ensure question renaming works with legacy data

emitche:
- Adjust legacy models to use naming of older database in places that call data from the legacy table


emitche:
See #485 for the original changes for that changes LoanQuestion to Question.

emitche:
The failed test is accounting specific.

emitche:
@meltheadorable is good with these changes.

PR #511, opened 2018-04-03
8228 translations on login

arisah-ijeoma:


app/controllers/application_controller.rb
```diff
@@ -9,6 +9,11 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
+  # prepend_before_action is being used in devise not before_action
+  # we're attempting to have prepend_before_action happen for just devise powered pages
+  # while before_action happens for all other pages
+  prepend_before_action :set_locale
```
meltheadorable: I think you could do this whole thing as:

```ruby
prepend_before_action :set_locale, if: :devise_controller?
before_action :set_locale, unless: :devise_controller?
```
but do you know if anything else breaks if you just use `prepend_before_action` for everything?

config/locales/en/en.yml
```diff
@@ -341,6 +341,7 @@ en:
     misc: "There was an error communicating with QuickBooks. An administrator has been notified. Some data may be out of date. (Error: %{msg})"
     service_unavailable: "QuickBooks is temporarily unavailable. Please try again later."
     negative_balance: "This transaction is not allowed because it would result in a negative balance. The total balance on this date is %{amt}."
+  remember_me: "Remember Me"
```
meltheadorable: Does devise not have this in their own locale files? I'd check https://github.com/tigrish/devise-i18n if we aren't already using it, as it might save us some work in the future.

app/views/devise/sessions/new.html.erb
```diff
@@ -24,9 +24,10 @@
   <% if devise_mapping.rememberable? -%>
     <div class="field checkbox-field remember">
       <%= f.check_box :remember_me %>
-      <%= f.label :remember_me, class: "checkbox-label" %>
+      <% f.label :remember_me, class: "checkbox-label" %>
+        <%= t("remember_me") %>
```
meltheadorable: See other comment about the `devise-i18n` project.

app/controllers/application_controller.rb
arisah-ijeoma: I attempted this but it was breaking for some reason. I'll take a look at the `devise-i18n` gem

config/locales/en/en.yml
arisah-ijeoma: This was what I found in their documentation

```
Assuming you have not previously generated Devise's views into your project, that's all you need to do. If you have previously done this, *you will need to regenerate your views and then reapply whatever customizations that made you do this in the first place.*
```

Do we still want to proceed with this?
@meltheadorable

app/controllers/application_controller.rb
meltheadorable: This shouldn't be working either, unfortunately, because `if :devise_controller?` will always return true (symbols are truthy, apparently), so this is only working by accident right now. That's why you need to use the `if:` and `unless:` symbols as arguments. We can pair if you want to figure out why things were breaking/what was breaking when you did this the other way.

config/locales/en/en.yml
meltheadorable: @arisah-ijeoma I'm honestly not sure what the best approach will be here. I'd like to "uncustomize" the views here and then push to the i18n gem if at all possible, because it will save us maintenance difficulties later, and I'm pretty sure that there are things like password reset emails and other stuff associated with devise that are completely untranslated instead of only partially untranslated. The more we can rely on the gem for this the better, but I don't know what customizations we have made to the base devise setup.

config/locales/es/es.yml
```diff
@@ -329,6 +329,7 @@ es:
     misc: "Hubo un error en la comunicacíon con QuickBooks. Se ha notificado a un administrador. Algunos datos pueden estar desactualizados. (Error: %{msg})"
     service_unavailable: "QuickBooks no está disponible temporariamente. Por favor intentar de nuevo más tarde."
     negative_balance: "Esta transacción no es permitida porque resultaría en un saldo negativo. El saldo total a esta fecha es  %{amt}."
+  remember_me: "Recuérdame"
```
meltheadorable: Is this still needed or is this leftover from when we were doing it manually?

app/controllers/application_controller.rb
```diff
@@ -9,6 +9,8 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
+  # overrides 'route_translator' method to reset locale to English
+  skip_around_action :set_locale_from_url
```
meltheadorable: interesting that this is needed

config/locales/es/es.yml
arisah-ijeoma: it's a leftover
should be cleared when transifex does its thing

app/controllers/application_controller.rb
arisah-ijeoma: that method resets the locale to English if there is an error in any devise form

PR #512, opened 2018-04-04
6476 fix bug

arisah-ijeoma:


emitche:
I used to add `small-code-change` to signal to the tech lead that the amount of time to review will likely be short on the ticket. I also sometimes end, up reviewing the tiny ones to move them along quicker.

PR #514, opened 2018-04-06
8262 headstart on formatted textareas

arisah-ijeoma:


arisah-ijeoma:
CKEditor allows us format text in textareas. It does not support markdown
I intended to use javascript to show and hide the editor based on what view was showing; edit or show.

There are some failing specs here due to CKEditor but I'm not sure if the story is moving forward so they are being ignored for now

PS: If everything is going to boil down to using javascript to show or hide the editor, then I believe we can move forward with Rich text as earlier intended

arisah-ijeoma:
I am closing this ticket for now and redoing https://redmine.sassafras.coop/issues/8262 using summernote since the client specifically requested for it and both summernote and ckeditor came to the same blocking point (using js to hide the editor)

PR #516, opened 2018-04-10
7801 change dropdown to divisions

arisah-ijeoma:


spec/features/loan_index_spec.rb
```diff
@@ -45,7 +45,6 @@
     context 'with translations' do
       before { @loans = create_list(:loan, 3, :with_translations, :featured) }
       it 'renders translated loan description' do
-        visit public_loans_path
```
meltheadorable: Why were these removed?

app/models/loan.rb
```diff
@@ -81,9 +82,23 @@ def self.filter_by_params(params)
...
+
+    if self.check_before_division_filter(params[:division])
+      scoped = scoped.by_division(self.find_division_id_with(params[:division]))
+    end
+
     scoped
   end
 
+  def self.find_division_id_with(name)
+    Division.find_by(name: name).id if name
```
meltheadorable: Does this work? `get_division_from_url` returns a symbol like `:us` or `:argentina`, but division names are something like `La Base USA`, I wouldn't expect this to behave correctly as written when there isn't a division param passed in.

spec/features/loan_index_spec.rb
arisah-ijeoma: there was a duplication

app/models/loan.rb
arisah-ijeoma: I had an issue with git
perhaps that change wasn't added

I'll work on it
thanks for the find

arisah-ijeoma: There's a check for this below but because of your comment, I've realized that the symbols are not country codes and I have updated the check accordingly

Thanks!

meltheadorable: The problem is that we want each division to default to showing only its own loans, so we actually _want_ it to filter by the value of `get_division_from_url` if it can -- if I'm following this correctly, right now it will not do that, and instead will default to showing all unless it specifically detects the name param from the dropdown.

That's why I kept mentioning that other `short_name` issue in conjunction with this -- if we set the short name for each division to match the param we get from `get_division_from_url` we can use the correct filtering based on that.

emitche:
Great work here, @arisah-ijeoma.

For QA:
In the spec there is this piece: "When a division is selected, its loans and its descendant divisions' loans should be shown." That's not happening right now. Right now loans from a child division do not show up in the loans index when the list is filtered by a parent division.

I think we can keep root division as it though, as All Divisions captures all of the root division's descendants.

emitche:
@arisah-ijeoma If the last request will require a lot of work, just let me know. We can move that specific piece of this feature into a new ticket. Everything else looks great and ready to go. Thank you!

PR #517, opened 2018-04-10
8278 add short name to division

arisah-ijeoma:
Could you please look this over for me @meltheadorable 

I feel I'm missing something

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,8 +1,7 @@
 # This is generic data that is needed for any instance of this app to work properly.
 # It should not be specific to a particular instance.
 
-Division.root.destroy if Division.root.present?
-Division.find_or_create_by(id: 99, name: 'Root Division', short_name: 'root-division') unless Division.root
+Division.find_or_initialize_by(id: 99).assign_attributes(name: 'Root Division', short_name: 'root-division').save!
 Division.recalibrate_sequence(gap: 1)
 
 Currency.find_or_create_by(id: 1, name: 'Argentinean Peso', code: 'ARS', symbol: 'AR$',


app/controllers/admin/divisions_controller.rb
```diff
@@ -1,4 +1,6 @@
 class Admin::DivisionsController < Admin::AdminController
+  before_action :find_division, only: [:show, :update, :destroy]
```
meltheadorable: My expectation was that we would only use these for the public site for now -- don't wanna muck too much with the behavior of the admin panel right now. I thought we may need to use this so that we could find divisions by the kinds of symbols that we get back from the public frontend (`:us`, `:argentina`, `:nicaragua`), especially if `friendly_id` supported i18n, but its i18n support seems pretty minimal, so it might not be worth using for this just now.

app/models/division.rb
```diff
@@ -86,6 +91,7 @@ class Division < ActiveRecord::Base
   has_attached_file :logo, styles: { banner: "840x195>" }
   validates_attachment_content_type :logo, content_type: /\Aimage\/.*\z/
 
+  validate :ascii_encoding
```
meltheadorable: On second thought, lets just drop this validation. I don't think it accomplishes anything.

db/migrate/20180410173704_update_division_with_short_name.rb
```diff
@@ -0,0 +1,9 @@
+class UpdateDivisionWithShortName < ActiveRecord::Migration[5.1]
+  def change
```
meltheadorable: Did you test rolling this back? I anticipate that rails won't know how to reverse this, so it might be better to do separate `up` and `down` methods for this one, where `down` changes all the short name fields to `nil`.

spec/factories/divisions.rb
```diff
@@ -55,6 +57,7 @@ def root_division
     description { Faker::Lorem.sentence }
     name { Faker::Address.city }
     parent { root_division }
+    sequence(:short_name) { |n| "#{Faker::Lorem.word}-#{n}" }
```
meltheadorable: Neat!

db/seeds.rb
```diff
@@ -2,7 +2,7 @@
 # It should not be specific to a particular instance.
 
 Division.root.destroy if Division.root.present?
-Division.create(id: 99, name: 'Root Division') unless Division.root
+Division.find_or_create_by(id: 99, name: 'Root Division', short_name: 'root-division') unless Division.root
```
meltheadorable: Good call, except that `find_or_create_by` tests that _all_ things match, and the `Division.root.destroy` makes the whole thing sorta moot anyway. I want us to spend some time gradually refactoring the seeds so that we _can_ safely re-run them though, so I suggest using something like

`Division.find_or_initialize_by(id: 99).assign_attributes(name: 'Root Division', short_name: 'root-division').save!` (dropping the unless was intentional) and then cutting the `Division.root.destroy` line.

PR #518, opened 2018-04-11
filter division dropdown

arisah-ijeoma:


app/models/division.rb
```diff
@@ -90,6 +90,7 @@ class Division < ActiveRecord::Base
   validates :parent, presence: true, if: -> { Division.root.present? && Division.root_id != id }
 
   scope :by_name, -> { order("LOWER(divisions.name)") }
+  scope :by_public, -> { where(public: true) }
```
meltheadorable: I think this should just be `public`

arisah-ijeoma: `public` is a keyword in rails @meltheadorable

arisah-ijeoma:
I missed that!

spec/features/loan_index_spec.rb
```diff
@@ -56,14 +60,28 @@
...
       end
     end
+
+    context 'with divisions' do
+      scenario 'filters with division' do
+        select 'chicken', from: 'division'
+        @loans.each do |loan|
+          expect(page).to have_content loan.summary
+        end
+      end
```
linesarefuzzy: I don't see how this test works. The `@loans` were not assigned to the "chicken" division, were they? If they weren't, then they should *not* show when that division is selected. Either way, in order to test filtering, we need to test that some loans do show and also that some *don't* show. And the "don't show" expectation needs to come first, to make sure capybara waits for the filtering to happen.

app/views/public/loans/index.html.slim
```diff
@@ -13,8 +13,8 @@
           li class=('active' if @params[:status] == 'all')
             = link_to t(:show_all), @params.merge(status: 'all', pg: nil)
       .country
-        = label_tag(:country, "#{t :country}: ")
-        = select_tag(:country, options_for_select([[t(:show_all), 'all']] + @countries, @params[:country]),
+        = label_tag(:division, "#{t :division}: ")
+        = select_tag(:division, options_for_select([[t(:show_all_divisions), 'all divisions']] + @divisions, @params[:division]),
```
linesarefuzzy: Why not just use "all" for the value? I have a feeling the space character in that value will cause problems later. For that matter, the label could also be simply "All", since "Division" is displayed right next to it.

spec/policies/division_policy_spec.rb
```diff
@@ -89,6 +90,22 @@
       end
     end
 
+    context 'with a signed user' do
```
linesarefuzzy: Typo: should be "signed in user"

spec/policies/division_policy_spec.rb
```diff
@@ -89,6 +90,22 @@
...
+      end
+    end
+
+    context 'without a signed user' do
+      let(:user) { nil }
+
+      it 'shows all divisions' do
+        expect(division_scope(user)).to contain_exactly(parent_division, division, child_division)
+      end
+    end
```
linesarefuzzy: Nice tests!

app/models/loan.rb
```diff
@@ -81,9 +82,23 @@ def self.filter_by_params(params)
...
 
+  def self.find_division_id_with(name)
+    Division.find_by(name: name).id if name
+  end
+
+  def self.check_before_division_filter(div_params)
+    # division defaults when no parameter is set and this messes with the scoping
+    div_params != 'all divisions' && !div_params.is_a?(Symbol) && div_params.length > 2
+  end
+
```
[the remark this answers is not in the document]
arisah-ijeoma: that was actually changed but I lost it in a commit
thanks for the find @linesarefuzzy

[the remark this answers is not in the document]
arisah-ijeoma: sorry....this was fixed in #516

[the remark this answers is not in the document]
arisah-ijeoma: I will update #516 with some of these changes

app/policies/division_policy.rb
```diff
@@ -33,7 +33,7 @@ def select?
 
   class Scope < Scope
     def resolve
-      scope.where(id: accessible_ids)
+      user ? scope.where(id: accessible_ids) : scope.where(public: true)
```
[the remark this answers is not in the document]
arisah-ijeoma: this has been implemented in a different story but it wouldn't hurt to do it here as well

app/views/public/loans/index.html.slim
arisah-ijeoma: 👍

spec/features/loan_index_spec.rb
```diff
@@ -1,7 +1,12 @@
 require 'rails_helper'
 
 feature 'visit loan index page' do
-  before { @loans = create_list(:loan, 3, :active, :featured) }
+  before do
+    @loans = create_list(:loan, 3, :active, :featured)
+    create(:division, name: 'chicken')
+    create(:division, name: 'kale', public: false)
+  end
```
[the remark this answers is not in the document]
arisah-ijeoma: I meant for it to be there
Fore some reason, it was not getting created in the context's before do and I figured divisions were needed to run that page anyway

spec/policies/division_policy_spec.rb
arisah-ijeoma: 👍

arisah-ijeoma:
this pr is ready to be merged into #516 
@linesarefuzzy @emitche

emitche:
Alright. I'll get to this soon. Thank you!

arisah-ijeoma:
@emitche, the said change for #516 has to be done here since this is where most of the related code are

app/models/loan.rb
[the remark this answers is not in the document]
arisah-ijeoma: This PR has the latest change for this comment so I'll work wit this PR
For the symbol issue, it is not from the url. `get_division_from_url` returns a symbol by default if params[:division] is not set

PR #520, opened 2018-04-20
8240 public loans js

emitche:
- Fix carousel in loan details page
- Convert loan details tabs to sections
  - Due to Bootstrap 3 bugs with tabs not working, tabs have been removed
- Open images on loan gallery page in a large modal to replace broken lightbox functionality
- Display the long caption for an image in the header of the modal


linesarefuzzy:
I found some issues on this branch that had already been fixed on develop. So I merged in develop and resolved conflicts.

I found the following problems in QA that I have not fixed:

- The active tab is not highlighted
- The "not found" message is not displayed if the description is not found
- On some loans (e.g. 1347), the JS console shows `Uncaught TypeError: Cannot read property 'offsetWidth' of undefined` (it takes a few seconds to appear)
- The left and right arrows on the carousel don't work
- Clicking an image in the gallery does cause it to be enlarged, but it is grayed out and no scrolling or clicking is possible

All of the above are true both before and after merging develop, on my Mac in Chrome.

emitche:
Let's move this to 2.9 and resolve these issues at a later time. Thank you for your QA!

emitche:
"Clicking an image in the gallery does cause it to be enlarged, but it is grayed out and no scrolling or clicking is possible"

image

Yes, this is very much a bug. It used to work, but I will see what's going on. The test image used is an error message screenshot from a long time ago. Please ignore the content of the image.

emitche:
"The "not found" message is not displayed if the description is not found"

It displays for me. That is weird.

image

emitche:
"The active tab is not highlighted"

The active tab is distinctly styled from what is displayed for the non-active tab. Is there a specific way you want the active tab to be styled? The styling is different from what we had in our earlier version of the public loans page.

image

emitche:
"The left and right arrows on the carousel don't work"

I'm seeing this on the current public loans page, but not in this branch.

Active site bug (the page actually jumps when clicking the arrows):
https://madeline.theworkingworld.org/us/loans/2505

app/views/public/loans/show.html.slim
```diff
@@ -13,27 +13,32 @@
...
-              a.carousel-control.right data-slide="next" href="#myCarousel"  &rsaquo;
-
         / .lend
         /   a.button.lend_button.pull-right href="#"
         / h5.coop_name Current Report
 
+        a href=public_gallery_path
+          #loanImageCarousel.carousel.slide data-ride="carousel"
+            ol.carousel-indicators
+              - remainingPictures = @pictures.length
```
emitche: @meltheadorable There is likely a cleaner way to do this loop. Please let me know how I can improve it. Thank you!

meltheadorable: I'm not fully grasping the goal of this loop, what's it doing?

meltheadorable: Without being 100% sure what the goal is, this might be what you're looking for:

```slim
ol.carousel-indicators
  - @pictures.each.with_index do |picture, i|
    li data-target="#loanImageCarousel" data-slide-to="#{i}" class="#{'active' if i.zero?}"
```

app/views/public/loans/show.html.slim
```diff
@@ -60,67 +65,64 @@
...
+                td
+                  - if loan.id == @loan.id
+                    = loan.signing_date_long
+                  - else
+                    a href=public_loan_path = loan.signing_date_long
+                td = format_currency(loan.amount, loan.currency)
+                td = loan.status
+                td.hidden-xs style="width: 50%" = render_translation(loan.details)
 
+      // Tab content for timeline and repayments
```
meltheadorable: I get the motivation to leave code like this lying around, but it's all in git, I think we should delete big blocks like this when we aren't using them. We can always check out an old branch to see what it looked like if we reimplement this tab.

emitche: Ok.

emitche:
I like your loop suggestion.

linesarefuzzy:
I don't know why it works for you and not me, @emitche . We should probably look at this together sometime. It's working even less for me now, the two carousel images are overlapped: 

screencapture-localhost-3000-us-loans-1420-2018-06-06-16_12_42

emitche:
Bumped to next sprint. Thanks for the QA review, @linesarefuzzy!

emitche:
@linesarefuzzy This is ready for review. This code works with the most recent version of the themes for La Base/TWW United States, La Base Nicaragua, and La Base Argentina.

Please feel free to do a code review, but there's nothing complex in the code. Just lots of moving things around and cleaning up things.

I worked hard to get tabs working, but it's not worth it in terms of time. I have moved the tabs to sections instead.

linesarefuzzy:
- Tabs:
  - Totally fine to drop tabs for now and use sections. 
  - "All Loans to this Co-op" section:
    - The row links don't work. They just reload the current loan. 
    - The rows should be ordered by signing date descending
    - Probably the Summary should be shown in the list rather than the Details
    - Since the current loan is not included in the list, the title should be changed to "Other Loans to this Co-op"
- Carousel:
  - The bottom of the image is cut off. The old carousel took up more vertical space. It also showed black bars on the sides or top and bottom of the image to show the whole image. I think I like that better. If you like filling the space better, that's okay, but the image should be centered in the space rather than just cutting off the bottom. 
  - The top corners are rounded, the bottom corners are not. 
  - ~~I don't see any back and forward buttons, but the selector works so not a big deal.~~ Oh, I just saw the comment about that on the Redmine ticket. No problem. 
- Lightboxes: Look great!

The changes to the "All Loans to this Co-op" section are the only ones I think really need to be fixed before deployment. The rest could be done later in a separate issue. Or the "All Loans to this Co-op" section could be removed for now—the current public site doesn't have it. 

Thank you for slogging through all this yucky stuff!

emitche:
Thanks for this update, @linesarefuzzy.

emitche:
### Tabs

"The rows should be ordered by signing date descending"
- This is not a JS site fix. I will add it to the backlog and make sure it is prioritized. But this is not part of the scope of this work.

All other tabs issues noted have been fixed.

### Carousel

- Border radius issue fixed
- Size and centering will go in a separate ticket and be prioritized in the backlog.

### Lightboxes

- Fixed for Argentina

linesarefuzzy:
I think this is finally ready to release! Yay!

I just pushed a commit to order the "other loans" by signing date, so you don't have to make another ticket for that. 

The only item still remaining is this one—not a high priority:

* Carousel: The bottom of the image is cut off. The old carousel took up more vertical space. It also showed black bars on the sides or top and bottom of the image to show the whole image. I think I like that better. If you like filling the space better, that's okay, but the image should be centered in the space rather than just cutting off the bottom.

So good to finally have this finished! 

🎆 🍾 🎊

emitche:
Yay! The carousel work will go in a separate ticket for a future release. Thank you so much, @linesarefuzzy!

PR #521, opened 2018-04-24
8305: Add public to public level options

arisah-ijeoma:


app/controllers/concerns/wordpress_embeddable.rb
```diff
@@ -26,7 +26,7 @@ def update_template
 
     base_uri = Rails.configuration.x.wordpress_template[:base_uri][get_division_from_url]
 
-    WordpressTemplate.update(
+    ::WordpressTemplate.update(
```
linesarefuzzy: Didn't we determine that this change was unnecessary?

db/migrate/20180425170810_make_nil_projects_public.rb
```diff
@@ -0,0 +1,13 @@
+class MakeNilProjectsPublic < ActiveRecord::Migration[5.1]
+  def change
+    nil_projects = Project.where(public_level_value: nil)
+
+    if nil_projects
+      nil_projects.each do |np|
+        np.update_attribute(:public_level_value, 'public')
+      end
+    end
```
linesarefuzzy: It's fine for this migration, but for future reference, this could have been a one-liner:

`Project.where(public_level_value: nil).update_all(public_level_value: 'public')`

arisah-ijeoma: 👍

app/controllers/concerns/wordpress_embeddable.rb
arisah-ijeoma: not really but I'll make the change in this PR

PR #523, opened 2018-04-30
8359: removes migration, adds validations

meltheadorable:
This needs QA, preferably running the migrations against a copy of the v2.6 prod data, to make sure that nothing in the admin forms for divisions or anything else breaks because of the change.


db/migrate/20180410173704_update_division_with_short_name.rb
```diff
@@ -1,13 +0,0 @@
...
-    div_without_short_names.each do |division|
-      division.short_name = division.name.parameterize
-      division.save
-    end
-  end
-
-  def down
-    Division.update_all(short_name: nil)
-  end
-end
```
arisah-ijeoma: are we going with a new strategy?

meltheadorable: Yep, manually adding these to divisions when we need them.

PR #524, opened 2018-05-01
8340 show loan division children

arisah-ijeoma:


app/models/loan.rb
```diff
@@ -92,7 +91,8 @@ def self.filter_by_params(params)
 
     div_check = div_param != 'all' && !div_param.is_a?(Symbol) && !URL_DIVISIONS.include?(div_param)
     if div_check
-      scoped = scoped.by_division(params[:division])
+      division = Division.find_by(short_name: div_param)
```
meltheadorable: I don't think this is going to work, there have been some changes now with all the 2.7.3 stuff and we may need to regroup on this and make sure we have adequate tests.

PR #525, opened 2018-05-02
7821 add delete fxn for logs

emitche:
- Add delete functionality to logs list in step modal
- Adjust UI on log deletion to display correct information in calendar and timeline


app/assets/javascripts/admin/log_list_view.coffee
```diff
@@ -15,13 +15,20 @@ class MS.Views.LogListView extends Backbone.View
...
+    context = @$(e.currentTarget).data('context')
     if (resp)
-      $.ajax(method: "DELETE", url: "/admin/logs/#{logId}")
-      .done => @refresh()
+      $.ajax(method: "DELETE", url: "/admin/logs/#{logId}/?context=#{context}")
+      .done (response) =>
+        # Replace log list
+        @$el.html(response)
+
+        # Remove log from timeline
```
emitche: This is only needed if we are in the timeline context and the "More" link is not deleted if there are no more logs left for a specific step.

I started trying to pass a context to the step modal and then this would later pass the context to the logs list. However, the step modal did not render correctly.

I've stopped on that piece for now. Perhaps removing the "More" when no logs are left can be added to the backlog.

app/assets/javascripts/admin/log_list_view.coffee
```diff
@@ -15,13 +15,20 @@ class MS.Views.LogListView extends Backbone.View
...
+        @$el.closest(".content").find(".project-step .log-summary[data-log-id='#{logId}']").remove()
+
       .fail (response) -> MS.alert(response.responseText)
 
+#  the last bit is to fix the refresh since it breaks the step modal log list on delete
```
emitche: Not sure this comment remains true now.

app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -1,25 +1,29 @@
-- if @logs.blank? && @context != 'calendar'
-  = content_tag(:div, t('common.no_records'), class: 'no-records-msg')
+- @context = @context ? @context : 'logs_list'
+- step_modal_context = @context == ('calendar' || 'timeline') ? true : false
```
emitche: I added this logic to not have to use `if context == `calendar'` in tons of places.

Eventually it would be nice for "calendar" context not to represent both the calendar and timeline table.

app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -1,25 +1,29 @@
-- if @logs.blank? && @context != 'calendar'
-  = content_tag(:div, t('common.no_records'), class: 'no-records-msg')
+- @context = @context ? @context : 'logs_list'
```
emitche: Usually the calendar context (which is both the timeline table and calendar) and the dashboard context are passed in as the context variable. However, this is not the case when the logs list is a page. So context is being set here.

app/views/admin/project_logs/_log_list.html.slim
```diff
@@ -1,25 +1,29 @@
...
 
-  - unless @context == 'calendar'
-    - if @step
-      h5
-        - if @step.milestone?
-          i.fa.fa-fw.fa-flag-o
-        - elsif @step.checkin?
-          i.fa.fa-fw.fa-calendar
-        = @step.summary
+  / TODO: Under what context is the large heading for step needed?
```
emitche: This is a styled choice to show a step heading.

However, later down in the logic, you'll see that the step name already is displayed as link in the step. I'm unsure why we had this logic previously. We are only using the linked step name at this time.

emitche:
The missing UI pieces are big enough that this needs to be revised before it is ready for review.

app/controllers/concerns/log_controllable.rb
```diff
@@ -43,12 +43,12 @@ def save_and_render_partial
...
-        step: @step,
-        context: 'timeline',
-        mode: :show
-      }
+      if @step
+        @expand_logs = @step.logs_count > 0
+        render partial: 'admin/project_logs/log_list'
+      else
+        render partial: 'admin/project_logs/log_list'
+      end
```
arisah-ijeoma: this can be

```
@expand_logs = @step.logs_count > 0 if @step
render partial: 'admin/project_logs/log_list'
```

emitche: Thanks!

emitche:
This did not get QAed in time for the 2.8 release.

linesarefuzzy:
QA looks beautiful. I pushed a small style change to use pointer when hovering over clickable steps.

PR #530, opened 2018-05-08
8209 questionnaire rich text

arisah-ijeoma:
I got some failing spec here because of the summernote editor and wouldn't mind pairing on it

Otherwise, summernote works

We might need to hasten up the page load


Gemfile
```diff
@@ -20,6 +20,11 @@ gem 'uri-js-rails' # URI manipulation
 gem 'bootstrap-datepicker-rails'
 gem 'paperclip', '~> 5.2'
 
+gem 'turbolinks', '~> 5.0.0'
```
meltheadorable: Is turbolinks really required? We removed this for good reasons, unfortunately, and I don't trust that adding it back will be consequences-free.

arisah-ijeoma: I'll remove it an see if we can go without

arisah-ijeoma:
I can not find `Edit Response` on the page

arisah-ijeoma:
got rid of turbolinks

emitche:
Line breaks that users have previously created in their answers are not being respected when Summernote opens the text of the answer. I wonder if there's a way to fix that.

arisah-ijeoma:
I'll branch of this branch and look into that @emitche

arisah-ijeoma:
@emitche, I have a fix out ready for testing in #543 
it should be merged into this branch before the branch is merged into master

PR #532, opened 2018-05-09
8377 breakeven styling improvements

emitche:


emitche:
Given that this is a small code change, I'm reviewing it and it can be removed as part of the larger pull request it is a part of.

PR #537, opened 2018-05-29
7997 set journal number field

arisah-ijeoma:


app/models/accounting/quickbooks/transaction_builder.rb
```diff
@@ -29,6 +29,7 @@ def build_for_qb(transaction)
 
         je.private_note = transaction.private_note
         je.txn_date = transaction.txn_date if transaction.txn_date.present?
+        je.doc_number = set_journal_number(transaction)
```
meltheadorable: This will overwrite the existing doc number if there is one, and we don't want to interfere with the existing notes if it's avoidable. We should find a way to prepend/append this.

linesarefuzzy: Right. Perhaps for now we should just move this line to an `else` clause of the preceding `if transaction.qb_id` block. The ticket is only for new txns. And ultimately we may want to make any needed changes to the JN field more intentionally anyway.

linesarefuzzy:
Ok I just did the QA and it looks good! Fuzzy stamp of approval!

PR #539, opened 2018-05-30
7015 interest transaction currency

emitche:
- Ensure currency shows up on new transactions regardless of transaction type


emitche:
Thanks. I'll bump this to sprint 2.10. I also was seeing that, but then it appeared to fix itself. But it didn't, I am seeing. Thank you!

PR #540, opened 2018-06-11
8402 not applicable heading do not show

arisah-ijeoma:
Turns out the fix was already in the system, not just applied


emitche:
If no response is initially given and the item is marked N/A or has never had a response, then the group and question are hidden appropriately.

If an item is marked as N/A but there is a response or there has been a response that has been removed, the group and question do not hide as expected.

Maybe the order of the items in the if statement need to be reordered.

emitche:
I'm finding that items that are marked N/A are not hidden in the loan memo table of contents and content.

Loan questionnaire
image

image

image

Then when marking a never answered question to not applicable:
image

image

image

linesarefuzzy:
It works, YAY!

PR #542, opened 2018-06-14
add new loan type to option set

arisah-ijeoma:


arisah-ijeoma:
this was done directly on the server

PR #543, opened 2018-06-15
8528 fix bug

arisah-ijeoma:
this fixes the bug of line breaks when editing in summernote

it should be merged into #530 before #530 is merges into master


emitche:
Code looks fine.

emitche:
Pasting in rich text to the answer leaves some weird issues that can't be fixed. I think this might be a separate bug though. Here is an example:

image

emitche:
@arisah-ijeoma Please go ahead and merge this into 8209. Thank you!

PR #545, opened 2018-06-21
8554 make country required in coops

arisah-ijeoma:
With this, coops that have divisions that don't have currencies will have to be set manually

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -49,7 +49,7 @@ class Organization < ActiveRecord::Base
   include DivisionBased
 
   belongs_to :division
-  belongs_to :country
+  belongs_to :country, required: true
   belongs_to :primary_contact, class_name: 'Person'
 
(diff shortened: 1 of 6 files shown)


app/models/organization.rb
```diff
@@ -49,7 +49,7 @@ class Organization < ActiveRecord::Base
   include DivisionBased
 
   belongs_to :division
-  belongs_to :country
+  belongs_to :country, required: true
```
linesarefuzzy: "You can use :inverse_of to avoid an extra query during validation."
https://apidock.com/rails/ActiveRecord/Associations/ClassMethods/belongs_to

arisah-ijeoma: this would only work out if there is an association on both models

as it stands, there is only a belongs_to association in the organization model but there is no corresponding association to organization in the country model

PR #547, opened 2018-07-03
8583 nested divisions public site

arisah-ijeoma:


emitche:
@arisah-ijeoma Do you know why the tests are failing? Are you able to fix them at this time? Thank you!

app/helpers/application_helper.rb
```diff
@@ -72,4 +72,28 @@ def content_tag_if(condition, *args, &block)
   def app_version_number
     @app_version ||= File.read(File.join(Rails.root, "VERSION"))
   end
+
+  def division_select_options(include_root: true, include_all: false, public: false)
+    divisions = public ? Division.published : current_user.accessible_divisions
```
meltheadorable: I think these calls should really be wrapped in a `policy_scope` call -- we don't want to be setting up any public-facing records that haven't been filtered through the policy/authorization system

app/helpers/application_helper.rb
```diff
@@ -72,4 +72,28 @@ def content_tag_if(condition, *args, &block)
   def app_version_number
     @app_version ||= File.read(File.join(Rails.root, "VERSION"))
   end
+
+  def division_select_options(include_root: true, include_all: false, public: false)
```
meltheadorable: Can we default to the more restrictive option (`public_only: true` [and also change `public` to `public_only`]) and change the other calls to match? 

In general it's a best practice, when you're dealing with collections that can either be more or less restricted, to make the default the more restricted option, that way there's less of a chance that you'll leak any extra information.

arisah-ijeoma: thank you

app/helpers/application_helper.rb
arisah-ijeoma: 👍

PR #548, opened 2018-07-04
fix table not rendering

arisah-ijeoma:


emitche:
`html_safe` shouldn't be used on user-generated content. We could use `sanitize` instead. But we'd need to specify which attributes are allowed.

Example:
`<%= sanitize @comment.body, tags: %w(strong em a), attributes: %w(href) %>`
See the [Rails Helper documentation](http://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html).

emitche:
To capture all the features that have not been captured since adding sanitize, we also need to add `b i u ol ul p` to tags.

We also need to add the attributes `style src data-filename`.

Underline using `u` was phased out, so it's interesting that Summernote uses it.

emitche:
Looking good so far. The bug is fixed. Thank you!

arisah-ijeoma:
for the other bit with images, I'm going to update #549 
that pr will be merged in here before this is merged into develop

PR #549, opened 2018-07-04
8599 images can be added in summernote

arisah-ijeoma:


emitche:
The failed spec is unrelated to the changes.

arisah-ijeoma:
the pr depends on #548 8598

it has to be merged into it first

I have already set it as the base branch for this pr

emitche:
Thanks for the update!

emitche:
The styling is not saving on the images, just the images themselves. You might need to add the attributes `style src data-filename`.

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
```diff
@@ -84,12 +84,14 @@
               = f.input_field :"#{question.attribute_sym}[number]", as: :decimal,
                 value: response.number
 
-            - tags = %w(table tbody tr td b i u ol ul li p)
+            - tags = %w(table tbody tr td b i u ol ul p img)
```
emitche: We need to add `li` here.

emitche:
I didn't do a QA review on this yet. But there is a code revision that is needed because of changes since the last merge.

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
arisah-ijeoma: the lists were not saving appropriately

unordered and ordered @emitche

emitche: Oh. So we need to not include `li`?

emitche: I've tested the current functionality on staging. v2.11.1 uses this line: `- tags = %w(table tbody tr td b i u ol ul li p)` So we do need to include `li` in this list.

I will make that change.

emitche:
Saved tables do not display properly.

#### Edit view
image

#### Show view
image

This is how the table should display:

#### Edit view
(The last cell was initially D3 and changed to D34 in later images.)
image

#### Show view
image

emitche:
I am still working on QA for this ticket. It will be moved to a different sprint.

emitche:
summernote-init.js probably should be this when resolving the merge conflicts:

```
        ["table", ["table"]],
        ["para", ["ul", "ol", "paragraph"]],
        ["insert", ["link", "picture"]]
```

emitche:
QA review:

The image is not saving properly when tested in the UI. Instead, in the html, an empty `` tag exists.

I think the problem is that attributes are not enabled in the santitizer. These attributes are required to properly display the image. At a minimum, we need to include the attributes: style, src, and data-filename. We may also want to include _moz_resizing.

What shows in the html when I add an image into a loan memo response for Summernote while in the edit form view before saving:
image

What shows in the html when I save the form: (No image displays in the UI.)
image

emitche:
Test failure is a known failure unrelated to this ticket.

emitche:
I'd like to hold on merging this work until the Summernote work for documentation is completed.

PR #551, opened 2018-07-05
8577 main header divisions bug

emitche:
- Fix UI bug where header items jump when the division selector shows


emitche:
I can scope this more (by size of screen and container) if that would be useful.

PR #554, opened 2018-07-10
8613 add link to public loan card

arisah-ijeoma:
I branched off #547 so it has to be merged into it first before #547 goes into develop

### DIFF
--- a/app/models/loan.rb
+++ b/app/models/loan.rb
@@ -87,9 +87,9 @@ class Loan < Project
     scoped = scoped.status(params[:status]) if params[:status] != 'all'
 
     if params[:division] != 'all'
-      division_ids = Division.find_by(short_name: params[:division])&.self_and_descendant_ids
+      division_ids = Division.find_by(short_name: params[:division])&.self_and_descendants&.select(&:public?)&.map(&:id)
     else
-      division_ids = Division.root.descendant_ids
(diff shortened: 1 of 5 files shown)


app/models/loan.rb
```diff
@@ -87,9 +87,9 @@ def self.filter_by_params(params)
     scoped = scoped.status(params[:status]) if params[:status] != 'all'
 
     if params[:division] != 'all'
-      division_ids = Division.find_by(short_name: params[:division])&.self_and_descendant_ids
+      division_ids = Division.find_by(short_name: params[:division])&.self_and_descendants&.select(&:public?)&.map(&:id)
     else
-      division_ids = Division.root.descendant_ids
+      division_ids = Division.root.descendants.select(&:public?)&.map(&:id)
```
meltheadorable: I _think_ this would be cleaner and a little more efficient, but should still work. In this version you are only using DB queries, so there's no iterating over ruby methods. In theory this works, because `closure_tree` returns chainable scopes.

```ruby
if params[:division] != 'all'
  division_ids = Division.find_by(short_name: params[:division])&.self_and_descendants&.where(public: true)&.pluck(:id)
else
  division_ids = Division.root.descendants.&where(public: true).&pluck(:id)
end
```

PR #555, opened 2018-07-10
8621 add nested divisions to division form

arisah-ijeoma:
this needs to be merged into #554

### DIFF
--- a/config/locales/en/en.yml
+++ b/config/locales/en/en.yml
@@ -165,7 +165,7 @@ en:
         division:
           attributes:
             parent:
-              invalid: 'Parent Division and Name can not have the same value'
+              invalid: 'Division and Parent Division cannot be the same'
         loan:
           attributes:
             primary_agent:
--- a/config/locales/en/en.yml
+++ b/config/locales/en/en.yml
@@ -162,6 +162,10 @@ en:
           attributes:
             primary_agent:
               same_as_secondary: 'The point person for this project cannot be the same as the second point person'
+        division:
+          attributes:
+            parent:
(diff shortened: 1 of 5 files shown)


arisah-ijeoma:
Considering that the nested division is being used on both the public and admin site in a generic manner, having it specially doctored for the new/edit division felt like it would have some sort of code smell

As a solution, I added a validation that would not allow the user save the same division as its parent division. There is also an accompanying test @emitche

emitche:
That works. Thanks.

config/locales/en/en.yml
```diff
@@ -162,6 +162,10 @@ en:
           attributes:
             primary_agent:
               same_as_secondary: 'The point person for this project cannot be the same as the second point person'
+        division:
+          attributes:
+            parent:
+              invalid: 'Parent Division and Name can not have the same value'
```
emitche: Change to "Division and Parent Division cannot be the same".

spec/features/admin/division_flow_spec.rb
```diff
@@ -12,4 +13,12 @@
   include_examples :flow do
     subject { division }
   end
+
+  scenario "division name and parent division can't be the same" do
+    visit admin_division_path(division)
+    find('.edit-action').click
+    select 'Cream', from: 'division_parent_id'
+    click_on 'Update Division'
+    expect(page).to have_content('Parent Division and Name can not have the same value')
```
emitche: Update to "Division and Parent Division cannot be the same" or use the yml content.

PR #556, opened 2018-07-10
8622 update card on loan detail page

arisah-ijeoma:


app/views/public/loans/show.html.slim
```diff
@@ -38,6 +38,14 @@
       h2.section_title = t :about_this_project
       .status-box
         table.info
+          tr
+            th #{t(:division)}:
```
emitche: Use `= t(:division)`

app/views/public/loans/show.html.slim
```diff
@@ -38,6 +38,14 @@
...
+            td
+              a href=public_loans_path(division: @loan.division.short_name)
+                = @loan.division.name
+          tr
+            th #{t(:coop)}:
```
emitche: Use `= t(:coop)`

PR #558, opened 2018-07-24
8652 add and test nested divisions on loan filters

arisah-ijeoma:
As an aside question, I noticed that the filter only returns loans associated with the division and doesn't consider the division's children
Is this the expected behavior?


emitche:
I spoke with @meltheadorable. The code is fine for this.

As for the filtering, that's an underlying Wice Grid issue that will take a lot of time to figure out and fix.

PR #559, opened 2018-07-25
8681 update sprockets version

arisah-ijeoma:
Since sprockets is a dependency under sass-rails, it is not listed in the Gemfile

Sass-rails is already up to date so running `bundle update sprockets` updated the Gemfile.lock file only


emitche:
Okay. Would we need to run `bundle update sprockets` on deployment on staging and production?

arisah-ijeoma:
I think merging this should suffice. However if it doesn't then we can run that command on both environments

emitche:
Okay. Thanks for the update. :)

PR #560, opened 2018-07-25
8682 short name division error

arisah-ijeoma:


emitche:
This works beautifully. Thank you!

PR #561, opened 2018-07-27
8691 summernote list extra space bug

arisah-ijeoma:


app/assets/stylesheets/admin/loans/_questionnaires.scss
```diff
@@ -71,6 +71,10 @@ section.questionnaires {
     margin-bottom: 1em;
   }
 
+  .note-editable li, .answer li {
+    min-height: 1em;
```
meltheadorable: you may also want to set the margin-bottom to `0` here, i don't think list items should have quite that much space between them?

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
```diff
@@ -84,7 +84,8 @@
               = f.input_field :"#{question.attribute_sym}[number]", as: :decimal,
                 value: response.number
 
-            - tags = %w(table tbody tr td b i u ol ul p)
+            - tags = %w(table tbody tr td b i u ol ul li p)
+            = response.text
```
meltheadorable: i believe this is a debugging line, you'll want to remove it

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
```diff
@@ -84,7 +84,8 @@
               = f.input_field :"#{question.attribute_sym}[number]", as: :decimal,
                 value: response.number
 
-            - tags = %w(table tbody tr td b i u ol ul p)
+            - tags = %w(table tbody tr td b i u ol ul li p)
```
meltheadorable: Great catch!

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
arisah-ijeoma: thanks for the find

PR #562, opened 2018-07-31
6997 update step when precedent is late

arisah-ijeoma:


arisah-ijeoma:
Running into intermittently failing spec Should the code modification be done in `scheduled_*_date` instead of `display_*_date` since it affects other parts of the application?  

We need to account for project duplicator, batch_action and date adjustment when it comes to late parents. At the moment, these specs all have parents that are on time

If further work is needed to cover for late parents, I suggest creating new issues that can branch off this one

app/models/project_step.rb
```diff
@@ -358,10 +367,22 @@ def max_descendant_group_depth
 
   def validate_scheduled_start_date
     if schedule_parent && display_start_date != schedule_parent.dependent_step_start_date
```
emitche: `display_start_date` should be replaced with `scheduled_start_date`.

arisah-ijeoma: `dependent_step_start_date` uses `display_end_date` which has also been tweaked for this story

will update specs

emitche:
```
def dependent_step_start_date
    display_end_date.try(:+, 1)
  end
```

We should change this to:

```
  def dependent_step_start_date
    scheduled_end_date.try(:+, 1)
  end
```

PR #563, opened 2018-08-02
8645 add postal code to coop form

arisah-ijeoma:
I have added the field to the coop form

The story said not to make the field required but the chat with Brendan said to make it required for coops in the US division

This PR has it set to not required a the moment


emitche:
Thanks, @arisah-ijeoma. That's the right way to go with not making it required.

emitche:
Postal code is not saving.

PR #565, opened 2018-08-07
7986 add currency names to yml files

arisah-ijeoma:


config/locales/en/en.yml
```diff
@@ -369,6 +369,15 @@ en:
     created: "Created"
     caption: "Caption"
     condense_all: "Condense All"
+    currency:
```
emitche: Let's use one and other here to account for different amounts and to extend this translated functionality.

For example:
```
argentinean_peso:
  one: "Argentinean Peso"
  other: "Argentinean Pesos"
```

There are other parts of the system that use the currency list without translations. We'd like to provide support moving forward for the currency names to be translated into Spanish when a user is viewing any of the sites in Spanish.

For example, this is the loan details form. When selecting the currency and viewing the site in Spanish, the currencies are untranslated.

image

Making changes to the loan details form is not needed in this ticket. We want these translations to work for both singular and plural currency names. That way we can do easy replacements as we see them moving forward (in different tickets).

arisah-ijeoma: This was what I originally did but I figured that it was only used in that one place in the app. Didn't notice the other

Will update right away

app/helpers/currency_helper.rb
```diff
@@ -1,10 +1,11 @@
 module CurrencyHelper
   def format_currency(amount, currency, tooltip: true)
+    currency_name = I18n.t("common.currency.#{currency.try(:locale_name)}")
```
meltheadorable: I think we should use`currency.code` for this instead of this extra `locale_name` method -- the currency codes should already uniquely identify the currencies and not have any spaces/other troublesome characters.

config/locales/en/en.yml
emitche: Thank you!

app/controllers/admin/divisions_controller.rb
```diff
@@ -112,7 +112,9 @@ def index_filter
   end
 
   def prep_form_vars
-    @currency_choices = Currency.order(:name)
+    @currency_choices = Currency.order(:name).map do |currency|
+      [I18n.t("common.currency.#{currency.try(:locale_name)}", count: 1)]
```
emitche: See the note from @meltheadorable for simplifying this.

app/controllers/admin/divisions_controller.rb
```diff
@@ -112,7 +112,9 @@ def index_filter
   end
 
   def prep_form_vars
-    @currency_choices = Currency.order(:name)
+    @currency_choices = Currency.order(:name).map do |currency|
+      helpers.currency_name(currency, count: 1)
```
meltheadorable: I'm not sure about this approach, this feels like a bit of a code smell to me. Maybe we can figure this out next week? This story wasn't scheduled for this sprint anyway.

arisah-ijeoma: ok

found this btw
https://blog.bigbinary.com/2016/06/26/rails-add-helpers-method-to-ease-usage-of-helper-modules-in-controllers.html

meltheadorable:
I think we need to pop this one back for a redesign, I can't really answer the open questions about this offhand, but something feels off to me about this implementation. I'd like to revisit it next sprint.

PR #569, opened 2018-08-10
8744 update conditional

arisah-ijeoma:


app/helpers/media_helper.rb
```diff
@@ -1,13 +1,11 @@
...
   def media_thumbnail(media_item)
-    if media_item.thumbnail?
-      if media_item.caption.text.present?
-        alt_text = truncate(media_item.caption.text, length: 36)
-        return image_tag(media_item.item.thumb.url, alt: alt_text)
-      else
-        # Use auto-generated alt text
-        return image_tag(media_item.item.thumb.url)
-      end
+    if media_item.thumbnail? && media_item.caption
```
emitche: I'd take out `media_item.caption` here. We still want to account for any cases, if they exist, where there is no caption tied to a thumbnail.

It looks like the reason the loan that is broken is failing is because there exists an image that does not have a caption tied to it at all.

app/helpers/media_helper.rb
```diff
@@ -1,13 +1,11 @@
...
+      # Use auto-generated alt text if no caption text
+      return image_tag(media_item.item.thumb.url) unless media_item.caption.text.blank?
```
emitche: We should be returning the image tag as is without a user-generated caption if no caption exists. This current item would only run this line if the caption was not blank, which is the opposite of what we want.

emitche: Maybe use something like this psuedocode:

```
    if media_item.thumbnail?
      if media_item.caption? (and possibly && media.caption.text?)
        alt_text = truncate(media_item.caption.text, length: 36)
        return image_tag(media_item.item.thumb.url, alt: alt_text)
      else
        # For images with no captions
        # Use auto-generated alt text
        return image_tag(media_item.item.thumb.url)
      end
```

app/helpers/media_helper.rb
```diff
@@ -1,13 +1,11 @@
...
+
+      alt_text = truncate(media_item.caption.text, length: 36)
+      image_tag(media_item.item.thumb.url, alt: alt_text)
     else
       content_tag(:div, class: "media-block") do
```
emitche: The else refers to other types of mdia that are not images. So we do want to capture all images (with and without captions) in the first part of the conditional.

app/helpers/media_helper.rb
arisah-ijeoma: 👍

app/helpers/media_helper.rb
emitche: There's different icons for other types of media not images. You can ignore this part.

But any images you test should be displaying as images and not other types of media.

app/helpers/media_helper.rb
```diff
@@ -1,13 +1,10 @@
...
-        alt_text = truncate(media_item.caption.text, length: 36)
-        return image_tag(media_item.item.thumb.url, alt: alt_text)
-      else
-        # Use auto-generated alt text
-        return image_tag(media_item.item.thumb.url)
-      end
+      caption_text_exists = media_item.caption && media_item.caption.text
+
+      alt_text = caption_text_exists ? truncate(media_item.caption.text, length: 36) : nil
+      image_tag(media_item.item.thumb.url, alt: alt_text)
```
emitche: The alt text is getting set to nil here in some cases. However, Rails has some special magic where is takes the file name and turns it into a readable alt tag if not other alt text has been provided. This line doesn't let Rails do it's magic, if that makes sense.

Can you adjust this so that if no caption text exists or if the caption text is an empty string, `alt:` does not get set inside the `image_tag`? That way the alt text can be auto generated when needed.

Thank you!

PR #571, opened 2018-08-15
8764 fix

arisah-ijeoma:


emitche:
This is already part of another ticket that will be merged in develop.

PR #575, opened 2018-08-22
8778 new documentation

arisah-ijeoma:


app/controllers/admin/documentations_controller.rb
```diff
@@ -0,0 +1,33 @@
+class Admin::DocumentationsController < Admin::AdminController
+  include TranslationSaveable
+
+  def new
+    @documentation = Documentation.new(html_identifier: params[:html_identifier])
+    authorize @documentation, :new?
```
meltheadorable: it should already be checking `:new?` when you call the base authorization, and in the `ApplicationPolicy`, `new?` is set as an alias to `create?`, so you should be able to leave this as just `authorize @documentation` like in the `create` action.

app/controllers/admin/documentations_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+      @documentation.calling_action = controller_action[1]
+    end
+  end
+
+  def create
+    @documentation = Documentation.new(documentation_params)
+    authorize @documentation
+
+    if @documentation.save
+      # placeholder till other actions are defined
```
meltheadorable: Any code that we're intending to fix/change later should have comments that start with `TODO:` that way it's easier to find that code later.

arisah-ijeoma: 👍

app/controllers/admin/documentations_controller.rb
arisah-ijeoma: thanks for the catch! forgot to clear this

PR #577, opened 2018-08-23
8779 edit documentation

arisah-ijeoma:
The action buttons for the form look meh. I'm not sure what is causing it. Didn't want to take up more time looking into something I'm not sure of but the functionality works as it should
screen shot 2018-08-23 at 1 59 02 pm


emitche:
We can probably polish the UI during one of the frontend tickets.

meltheadorable:
Unfortunately for this ticket we can't re-use the `show` view for `edit` like we do for some of the other pages. The `show` view needs to render the page content, that also means we don't want to modify the show policy in quite that way, and we need a separate edit action.

PR #582, opened 2018-08-24
8799 show documentation

arisah-ijeoma:


app/controllers/admin/documentations_controller.rb
```diff
@@ -24,11 +26,6 @@ def create
     end
   end
 
-  def edit
```
meltheadorable: doesn't this break the edit page?

app/policies/documentation_policy.rb
```diff
@@ -11,4 +11,8 @@ def edit?
   def update?
     create?
   end
+
+  def show?
+    create?
```
meltheadorable: this is probably the wrong policy for this, but i'll make a separate story to fix the policies

app/controllers/admin/documentations_controller.rb
```diff
@@ -1,6 +1,8 @@
 class Admin::DocumentationsController < Admin::AdminController
   include TranslationSaveable
 
+  before_action :find_documentation, only: [:edit, :show]
```
meltheadorable: `update` should probably be in this list too

app/views/admin/documentations/show.html.slim
```diff
@@ -0,0 +1,5 @@
+- content_for(:title, @documentation.page_title)
+
+section.documentation
+  div.documentation-content
+    = simple_format(@documentation.page_content)
```
meltheadorable: Great!

app/controllers/admin/documentations_controller.rb
meltheadorable: make sure you remove the other find method from update

spec/features/admin/documentation_spec.rb
```diff
@@ -7,7 +7,9 @@
   before do
     login_as user
     doc.summary_content = 'original summary content'
+    doc.page_title = 'original page title'
```
meltheadorable: could you refactor these to use a `let!` for creating the doc?

app/controllers/admin/documentations_controller.rb
arisah-ijeoma: for some reason, this was breaking the page
will take another look

app/controllers/admin/documentations_controller.rb
arisah-ijeoma: nope

app/policies/documentation_policy.rb
arisah-ijeoma: 👍

spec/features/admin/documentation_spec.rb
arisah-ijeoma: 👍

app/controllers/admin/documentations_controller.rb
meltheadorable: maybe try doing `find(params[:html_identifier])` for all of them instead then, but they should all share the same find action one way or another.

PR #583, opened 2018-08-24
8781: Summernote support for translatable

meltheadorable:


app/models/question.rb
```diff
@@ -61,8 +61,8 @@ class Question < ApplicationRecord
   has_closure_tree order: 'position', dependent: :destroy
 
   # define accessor like convenience methods for the fields stored in the Translations table
-  attr_translatable :label
-  attr_translatable :explanation
+  translates :label
+  translates :explanation
```
arisah-ijeoma: this can be in one line

meltheadorable: Some of them are on two lines like this -- in this case that's good, because soon that `explanation` field is going to get an `allow_html` tag but the label won't.

PR #587, opened 2018-08-28
8807 summernote links

arisah-ijeoma:


app/assets/javascripts/summernote-init.js
```diff
@@ -5,7 +5,8 @@ $(document).on('ready', function() {
       toolbar: [
         ["style", ["bold", "italic", "underline", "clear"]],
         ["table", ["table"]],
-        ["para", ["ul", "ol", "paragraph"]]
+        ["para", ["ul", "ol", "paragraph"]],
+        ['insert', ['link']]
```
meltheadorable: could you use double quotes here to match the rest of the list?

arisah-ijeoma: oops

emitche:
QA review: works great.

I am starting to notice some slow down with page updating and loading, but we will handle that outside of this ticket.

Thanks a bunch!

PR #596, opened 2018-09-01
6921 duplicate response sets bug

linesarefuzzy:
I got it working to the point where it catches conflicts and displays the error. Getting the buttons (overwrite, discard) to work was more complicated—Rails' optimistic locking doesn't work on create. After spending about an hour and a half on that, I decided this situation is rare enough that it's probably not worth it. The user can compare the two sets side by side if they need to. If someone else wants to take a crack at getting the buttons working, they're welcome, but I tried several different approaches.


emitche:
QA and functional review: This works great. Thank you!

PR #597, opened 2018-09-03
8845 add specs to cover bugged area

arisah-ijeoma:
I have added specs to cover the scenarios mentioned and also did a manual test. I was not able to recreate this bug.


emitche:
Hmm. Thanks, @arisah-ijeoma!

emitche:
I'm going to merge in the specs. But the bug is still an active one. So let's pair and figure it out together. Thank you so much, @arisah-ijeoma.

PR #599, opened 2018-09-03
8836 language link broken on loan

arisah-ijeoma:


emitche:
I get this bug in Chrome and Firefox.

PR #608, opened 2018-09-11
8846 add summernote to loan questions

arisah-ijeoma:


app/views/admin/questions/_form.html.slim
```diff
@@ -15,14 +15,14 @@
...
+    - I18n.available_locales.each do |l|
       = f.input :"label_#{l}", as: :text,
-          input_html: { data: { translatable: 'questions.text' } },
-          placeholder: I18n.t('questions.text', locale: l),
+          input_html: {data: {translatable: 'questions.text'}},
+          placeholder: t('questions.text', locale: l),
           label: t('questions.text', locale: l)
       = f.input :"explanation_#{l}", as: :summernote,
-          input_html: { id: 'question_explanation', data: { translatable: 'questions.explanation' } },
-          placeholder: I18n.t('questions.explanation', locale: l),
```
emitche: Should the placeholder remain?

config/environments/development.rb
```diff
@@ -51,7 +51,7 @@
   }
 
   # Raises error for missing translations
-  config.action_view.raise_on_missing_translations = true
+  # config.action_view.raise_on_missing_translations = true
```
emitche: We probably want to keep this.

meltheadorable: We don't, I told her to get rid of it. That setting really breaks the transifex workflow, which is supposed to be:

1. Add translations to your primary language
2. Push those translations to the branch
3. Update the additional languages in transifex
4. Download the new transifex files for the non-primary languages and merge them into the project

That breaks down if the app crashes in development every time you don't immediately manually translate all of the fields, and it makes it so we can't offload the translation work to people from the client's org/team.

emitche:
The question form cannot be opened in manage loan questions.

app/views/admin/questions/_form.html.slim
arisah-ijeoma: oh....since it's summernote, I think we can get rid of it

arisah-ijeoma:
this works; needed to restart the server

emitche:
@arisah-ijeoma Will check it out. Thank you!

emitche:
We created another pull request that addresses the changes needed. There may still be work in here that we want to use.

emitche:
Translations were added into develop.

PR #616, opened 2018-09-20
8883 loan featured image

arisah-ijeoma:


app/models/media.rb
```diff
@@ -36,6 +37,8 @@ class Media < ApplicationRecord
 
   validates :item, :kind_value, presence: true
   validate :update_item_error
+  validate :non_image_cannot_be_featured
+  validate :one_featured_image_per_loan
```
meltheadorable: This doesn't really behave like a validation, this should probably be in some kind of a callback or something instead.

app/models/media.rb
```diff
@@ -80,4 +83,17 @@ def error_but_not_item_error?
   def update_item_error
     errors.add(:item, :reattach) if error_but_not_item_error?
   end
+
+  def non_image_cannot_be_featured
+    errors.add(:featured, :non_image) if kind_value != 'image' && featured
+  end
+
+  def one_featured_image_per_loan
+    featured_presently = self.class.where(media_attachable_type: 'Project', media_attachable_id: media_attachable_id).
```
meltheadorable: I think, as written, this will remove the featured status if you change the caption or another field on the item that's already featured. Maybe we should reverse this logic (if the current item is set to featured, use `update_all` to remove the featured flag from all the other media for that attachable type)

arisah-ijeoma: this is genius!

thank you

I was originally trying to skip the loop with validations hence the validate: false option

emitche:
@arisah-ijeoma You are still working on this ticket, correct?

emitche:
The migration cannot be run:

```
$ be rake db:migrate

Rails.env: development
database: madeline_system_development
== 20181106135911 MakeUnfeaturedImagesFeatured: migrating =====================
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

undefined method `update' for nil:NilClass
madeline/db/migrate/20181106135911_make_unfeatured_images_featured.rb:5:in `block in up'
madeline/db/migrate/20181106135911_make_unfeatured_images_featured.rb:3:in `up'
.rbenv/versions/2.4.2/bin/bundle:23:in `load'
.rbenv/versions/2.4.2/bin/bundle:23:in `<main>'

Caused by:
NoMethodError: undefined method `update' for nil:NilClass
madeline/db/migrate/20181106135911_make_unfeatured_images_featured.rb:5:in `block in up'
madeline/db/migrate/20181106135911_make_unfeatured_images_featured.rb:3:in `up'
.rbenv/versions/2.4.2/bin/bundle:23:in `load'
.rbenv/versions/2.4.2/bin/bundle:23:in `<main>'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```

emitche:
I don't know how to fix this migration. I will get help.

emitche:
Thanks for fixing the migration, @meltheadorable.

emitche:
I can't access the print loan memo with these changes. I'll need to look into this further.

```
Can't resolve image into URL: undefined method `to_model' for #<MediaItemUploader:[item]> Did you mean? to_xml
```

emitche:
I'll fix the Hound errors and merge this.

PR #618, opened 2018-10-01
8923: Fixes log validator

meltheadorable:


app/models/concerns/translatable.rb
```diff
@@ -190,9 +190,13 @@ def validate_each(record, attribute, value)
       # Ensure at least one translation given is not blank
       translations = record.send("#{attribute}_translations")
 
-      translations.each do |t|
-        return if t.text.present?
-        record.errors.add("#{attribute}_#{t.locale}", :blank)
+      empty_translations = translations.map do |translation| 
+        return if t.text.present?;
```
arisah-ijeoma: `t` doesn't exist in this block

@meltheadorable

PR #620, opened 2018-10-25
8150: Upgrade gems, rails, and dev goodies

meltheadorable:


emitche:
@meltheadorable I created a new ticket for this work, so that it can be included in the list of tickets included the next time we do a release. https://redmine.sassafras.coop/issues/9021

.eslintrc.yml
```diff
@@ -0,0 +1,105 @@
+env:
+  es6: true
+parserOptions:
+  ecmaFeatures:
+    jsx: true
+plugins:
+  - react
```
DevneyHamilton: Does Madeline have react? or is this a leftover from NEMO?

.eslintrc.yml
```diff
@@ -0,0 +1,105 @@
...
+extends:
+  - eslint:recommended
+  - plugin:react/all
+globals:
+  React: true
+  ELMO: true
```
DevneyHamilton: ELMO looks like a holdover from NEMO as well that we probably don't want

.eslintrc.yml
meltheadorable: some leftovers, yeah, thanks

PR #624, opened 2018-11-16
8704 deposit extractor

DevneyHamilton:


DevneyHamilton:
@meltheadorable This branch has a lot of fixes for 8702 and 8703 as well. Some things weren't clear to me until I could look at all 4 of them together.

app/models/accounting/quickbooks/data_extractor.rb
```diff
@@ -5,8 +5,14 @@ def initialize(object)
         @object = object
 
         case @object.qb_object_type
-        when "JournalEntry", "Purchase", "Deposit", "Bill"
+        when "JournalEntry"
```
DevneyHamilton: Should I use constantize on "Accounting::Quickbooks::#{@object.qb_object_type}Extractor" instead of switch statement?

meltheadorable: I like the explicitness of the switch, and it should keep us from accidentally calling an extractor that doesn't exist, otherwise we'd need to write some other code to check that `qb_object_type` has a matching extractor class before we call it.

PR #627, opened 2018-11-20
9111: enable txns

meltheadorable:


emitche:
@meltheadorable Do you want any code review or anything else on this ticket? I'll be putting out a bug fix release. We can merge this into develop after that, if that works for you.

PR #631, opened 2018-12-07
9028 reduce error messages

emitche:
- Reduce error messages from bots (Redmine 9028)
- Address Pundit Not Authorized errors on the public namespace by showing an error page and error message (Redmine 8688)


emitche:
After making the change to a 401 error, the loan page displays this. Instead of redirecting, there is a link.

image

emitche:
Message shown on pages that are not Loans Show pages:

image

This is a public loan gallery page for a hidden loan.

emitche:
Message shown on public loans show pages:

image

PR #632, opened 2018-12-14
8031 subclasses for transactions

DevneyHamilton:


.travis.yml
```diff
@@ -10,7 +10,7 @@ addons:
 services:
   - postgresql
 before_install: # Install PhantomJS 2.1.1 or some of our specs fail with default (v1.9.8 as of 3/2/2017)
-  - gem update bundler --no-rdoc --no-ri
+  - gem update bundler
```
DevneyHamilton: I did this because the error messages in CI told me to. I'm not sure if there's another way to handle that CI was breaking on those flags or not. (error messages were ```$ gem update bundler --no-rdoc --no-ri
ERROR:  While executing gem ... (OptionParser::InvalidOption)
    invalid option: --no-rdoc``` and ```$ gem update bundler --no-ri
ERROR:  While executing gem ... (OptionParser::InvalidOption)
    invalid option: --no-ri```

emitche: I may add this change into develop sooner, as it's an active issue. Thanks for looking into this.

PR #634, opened 2019-01-23
9328: change :disabled on html_identifier to :readonly so that field …

DevneyHamilton:
…is submitted with form

This seems to be all that is needed based on manual testing. 

Wondering if I should tackle summernote or not - I didn't see the comment and got really bogged down on trying to get those specs working.

### DIFF
--- a/app/views/admin/documentations/_form.html.slim
+++ b/app/views/admin/documentations/_form.html.slim
@@ -2,7 +2,7 @@
 
   = f.error_notification
 
-  = f.input :html_identifier, value: params[:html_identifier], disabled: true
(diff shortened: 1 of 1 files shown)


DevneyHamilton:
The issue turned out to be that having the html_identifier field as "disabled" meant the value was not submitted with the form. Changing this to "readonly" means that value is submitted with the form and everything seems to be working like it should. Another round of QA wouldn't hurt since we don't have specs. 

saved & can view documentation:
screen shot 2019-01-25 at 11 26 35 am
editing existing documentation
screen shot 2019-01-25 at 11 26 50 am

emitche:
Thank you so much! Code looks good. Moving this on to the QA process.

PR #637, opened 2019-02-18
9208: fixes bugs

meltheadorable:


DevneyHamilton:
Thanks for catching the existing classes case in the  transaction builder. Sorry I didn't get it before. Glad it's synced up now! Thank you!

DevneyHamilton:
For posterity: fixes issue where lack of date info caused "negative balance" errors during import, esp affected $0.00 txns. Also fixes finding qb subclasses on existing loans.

PR #639, opened 2019-03-12
9436 handle duplicate name error

DevneyHamilton:


DevneyHamilton:
I manually tested these cases w/ sandbox: a) customer with org name does not exist in qb 2) customer with org name does exist but madeline doesn't have qb id (this shouldn't happen in practice, it just came for free with the refactor) 3) customer name is in qb and madeline already has id 4) madeline does not have a qb id and there is not customer name in qb, but there is a vendor name that matches so qb errors, and we instead create a customer in qb with "(customer)" appended on the name.

DevneyHamilton:
I refactored to look in qb first for a matching customer name before trying to create a customer name, because it made the error handling a lot cleaner, and kept the rescue block to only handling the edge case that there is a vendor with a conflicting name.

meltheadorable:
This looks good except the remaining style errors (for the spaces in braces) -- take care of those and merge away!

PR #641, opened 2019-04-17
9570 add sidekiq

DevneyHamilton:


config/routes.rb
```diff
@@ -58,6 +60,10 @@
       end
     end
 
+    authenticate :user do
```
meltheadorable: We should also require admin permissions for this -- I think the example showed that. It should also be in the admin namespace if it isn't (github is not showing the nexting context here)

DevneyHamilton: it's in admin namespace. I looked around for other examples of authenticating as an admin and didn't find one, and would need some guidance about how to do that. It seems like most permissions are handled with policies.

PR #642, opened 2019-04-24
9614-9616 misc tweaks

meltheadorable:


app/assets/javascripts/summernote-init.js
```diff
@@ -20,7 +20,7 @@ $(document).on('ready', function() {
       height: 150,
       toolbar: [
         ["style", ["bold", "italic", "underline", "clear"]],
-        ["table", ["table"]],
+        ["insert", ["table", "picture", "link", "video"]],
```
DevneyHamilton: will this affect any views that should NOT have picture, link, and video?

config/deploy.rb
```diff
@@ -75,4 +75,15 @@
   end
 end
 
+desc "Update the OptionSets"
+task :update_option_sets do
```
DevneyHamilton: thank you for adding this!

app/assets/javascripts/summernote-init.js
meltheadorable: As far as I'm aware, they should all have all that, but there are two separate summernote initializers and I believe this one is specific to loan questions.

PR #643, opened 2019-04-26
9573 wrap jobs in tasks model

DevneyHamilton:


DevneyHamilton:
Questions to talk about with Mel: 1) add separate field for 'first tried at' and rename :job_started_at to :job_first_started_at to Task? 2) add :num_attempts field to Task? 3) add validations that timestamps make sense? 4) move fwd with :job_dead? field and setting it with a monitor? 5)  do TaskJobMonitor in separate issue, or this one? 6) how to refactor specs, esp TestJob and TestFailureJob?

db/migrate/20190419151240_create_tasks.rb
```diff
@@ -0,0 +1,15 @@
+class CreateTasks < ActiveRecord::Migration[5.2]
+  def change
+    create_table :tasks do |t|
```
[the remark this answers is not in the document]
DevneyHamilton: Do we want created_at and modified_at here?

db/schema.rb
```diff
@@ -397,6 +397,18 @@
     t.index ["name"], name: "index_roles_on_name"
   end
 
+  create_table "tasks", force: :cascade do |t|
+    t.string "activity_message_value", limit: 65536, null: false
+    t.string "job_class", limit: 255, null: false
+    t.datetime "job_enqueued_for_retry_at"
```
DevneyHamilton: job_enqueued_for_retry_at won't really come into play until the TaskJobMonitor happens. I used it for thinking thru statuses tho, and have a placeholder for :stalled. Should I remove this and the :stalled code for now?

app/models/task.rb
```diff
@@ -0,0 +1,71 @@
...
+#  job_last_failed_at        :datetime
+#  job_succeeded_at          :datetime
+#  job_type_value            :string(255)      not null
+#  num_attempts              :integer          default(0), not null
+#  provider_job_id           :string
+#
+
+class Task < ApplicationRecord
+  def enqueue(job_params: {})
+    job = job_class.constantize.perform_later(job_params.merge(task_id: id))
```
DevneyHamilton: Not totally sure about this approach for job params.

DevneyHamilton:
@meltheadorable, Please feel free to do whatever makes sense with this PR while I'm gone - leave it alone, review it, fix it and merge, etc.

app/models/task.rb
```diff
@@ -0,0 +1,71 @@
...
+    elsif in_progress?
+      :in_progress
+    elsif succeeded?
+      :succeeded
+    elsif stalled?
+      :stalled
+    end
+  end
+
+  def start
```
DevneyHamilton: is there a better way to approach naming these 'lifecycle' methods?

spec/factories/task_factory.rb
```diff
@@ -0,0 +1,23 @@
...
+#  job_type_value            :string(255)      not null
+#  num_attempts              :integer          default(0), not null
+#  provider_job_id           :string
+#
+
+FactoryBot.define do
+  factory :task do
+    job_type_value { "monthly_interest_accrual_job" }
+    activity_message_value { "running monthly interest accrual" }
+    job_class { MonthlyInterestAccrualJob }
```
DevneyHamilton: better alternative than an existing job class here?

app/models/task.rb
meltheadorable: I don't have one at the moment, we can always refactor if it's not working out.

db/migrate/20190419151240_create_tasks.rb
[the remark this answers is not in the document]
meltheadorable: Yes, doesn't hurt anything, and could prove useful.

db/schema.rb
meltheadorable: No, it's fine, it'll come into play soon enough.

spec/factories/task_factory.rb
meltheadorable: I still haven't really come up with a good testing approach for this, this is probably okay for now, but we should keep an eye out for alternative techniques.

app/models/task.rb
meltheadorable: Me either, but I don't have a better one, let's just try working with it and see where it leaves us. If it ends up being a pain in the neck to work with in practice, we can revisit. Sometimes you just need to try things.

DevneyHamilton:
@emitche there isn't anything user-facing to QA on this branch separately.

PR #645, opened 2019-05-15
9724 refactor monthly interest accrual job as task

DevneyHamilton:


lib/tasks/madeline.rake
```diff
@@ -0,0 +1,11 @@
+namespace :madeline do
+  desc "Enqueue the interest accrual Task. The whenever gem runs this rake task monthly."
```
DevneyHamilton: What makes this 'monthly' is the `whenever` command that runs it monthly. Also, what the job actually does is is update all the loans that Madeline currently can access. Should we keep current language describing what it's _for_ or update to describe what it actually does?

app/jobs/monthly_interest_accrual_job.rb
```diff
@@ -1,5 +1,5 @@
-class MonthlyInterestAccrualJob < ApplicationJob
-  def perform
+class MonthlyInterestAccrualJob < TaskJob
+  def perform(_job_params)
```
DevneyHamilton: So having to add `_job_params` as an argument to perform for every TaskJob seems to be the main price to pay for the approach to arbitrary job params that's in 9573 (where we have ```def enqueue(job_params: {})
    job = job_class.constantize.perform_later(job_params.merge(task_id: id))
    self.update_attribute(:provider_job_id, job.provider_job_id)
  end```)

DevneyHamilton: args to `enqueue` are optional, but perform has to accept at least one argument if it's not used, because the Task model has to pass the task_id to perform_later. It's used in the ActiveJob callbacks implemented in TaskJob.

lib/tasks/madeline.rake
meltheadorable: I don't understand the question, sorry. Can you rephrase/recontextualize?

DevneyHamilton: resolved: it is in indeed monthly; no reason to change

PR #646, opened 2019-05-20
9725 task monitor

DevneyHamilton:
Ready for review. The sidekiq retry set doesn't tell us when a task was re-enqueued for retry. It tells us IF there will be a retry (boolean), when the last retry was (don't know if this is start or end time of the retry.) Also, it looks like from documentation (https://github.com/mperham/sidekiq/wiki/Error-Handling) that by default it will be days (like 20 days) before a job is declared dead, and on Friday I couldn't find a good way to force a job into the DeadSet for testing. So how do we really want to distinguish between stalled and dead?


app/jobs/task_job.rb
```diff
@@ -0,0 +1,22 @@
+class TaskJob < ApplicationJob
+  before_perform do |job|
+    task_for_job(job).start
+  end
+
+  around_perform do |job, perform_block|
+    perform_block.call
```
meltheadorable: You can cut the `before_perform` if you move `task_for_job(job).start` to the `around_perform` hook before the block call.

app/models/task.rb
```diff
@@ -0,0 +1,69 @@
...
+    elsif in_progress?
+      :in_progress
+    elsif succeeded?
+      :succeeded
+    elsif failed?
+      :failed
+    end
+  end
+
+  def start
```
meltheadorable: Can we make this `start!`

app/models/task.rb
```diff
@@ -0,0 +1,69 @@
...
+    self.update_attribute(:job_first_started_at, Time.current) if self.job_first_started_at.nil?
+    self.increment(:num_attempts).save
+  end
+
+  def end_successfully
```
meltheadorable: `finish!`

app/models/task.rb
```diff
@@ -0,0 +1,69 @@
...
+    self.update_attribute(:job_succeeded_at, Time.current)
+  end
+
+  def record_failure
```
meltheadorable: and `fail!`

config/application.rb
```diff
@@ -40,7 +40,7 @@ class Application < Rails::Application
     config.action_mailer.default_url_options = { host: ENV['MADELINE_HOSTNAME'] }
 
     # Send email on errors
-    unless Rails.env.test?
+    unless Rails.env.test? || Rails.env.development?
```
meltheadorable: Some dev mode configs can get meaningful error emails in this scenario, but if we want to disable this, we should just do if staging/production rather than unless test/development, resolving unless with compound boolean expressions tends to get confusing.

DevneyHamilton: will remove this change

PR #647, opened 2019-05-29
9670: specs for requiring state and postal on organizations with coun…

DevneyHamilton:
…try us added and passing

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -80,7 +80,12 @@ class Organization < ApplicationRecord
   def us_required_fields
     return if !country_is_us?
     return if postal_code.present? && state.present?
-    errors.add(:postal_code, :required_for_us)
+    if postal_code.blank?
+      errors.add(:postal_code, :required_for_us)
+    end
+    if state.blank?
+      errors.add(:state, :required_for_us)
+    end
   end
 
   def country_is_us?
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -58,7 +58,10 @@ class Organization < ApplicationRecord
   validates :name, :division, :country, presence: true
 
   validate :primary_contact_is_member
-  validate :us_required_fields
+  with_options if: ->(org) { org.country.iso_code == "US" } do |us_org|
+    us_org.validates :state, presence: true
+    us_org.validates :postal_code, presence: true
+  end
 
   def loans_count
     loans.size
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -58,6 +58,7 @@ class Organization < ApplicationRecord
   validates :name, :division, :country, presence: true
 
   validate :primary_contact_is_member
+  validate :us_required_fields
 
   def loans_count
     loans.size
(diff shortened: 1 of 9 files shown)


app/models/organization.rb
```diff
@@ -76,4 +77,19 @@ def primary_contact_is_member
...
+    if state.blank?
+      errors.add(:state, :required_for_us)
+    end
+  end
+
+  def country_is_us?
+    us_id = Country.find_by(name: 'United States').try(:id)
+    us_id.present? && country_id == us_id
+  end
 end
```
DevneyHamilton, line 81: Curious if there is a better way to arrange the logic here

DevneyHamilton: I'd been hoping to not need a custom validation, and just validate presence of state and postal_code if `country_is_us?` but the private method isn't available in a conditional validation.

DevneyHamilton:
Screen Shot 2019-05-29 at 12 18 35 PM
These are what the errors look like if you try to save a co-op with country US and a blank postal code and state. The errors show up individually if only one field is missing.

DevneyHamilton:
Screen Shot 2019-05-29 at 12 20 06 PM
US coop with state, missing postal.

DevneyHamilton:
Screen Shot 2019-05-29 at 12 20 44 PM
US coop with postal, missing state

DevneyHamilton:
Screen Shot 2019-05-29 at 12 21 19 PM
Argentina coop saves without postal or state

app/models/organization.rb
meltheadorable: A public method with a call signature like `country?(:us)` might be called for here, then you could use it in conditional evals, I'm not sure if there's a mapping somewhere for ruby symbol of two-letter code -> something you can query the database for, but I'd consider that.

meltheadorable: Alternatively, I'm pretty sure you can use lambdas in conditional validations, so combining that with the grouping syntax something like this

```ruby
with_options if: -> (org) { org.country.iso_code == "US" } do |us_org|
  us_org.validates :state, presence: true
  us_org.validates :postal_code, presence: true
end
```

might work

meltheadorable (commented):
Looks mostly good, and the current solution isn't a dealbreaker, but I think your suspicion we could do it another way was right, [check out this part of the rails docs](https://guides.rubyonrails.org/active_record_validations.html#using-a-proc-with-if-and-unless)

DevneyHamilton:
Updated screenshots with default error message: 

US coop without state or zip
Screen Shot 2019-06-12 at 7 31 59 AM

US coop without zip 
Screen Shot 2019-06-12 at 7 32 13 AM

US coop without state
Screen Shot 2019-06-12 at 7 32 35 AM

US coop with state and zip saves
Screen Shot 2019-06-12 at 7 32 49 AM

non-US coop without state or zip saves
Screen Shot 2019-06-12 at 7 33 20 AM

DevneyHamilton:
The hound violation counted here seems to be stale.

emitche:
@DevneyHamilton Thanks for the updates.

PR #650, opened 2019-06-03
9669 percentage q type

DevneyHamilton:


DevneyHamilton:
Read-only:
Screen Shot 2019-06-05 at 9 05 54 AM


Editing:
Screen Shot 2019-06-05 at 9 06 09 AM

app/assets/stylesheets/admin/questions/_questions.scss
```diff
@@ -78,3 +78,12 @@ section.questionnaires li.jqtree_common .jqtree-element
 .linked-document .help-block {
   margin: .25em 0;
 }
+
+.answer-wrapper.tree-view .percentage-input.form-element {
```
DevneyHamilton: Not loving how fragile this is; need all these selectors to override simple_form styles.

app/inputs/percentage_input.rb
```diff
@@ -0,0 +1,24 @@
+class PercentageInput < SimpleForm::Inputs::NumericInput
```
DevneyHamilton, line 1: Made a custom simple_form input to pull the different markup out of the view file, and have enough control to have the styles we need be moderately less cumbersome & fragile.

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
```diff
@@ -83,9 +83,11 @@ div
               data: {rating: true}
 
           - if response.has_number?
+            - input_type = response.has_percentage? ? :percentage : :decimal
+            - read_only_display_value = response.has_percentage? ? "#{response.number}%" : response.number
```
DevneyHamilton, line 87: should this move into response model?

app/assets/stylesheets/admin/questions/_questions.scss
```diff
@@ -78,3 +78,12 @@ section.questionnaires li.jqtree_common .jqtree-element
 .linked-document .help-block {
   margin: .25em 0;
 }
+
+.answer-wrapper.tree-view .input-with-postfix.form-element {
```
DevneyHamilton, line 82: Expecting to reuse the 'input with postfix' and 'input-postfix' classnames here for currency. Will need to rename to include prefix.

DevneyHamilton: Also, I'm not loving depending on simple_form classes here, but these styles get overriden if I only using custom classes

DevneyHamilton:
this was all combined with 9743 for currency, so this PR does not need to be reviewed or merged.

PR #651, opened 2019-06-05
9743 currency q type and 9669 percentage q type

DevneyHamilton:


app/assets/stylesheets/admin/questions/_questions.scss
```diff
@@ -78,3 +78,16 @@ section.questionnaires li.jqtree_common .jqtree-element
 .linked-document .help-block {
   margin: .25em 0;
 }
+
+.answer-wrapper.tree-view .decorated-numeric-input.form-element {
```
DevneyHamilton, line 82: @emitche do you have ideas about how to improve the maintainability of these styles? We want form-element on here because that's what simple_form uses to hide the input in read-only. But .answer-wrapper and .tree-view are only here to get enough specificity that these styles aren't overridden by simple_form styles.

DevneyHamilton: and/or is there a better place to put this new css?

app/assets/stylesheets/admin/questions/_questions.scss
```diff
@@ -78,3 +78,16 @@ section.questionnaires li.jqtree_common .jqtree-element
...
+  display: inline-flex;
+
+  span.input-prefix {
+    margin-right: 0.5em;
+  }
+
+  span.input-postfix {
+    margin-left: 0.5em;
+  }
+}
```
DevneyHamilton: These are separate bc putting margins on both left and right of something with the same class causes weird indentation when there is no prefix

DevneyHamilton, line 93: tried one class with .5em margin on each side, but having a left margin on prefix makes an indentation that doesn't look right.

app/helpers/responses_helper.rb
```diff
@@ -0,0 +1,26 @@
+module ResponsesHelper
```
DevneyHamilton, line 1: passing the response to each of these methods is similar to existing project_step_helper. Happy for suggestions about making it more concise

DevneyHamilton:
Editing:
Screen Shot 2019-06-05 at 1 53 53 PM

Read only:
Screen Shot 2019-06-05 at 1 52 44 PM

app/assets/stylesheets/admin/questions/_questions.scss
emitche: The placement and specificity is fine.

emitche:
This visually looks just fine. Thank you so much!

app/helpers/responses_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+    if response.has_currency?
+      "#{prefix(response)}#{response.number} #{postfix(response)}"
+    elsif response.has_percentage?
+      "#{response.number}%"
+    else
+      response.number
+    end
+  end
+
+  def prefix(response)
```
meltheadorable, line 13: These don't feel quite right, but I also recognize that the use of helpers here is also just a concession to us not having better options in this codebase yet, so I'd be willing to just keep this whole thing in mind for future refactoring.

meltheadorable (approved):
Looks really good overall, much cleaner than the last attempt. Identified one code smell I don't really have an alternative suggestion for.

app/assets/stylesheets/admin/questions/_questions.scss
DevneyHamilton: thanks for taking a look!

PR #652, opened 2019-06-07
9602: remove all mention of delayed job and drop delayed_jobs table

DevneyHamilton:


DevneyHamilton:
Both Staging and Prod have 0 Delayed::Jobs, so there aren't stuck jobs. Are there more checks I should do to make sure staging is ready for this to be released?

meltheadorable (approved):
YAY!

PR #654, opened 2019-06-11
9572 store changes on txns

DevneyHamilton:


app/models/accounting/transaction.rb
```diff
@@ -131,19 +133,10 @@ def associate_with_qb_obj(qb_obj)
...
-    @change_in_principal ||= net_debit_for_account(qb_division&.principal_account_id)
-  end
-
-  def change_in_interest
-    return 0 unless project
-
-    # See InterestCalculator for more documentation on principal/interest accounts.
-     @change_in_interest ||= net_debit_for_account(qb_division&.interest_receivable_account_id)
+  def calculate_deltas
+    raise Accounting::TransactionMissingProjectError unless project
```
DevneyHamilton, line 137: This is a holdover from the guard clause and comment above (`return 0 unless project`). Should we fix this special case here or in another issue?

DevneyHamilton:
Satisfied with comparing this branch to staging in terms of handling recalculation. Tested with this sequence and verified it's the same on both: https://docs.google.com/document/d/1CVHQ4b-kIIj3uccg6hoWTzaZ-yRFVhpJUIA4ipAxKBk/edit

app/models/accounting/transaction.rb
meltheadorable: Another issue is fine, the last step of our bigger refactor should be a sweep of the code to find dead logic and inefficiencies, we can mop things up then.

meltheadorable (approved):

PR #657, opened 2019-06-18
9740 move qb sync to task

DevneyHamilton:


DevneyHamilton:
Screenshots of different states:

Not connected to QB (same as on develop):
Screen Shot 2019-06-18 at 1 51 56 PM

Connected and latest fetch pending:
Screen Shot 2019-06-18 at 12 28 29 PM

Connected and latest fetch in progress:
Screen Shot 2019-06-18 at 12 48 30 PM


Connected and latest fetch succeeded:
Screen Shot 2019-06-18 at 1 48 31 PM


Connected and latest fetch failed:
Screen Shot 2019-06-18 at 12 29 12 PM

DevneyHamilton:
Seems like the "Last Updated At" Never is a little confusing with the pending and in progress ones. Should I change that? Perhaps we could remove the Last Updated At field and show the time the task succeeded or failed (and no time if pending or in progress?)

app/views/admin/settings/index.html.slim
```diff
@@ -32,6 +32,10 @@ section.block
               = ldate(@last_updated_at, format: :long)
             - else
               = t("quickbooks.connection.never_synced")
+        = f.input "qb_data_import" do
+          .read-only
+            - if @fetch_task.present?
```
DevneyHamilton: if we get here, fetch_task should be present except in cases when there's a qb connection in the db that is not actually still live. I think we're moving towards avoiding those cases, and they are rare and usually in testing situations. There's no `else` logic because that case shouldn't happen, but still need to check to avoid an error that prevents you from resetting.

app/views/admin/settings/index.html.slim
```diff
@@ -46,25 +50,26 @@ section.block
           / this will display a button that the user clicks to start the flow
           ipp:connectToIntuit
 
-- if @accounts && !@accounts.empty?
-  section.block
-    h2 = t("quickbooks.accounts.accounts")
-    p.help-block.note = t("quickbooks.accounts.note")
+- if @division.quickbooks_connected? && @fetch_task && @fetch_task.succeeded?
```
DevneyHamilton, line 54: Only display accounts form when connected and latest fetch task succeeded

DevneyHamilton:
Accounts form visible when latest sync succeeded:
Screen Shot 2019-06-19 at 4 41 44 PM

Accounts form not visible before that:
Screen Shot 2019-06-19 at 3 59 18 PM

app/views/admin/settings/index.html.slim
DevneyHamilton: This change is in latest commit; easy to take out if we should leave as it was for now (accounts form visible when there's the division has a  connection saved in db)

app/models/task.rb
```diff
@@ -47,6 +49,10 @@ def fail!
     self.update_attribute(:job_last_failed_at, Time.current)
   end
 
+  def succeeded?
+    job_succeeded_at.present?
+  end
+
```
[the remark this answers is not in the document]
DevneyHamilton: Git isn't picking up on this change; can fix when I push up revisions on this PR.

emitche:
> Seems like the "Last Updated At" Never is a little confusing with the pending and in progress ones. Should I change that? Perhaps we could remove the Last Updated At field and show the time the task succeeded or failed (and no time if pending or in progress?)

I would move "Last Updated At" beneath "QuickBooks Data Import."

What does the "Last Updated At" date represent? Does it represent the last time the QuickBooks Data successfully imported? Assuming this is correct, I have some questions and suggestions below.

If no sync has ever occured, then this would not be a date, correct?

I would not show "Last Updated" if there has never been a successful import.

If there has been a succesful import before, would showing the "Last Updated" time while an "in progress import" is running be confusing?

I would reorder the labels:

1. QuickBooks Connection
2. QuickBooks Status
3. QuickBooks Data Import
4. QuickBooks Data Last Imported

"QuickBooks Data Last Imported" - could include a tooltip/info circle here that says more - "the date and time QuickBooks Data was last successfully imported."

app/models/task.rb
```diff
@@ -15,8 +15,10 @@
 #  provider_job_id        :string
 #  updated_at             :datetime         not null
 #
-
 class Task < ApplicationRecord
+  scope :full_fetcher, -> { where(job_class: 'FullFetcherJob') }
+  scope :most_recent_first, -> { order("created_at DESC") }
```
meltheadorable: I'd consider doing this with

```ruby
# scope definition
scope :by_creation_time, -> (direction = :asc) { order(created_at: direction) }
# call with
Task.by_creation_time(:desc).full_fetcher
```

for future extensibility

config/locales/en/settings.en.yml
```diff
@@ -15,4 +15,9 @@ en:
       disconnect_message: "Your QuickBooks account has been successfully disconnected."
       data_reset_message: "Your QuickBooks data cache has been reset."
       data_reset_information: "If you're experiencing issues with QuickBooks data, click Reset to clear Madeline's cached copy and restore it from QuickBooks."
-      link_message: "Your QuickBooks account has been successfully linked. Note: You need to select accounts before you can add transactions."
+      link_message: "Your QuickBooks account has been successfully linked. Note: Once data is imported, you will need to select accounts before you can add transactions."
+    fetch_status:
+      failed: "The latest background job for fetching QuickBooks data failed."
```
meltheadorable: I think we can replace all "The latest background job for fetching QuickBooks data" strings with "QuickBooks data import", it's not so much about hiding the implementation details from the user but about using the language for the intended consequence

> QuickBooks data import failed
> QuickBooks data import in progress
> QuickBooks data import pending
> QuickBooks data import succeeded

It's a lot more compact and it frees us up to consider new strategies without changing the user experience.

app/models/task.rb
```diff
@@ -15,8 +15,10 @@
 #  provider_job_id        :string
 #  updated_at             :datetime         not null
 #
-
 class Task < ApplicationRecord
+  scope :full_fetcher, -> { where(job_class: 'FullFetcherJob') }
```
meltheadorable: Should we use `job_type_value` for these?

app/controllers/admin/accounting/quickbooks_controller.rb
```diff
@@ -15,7 +15,13 @@ def oauth_callback
       division: Division.root,
       params: params
     )
-    Accounting::Quickbooks::FullFetcher.new(current_division.qb_division).fetch_all
+
+    Task.create(
+      job_class: FullFetcherJob,
+      job_type_value: 'full_fetcher_job',
+      activity_message_value: 'fetching_quickbooks_data'
+    ).enqueue(job_params: {division_id: current_division.qb_division.id})
```
meltheadorable, line 23: really nice to see this system at work

meltheadorable (commented):
Overall this is really great, excited to see these changes coming together.

meltheadorable:
> > Seems like the "Last Updated At" Never is a little confusing with the pending and in progress ones. Should I change that? Perhaps we could remove the Last Updated At field and show the time the task succeeded or failed (and no time if pending or in progress?)
> 
> I would move "Last Updated At" beneath "QuickBooks Data Import."
> 
> What does the "Last Updated At" date represent? Does it represent the last time the QuickBooks Data successfully imported? Assuming this is correct, I have some questions and suggestions below.
> 
> If no sync has ever occured, then this would not be a date, correct?
> 
> I would not show "Last Updated" if there has never been a successful import.
> 
> If there has been a succesful import before, would showing the "Last Updated" time while an "in progress import" is running be confusing?
> 
> I would reorder the labels:
> 
> 1. QuickBooks Connection
> 2. QuickBooks Status
> 3. QuickBooks Data Import
> 4. QuickBooks Data Last Imported
> 
> "QuickBooks Data Last Imported" - could include a tooltip/info circle here that says more - "the date and time QuickBooks Data was last successfully imported."

Along these lines, I think "Quickbooks Data Import" label should change to "Data Import Status" just to reflect that it's also a status field.

app/models/task.rb
DevneyHamilton: use :full_fetcher, and make a constant TASK_JOB_TYPES in the Task model file. Validation in Task that task_job_type is in that constant

DevneyHamilton:
New labels and order:
Screen Shot 2019-06-25 at 1 50 38 PM

Only display last imported time when latest was successful:
Screen Shot 2019-06-25 at 1 44 10 PM

I left the "Disconnect" line the same for now; we can revisit when we revisit the ux on this page but for now it's an improvement.

emitche:
This looks good. The one non-obvious thing is a user might get this page:

image

To see the update, they need to reload the page:
image

If this UX is improved, it can be in the future. This is a big improvement already and works well.

emitche:
UX improvement for another ticket. The close button on the alert is not obvious due to the proximity of the text to close button.

image

PR #658, opened 2019-06-20
8603 full fetcher deletes cnxn on failure

DevneyHamilton:


app/models/accounting/quickbooks/full_fetcher.rb
```diff
@@ -26,23 +23,40 @@ def fetch_all
         qb_connection.update_attribute(:last_updated_at, started_fetch_at)
 
         restore_accounts!(accounts)
+       rescue StandardError => error
+         delete_qb_data
```
DevneyHamilton: I'm uncertain about doing this much in a rescue block that overlaps with what we're rescuing from. This sort of assumes the error is going to be in the 'fetching' stage.

DevneyHamilton: I can refactor so that rescuing from fetching is separate, and deletes data as well as cnxn.

app/models/accounting/quickbooks/full_fetcher.rb
```diff
@@ -12,23 +12,35 @@ def initialize(division)
 
       def fetch_all
         accounts = clear_accounts!
+        delete_qb_data
+        fetch_qb_data
+        restore_accounts!(accounts)
+      rescue StandardError => error
+        @qb_connection.destroy
+        ExceptionNotifier.notify_exception(error, data: {})
```
meltheadorable: I don't think this exception notifier is necessary, it should be caught by the one in either the job or the application controller.

meltheadorable (approved):
Liking this refactor, good work.

app/models/accounting/quickbooks/full_fetcher.rb
emitche: @DevneyHamilton Do you want to address this before the ticket is merged?

DevneyHamilton: yes, will update. thanks!

DevneyHamilton: updated

emitche:
QA for this in a separate ticket and most likely needs to happen on staging.

PR #660, opened 2019-06-25
9817: update type assignment and account assignment in journal entry …

DevneyHamilton:
…extractor to handle $0.0 debits in repayments


meltheadorable (approved):
In full recognition of the fact that the logic here is messy, I think we should definitely go ahead and merge this if it's working as expected, knowing this is an area of active refactoring.

PR #661, opened 2019-07-01
9805 numbers inside loan memo

emitche:
- Show table for numerical questions in print loan memo


app/models/loan_filtered_question.rb
```diff
@@ -47,4 +47,8 @@ def visible?
   def response_set
     @response_set ||= loan.send(question_set.kind)
   end
+
+  def numerical?
```
meltheadorable: I think you _might_ be able to use `has_number?` on the response object to control this rather than creating a new mapping here, it has a compatible value type.

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
```diff
@@ -84,19 +87,32 @@ div
...
               value: response.number, prefix: prefix(response), postfix: postfix(response)
 
-          - tags = %w(table tbody tr td b i u ol ul li p img a iframe)
 
-          / class attr should work automatically but
-          / adding class attr fixed the bug of having both a table and an image
-          - attrs = %w(class style src data-filename _moz_resizing href target frameborder width height)
 
           - if response.has_text?
+            / Display response as rich text. Specify which html is allowed.
```
meltheadorable, line 103: is the rich text logic supposed to be part of this PR?

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
```diff
@@ -145,5 +161,7 @@ div
           - if !response.group? && response.blank?
             .view-element.answer.blank = t("loan.no_answer")
 
+    - question.numerical? ? @previous_question_numerical = true : @previous_question_numerical = false
```
meltheadorable: see the comment about `has_number?`

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
```diff
@@ -39,7 +39,10 @@ div
           *header_tag
             = question.full_number_and_label
         - else
-          .question-label = question.full_number_and_label
+          // For numerical questions in print view, the question's label is displayed with
+          // the question's answer.
+          .question-label class=(question.numerical? ? 'hidden-print' : '')
```
meltheadorable: see `has_number?`

meltheadorable (approved):
Looking good overall, but we might be able to get rid of some duplication here. 😄

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
emitche: No. That's an older feature from many months ago. I just added a comment for context. I can remove, if it is not needed.

app/models/loan_filtered_question.rb
emitche: I'll look into `has_number?`.

app/views/admin/loans/questionnaires/_questionnaire_group.html.slim
emitche: The `tags` and `attrs` were scoped outside of the if statement to which they belonged. I simply moved those into the if statement and tweaked the comments in the section.

emitche:
**QA Notes - Things That Were Changed**

In print loan memo, questions that have a number (percentage, currency, and number) are displayed in tables. Adjacent number reponses visually display as one larger table.

If a number question is marked not applicable or is empty, it should not display in the print loan memo. That question's absence in print loan memo should not break the flow of adjacent number reponses in print loan memo displaying as one combined table. To test this, for 3 consecutive number questions, answer the first and last. Do not answer the middle question or mark it as not applicable.

In the the Buisness Planning tab for a loan (such as http://localhost:3000/admin/loans/4/questions?filter=criteria), no change in the UI should be visible to the user in both the view and edit mode. A number question displays with its heading and either its response (view mode) or an input box (edit mode).

DevneyHamilton:
QA - works for me! (skipped qs s13 and 16)
Screen St 2019-07-03 at 1 50 26 PM

PR #662, opened 2019-07-01
9870 no 0 amount line items

DevneyHamilton:
Followup from 9817, in progress

### DIFF
--- a/app/models/accounting/interest_calculator.rb
+++ b/app/models/accounting/interest_calculator.rb
@@ -152,6 +152,7 @@ module Accounting
             end
             pp "INTEREST CALCULATOR"
             pp tx.line_items
+            check_credits_equal_debits(tx, prev_tx)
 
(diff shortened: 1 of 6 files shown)


DevneyHamilton:
Screenshots for example in redmine issue; focused on Sept 26 $1 repayment, which was added before  an earlier-dated disbursement that raised the interest balance just before this repayment. 
If  `calculate_balances` balances happens after the transaction is saved in `recalculate` in the interest calculator, we have the 'old' split of the repayment, which is not correct because the interest balance is higher than the repayment amount:
Screen Shot 2019-07-03 at 9 55 46 AM
If `calculate_balances` happens BEFORE the transaction is saved in `recalculate` in the interest calculator, we have the 'old' amount for interest_receivable, and $0.00 for prin acct which is not correct because 83 cents is lost (that should go to interest receivable)
Screen Shot 2019-07-03 at 9 55 17 AM

PR #665, opened 2019-07-12
9904: use render_translation to handle rich text log details and notes

DevneyHamilton:


DevneyHamilton:
Using the existing render_translation method in the TranslationsHelper seems to take care of carriage returns and tables for logs. Here are screenshots from testing:

'Legacy' log with carriage returns that did not display correctly before this fix:
Screen Shot 2019-07-12 at 4 17 30 PM

Table that did not display correctly with 'simple_format' directly called in the view.
Screen Shot 2019-07-12 at 4 17 22 PM

emitche:
The code looks fine. I'm looking over the UI.

PR #666, opened 2019-07-16
9668 number separators

emitche:
- Add thousand separators for number, currency, and percentage loan question answers


emitche:
**QA**
To test: For a question type that is number, currency, or perentage, add a number greater than 1000. You can also test with decimals.

Strange existing UI quirks that may be bugs:
A number entered with commas cannot be saved. A number entered with a decimal point can be saved.

app/helpers/responses_helper.rb
```diff
@@ -2,11 +2,11 @@ module ResponsesHelper
   def display_value_for_number(response)
     return if !response.has_number?
     if response.has_currency?
-      "#{prefix(response)}#{response.number} #{postfix(response)}"
+      "#{prefix(response)}#{format_number(response.number)} #{postfix(response)}"
```
meltheadorable: any particular reason not to just use `number_with_delimiter` directly?

meltheadorable (commented):
Quick question, if the answer is yes, this is good to merge.

app/helpers/responses_helper.rb
emitche: Only if we need to support locales that use `.` in place of `,`. We'd just need to pass in the additional option. but I like your approach to just use `number_with_delimiter` directly.

PR #668, opened 2019-07-23
9903 rich text logs emails

emitche:
- Ensure log emails have both plain text and html/rich text templates
- Fixes bug where raw HTML was showing in emails being sent


meltheadorable (approved):
This looks like a good solution, thanks!

emitche:
This will need QA on staging to confirm this is working well in a live setup.

PR #671, opened 2019-08-08
9876: update documentations behavior such that a user may see documen…

DevneyHamilton:
…tations that belong to their division or any of its descendents or ancestors but may only edit a documentation if any of the user's divisions match the its division or any of its ancestors

This is a bit tedious to test. The desired behavior is:
A documentation popover is displayed if it belongs to the current division or one of its ancestors (this matches previous behavior). 
A user may VIEW a documentation that belongs to any division that is accessible to them or an ancestor of their division (this is a change from before, see Redmine issue). 
A user may EDIT only documentation that belongs to a division where it or one of its ancestors is either the user's base division or one of its descendents (this is a change from before, see Redmine issue). 
A user will see 'Learn more' link only for documentations they can view (this is a change from before, see Redmine issue).

I manually tested this with root division, financial coop (parent is root), and BRED (parent is financial coop), Buen Vivir, and CIELO (parent is Buen Vivir) on my local, for the 'no' scenarios here:
Screen Shot 2019-08-08 at 2 29 20 PM

### DIFF
--- a/spec/policies/documentation_policy_spec.rb
+++ b/spec/policies/documentation_policy_spec.rb
@@ -1,0 +1,37 @@
+require 'rails_helper'
+
+describe DocumentationPolicy do
+  subject { described_class.new(user, record) }
+
+  let!(:parent_division) { create(:division) }
+  let!(:division) { create(:division, parent: parent_division) }
+  let!(:child_division) { create(:division, parent: division) }
+  let(:user) { create(:user, :member, division: division) }
+
+  context "documentation belongs to user's division" do
+    let(:record) { create(:documentation, division: division) }
+
+    it { should permit_all }
+  end
+
+  context "documentation belongs to ancestor of user's division" do
+    let(:record) { create(:documentation, division: parent_division) }
+
+    it { should permit_action(:show) }
+    it { should forbid_action(:edit) }
+  end
+
+  context "documentation belongs to descendant of user's division" do
+    let(:record) { create(:documentation, division: child_division) }
+
+    it { should permit_action(:show) }
+    it { should permit_action(:edit) }
+  end
+
+  context "documentation belongs to a division that is not a descendant or ancestor of any of the user's divisions " do
+    let(:other_division) { create(:division) }
+    let(:record) { create(:documentation, division: other_division) }
+    it { should forbid_action(:show) }
+    it { should forbid_action(:edit) }
+  end
+end
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -72,7 +72,7 @@ module AdminHelper
     placement = options[:placement] || 'right'
     if documentation.present?
       data_content = documentation&.summary_content.to_s
-      if documentation.page_content.present?
+      if documentation.page_content.present? && policy(documentation).show?
         learn_more_link = link_to t("documentation.learn_more"), admin_documentation_path(documentation), target: :_blank
         data_content << "<br /><br />" << learn_more_link
       end
--- a/app/policies/documentation_policy.rb
+++ b/app/policies/documentation_policy.rb
@@ -1,6 +1,12 @@
 class DocumentationPolicy < ApplicationPolicy
   def show?
-    # show if any of the user's divisions match the record's division or any of its ancestors
+    # a user may see documentations that belong to any of their divisions or any ancestors of their division
+    user_divisions = user.accessible_division_ids + user.division.self_and_ancestor_ids
+    user_divisions.include?(record.division.id)
+  end
+
+  def update?
+    # allow edit if any of the user's divisions match the record's division or any of its ancestors
     user_divisions = user.accessible_division_ids
     record_divisions = record&.division&.self_and_ancestor_ids
     intersection = (user_divisions & record_divisions)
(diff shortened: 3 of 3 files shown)


meltheadorable (approved):
fab!

spec/policies/documentation_policy_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    it { should permit_action(:show) }
+    it { should permit_action(:edit) }
+  end
+
+  context "documentation belongs to a division that is not a descendant or ancestor of any of the user's divisions " do
+    let(:other_division) { create(:division) }
+    let(:record) { create(:documentation, division: other_division) }
+    it { should forbid_action(:show) }
+    it { should forbid_action(:edit) }
+  end
```
DevneyHamilton, line 36: I started trying to write a spec for the case that documentation belongs to an ancestor of another division the user has access to (besides the `user.division`). Then I looked for examples of users having access to divisions other than the one they are assigned to, and couldn't find them. The DivisionPolicy `base_accessible_ids` method looks for any division where the user is a member or admin. Is there an example of a user having a role on a division that they aren't assigned to? Does that happen in practice?  This is what I started writing: 

```context "documentation belongs to an ancestor of another division the user has access to" do
    let(:other_parent_division) { create(:division) }
    let(:other_division) { create(:division, parent: other_parent_division) }
    let(:record) { create(:documentation, division: other_parent_division) }
    # give user a role on `other_division` while keeping user.division `division`

    it { should permit_action(:show) }
    it { should permit_action(:edit) }
  end```

File: spec/policies/documentation_policy_spec.rb [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+
+    it { should permit_action(:show) }
+    it { should forbid_action(:edit) }
+  end
+
+  context "documentation belongs to descendant of user's division" do
+    let(:record) { create(:documentation, division: child_division) }
+
+    it { should permit_action(:show) }
+    it { should permit_action(:edit) }
```
DevneyHamilton, line 28: Do we really want people to be able to edit documentation on descendant divisions?

spec/policies/documentation_policy_spec.rb
DevneyHamilton: I'm generally a little confused about how :admin and :member roles should play into edit permissions here, but maybe that's not something we need to tackle in this issue. If we do, it will involve revisiting the use of `base_accessible_id`, which doesn't distinguish between the two roles. We could have edit permissions on documentation use `division_admin` like many other resources, instead, but I am assuming there was some good reason for members to be able to edit documentations and/or for documentations to have their permissions own logic.

spec/policies/documentation_policy_spec.rb
meltheadorable: Yes, but not on ancestors.

spec/policies/documentation_policy_spec.rb
meltheadorable: I think we should copy the show logic from the question policy and just leave edit at the default for the application policy for now. This is not a high stakes permissions issue, but I think that brendan is looking for behavior similar to the question policy here and it might be easier to just reuse it rather than continuing to go to trouble trying to sort out the right logic. I don't think the question policy's logic is written correctly for what it's trying to do, but it does seem to work okay for now, and there's no way the client wants us spending much more extra time on this.

Don't bother with the test case for giving a user a role outside of their tree, the basics are good enough here, lets not overthink it.

DevneyHamilton: ok, great - the question logic lines up with the specs I had in there, and I like removing the more complex code from the documentation policy a lot.

emitche:
Sorry it took so long to QA this.

## Results of QA and its impact on the PR
This is a good step. It allows documentation to be seen in the expected way. The work here fixes a bug and should be kept in `develop`. This addresses the initial reported bug.

There still appears to be some work needed on permissions for creating and editing documentation. The work completed does not introduce new bugs that do not already exist. Therefore the below work described can go into a future sprint. And the work done so far should be released.

## Future Work
Adjustments are needed on permissions for creating and editing documentation.

On staging, there is a Child Division which is the child of Test Division.

Root Division
- Test Division
-- Child Division

### Documentation cannot be created for child divisions if it exists on an ancestor
If the documentation for Test Division exists, let's say for the first item on the dashboard, documentation cannot be created for the same documentation item in the Child Division. While using an admin for Child Division, that user is unable to edit the documentation for the first Dashboard item because it already exists on a higher ancestor.

I believe the itent is for the lowest level descendant to be able to have as fine-grained documentation as desired (for viewing and editing/creating documentation.). They should also inherit from the closest ancestor with documentation for viewing documentation if no documentation yet exists.

At this time, there is no way for a child division to overwrite the documentation of an ancestor. 

### Documentation cannot be created on ancestor divisions if it exists on a descendant
If documentation exists on a descendant, but not an ancestor, the ancestor will show up as having no documentation. But that documentation will also not be editable.

If documentation exists for a child division, documentation cannot be created on a higher level ancestor. For example, I added documentation for the first item on the Logs page for Child Division (from the Child Division admin account). While using a different user, I cannot add documentation for that item inside Test Division.

image

Documentation items cannot be saved on the root division if any child has documentation. This is the message show when trying to create documentation on the dashboard's first info bubble.

image

### Documentation cannot be deleted
There is also no way from Madeline to delete documentation. Therefore, I cannot test what happens when the documentation for Test Division's first dashboard item is removed. I don't know if I would then be able to create documentation on Child Division while Child Division is the active division.

PR #673, opened 2019-08-28
10009 loan update task

DevneyHamilton:


lib/tasks/madeline.rake
```diff
@@ -8,4 +8,14 @@ namespace :madeline do
       activity_message_value: 'updating loans'
     ).enqueue
   end
+
+  desc "for testing update loan task"
```
DevneyHamilton: should this task be added to the whenever schedule.rb file? or will it be incorporated later?

DevneyHamilton:
Task show page with error details is below. The loan id is a link to the loan page.

Screen Shot 2019-08-28 at 12 14 43 PM


Task index page with link to see error details:
Screen Shot 2019-08-28 at 12 16 40 PM

app/jobs/update_all_loans_job.rb
```diff
@@ -0,0 +1,23 @@
+class UpdateAllLoansJob < TaskJob
+  def perform(job_params)
+    task = Task.find(job_params[:task_id])
+    errors_by_loan = []
+    loans = Loan.all
+    updater = Accounting::Quickbooks::Updater.new
+    task.update(activity_message_value: "syncing_with_quickbooks")
+    updater.qb_sync_for_loan_update
+    task.update(activity_message_value: "updating_all_loans")
```
DevneyHamilton: I wanted to show how many loans, but unfortunately the value is a key in i18n so I don't know how to do that without adding more complexity to the task model

app/jobs/update_all_loans_job.rb
```diff
@@ -0,0 +1,23 @@
+class UpdateAllLoansJob < TaskJob
+  def perform(job_params)
+    task = Task.find(job_params[:task_id])
+    errors_by_loan = []
+    loans = Loan.all
+    updater = Accounting::Quickbooks::Updater.new
+    task.update(activity_message_value: "syncing_with_quickbooks")
+    updater.qb_sync_for_loan_update
```
DevneyHamilton, line 8: I needed to move `qb_sync_for_loan_update` to its own method so that I could call it just once from the job. Including it in the loop over all the loans made my local take over 30 mins to process 1500 loans. Changing this made it more like 8 mins for 1500 loans.

lib/tasks/madeline.rake
meltheadorable: We will want it to run daily, I think, so yeah it should be added to the schedule. I'm not sure how long a typical run should take, it should run basically as often as possible with some padding on staging. If it typically takes about 5 minutes it should run every 15m, if it takes 30m it should run every hour on staging, if it takes 2 hours it should run every 3 or every 4, etc -- in prod it should run in the middle of the night, lets call it 2am.

app/jobs/update_all_loans_job.rb
meltheadorable: This is probably fine, but we could probably find a way to let us pass arguments into these, if we gave the task model an `activity_message_data` that contains key-value pairs for the I18n interpolations `{ current_loan: 1, total_loans: 1500 }` we could use a custom method on the model in the views:

```ruby
def activity_message=(message_value, data = nil)
  self.activity_message_value = message
  self.activity_message_data = data
end

def activity_message
  I18n.t(activity_message_value, activity_message_data)
end
```

and then `task.activity_message` should interpolate when there are interpolations -- the reason for the other assignment method is to make sure that when you're assigning messages you are clearing out the data from the previous message, should prevent some bugs.

meltheadorable (approved):
This looks great!

DevneyHamilton:
More specific information about quickbooks errors and a generic activity message if 'failed.':
Screen Shot 2019-08-30 at 3 59 54 PM

Activity message reports num loans processed so far out of total:
Screen Shot 2019-08-30 at 1 58 57 PM

DevneyHamilton:
I still need to update the rake task naming and add to schedule.rb, then will re-open PR. new code for activity message and error handling would benefit from a review

config/schedule.rb
```diff
@@ -7,9 +7,13 @@
   runner 'RecalculateAllLoanHealthJob.perform_later'
 end
 
+every 1.day, at: '2am' do
+  rake "madeline:enqueue_update_loans_task"
+end
+
```
DevneyHamilton, line 13: As far as I understand, `schedule.rb` just updates the crontab at deploy. If we want different frequency on staging versus production, perhaps we'd need to have different schedule.rb files per environment?

app/jobs/update_all_loans_job.rb
```diff
@@ -0,0 +1,39 @@
...
+      rescue StandardError => error
+        errors_by_loan << {loan_id: loan.id, message: error.message}
+        next
+      end
+    end
+    task.update(custom_error_data: errors_by_loan)
+    task.set_activity_message("finished_with_custom_error_data")
+  end
+
+  rescue_from(Accounting::Quickbooks::NotConnectedError) do |error|
```
DevneyHamilton, line 22: is there a better way to combine these two rescue_from overrides?

meltheadorable: Maybe, but unless we start adding a third/fourth error that we want to rescue from with very similar code, I'm not sure it's worth putting the energy into figuring it out, this is a very mild form of code duplication

DevneyHamilton: ok, thanks

config/schedule.rb
emitche: According to @meltheadorable, not a blocking concern.

PR #674, opened 2019-08-28
8587: spike that demonstrates displaying the company name on the sett…

DevneyHamilton:
…ings page


DevneyHamilton:
This is just a method that is added to the Connection class. The method connects to the QB CompanyInfo service. It's very fragile, and just displays "Unknown" if it fails. Only one company comes back from the api query, and it's the company we're currently connected to. See below that it shows "Madeline Staging Test Company" when I'm connected to that, and "Test Company Backup" when connected to that. 

Questions:
Should the CompanyInfo service and query be moved elsewhere? Into a PORO? 
Is it a problem that this is an additional syncronous QB API call when the settings page is loading?
Should we do more with an error? 
Should we do more with testing?
Should this wait til after OAuth2? 


Screen Shot 2019-08-28 at 4 14 30 PM
Screen Shot 2019-08-28 at 4 13 53 PM
Screen Shot 2019-08-28 at 12 16 40 PM

app/models/accounting/quickbooks/connection.rb
```diff
@@ -59,4 +59,19 @@ def auth_details
   def access_token
     Accounting::Quickbooks::Consumer.new.access_token(token: token, secret: secret)
   end
+
+  def company_name
+    begin
+      company_info_service = ::Quickbooks::Service::CompanyInfo.new(self.auth_details)
+      query_result = company_info_service.query("select * from CompanyInfo")
```
DevneyHamilton, line 66: I tried querying by realm_id, which is the 'Id' of the company, but that field is not queryable.

meltheadorable (approved):
This is fine for now, if OAuth2 changes things we can worry about it then. Worst case scenario we revert this capability, best case scenario it's obvious how we could refactor this to be cleaner or more reliable.

emitche:
Good to go. Just waiting on the tests to pass before merging.

PR #675, opened 2019-08-30
10010: adds taskable

meltheadorable:


app/models/task.rb
```diff
@@ -8,16 +8,24 @@
...
 #
+# Indexes
+#
+#  index_tasks_on_taskable_type_and_taskable_id  (taskable_type,taskable_id)
+#
+
 class Task < ApplicationRecord
   TASK_JOB_TYPES = %i(full_fetcher)
 
+  belongs_to :taskable, polymorphic: true
```
DevneyHamilton, line 27: Is this going to come into use with the "Standard Loan Data Export" story? Asking to check my own understanding of how these will fit together.

DevneyHamilton (approved):
This all looks good to me. I had one question for my own understanding.

app/models/task.rb
meltheadorable: No, this will likely be used for the first time in the DataExportTask story -- its only job is to allow creating a link between a task and another type of object, this is mostly going to be useful for creating links in the UI to go back and forth between a task and the associated data -- StandardLoanExport is just creating the `process_data` method that fetches the data and creates the 2D Array.

PR #676, opened 2019-08-30
10011: data export model, base controller

meltheadorable:


app/models/data_export.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  private
+
+  def set_name
+    export_type_key = DATA_EXPORT_TYPES.invert[self.type.to_s]
+
+    self.name ||= I18n.t(
+      "data_exports.default_name",
+      type: I18n.t("data_exports.types.#{export_type_key}"),
+      current_date: I18n.l(Time.zone.now, format: :short)
```
DevneyHamilton: Looks like this will display current_time, (not just date), which is good. Should current_date be renamed to current_time?

app/controllers/admin/data_exports_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  def new
+    @data_export = @export_class.new
+  end
+
+  private
+
+  def set_export_class
+    @export_class = DATA_EXPORT_TYPES[params[:export_type]].constantize
+  end
+end
```
DevneyHamilton, line 13: cool!

DevneyHamilton (approved):
All looks good to me. One suggestion to consider changing 'current_date' to 'current_time'

app/models/data_export.rb
meltheadorable:
```ruby
farore:~/p/s/madeline (10012_export_to_csv|✔)> rails c
Rails.env: development
database: madeline_development
Loading development environment (Rails 5.2.3)
[1] pry(main)> I18n.l(Time.zone.now, format: :short)
=> "Aug 30, 2019"
```

app/controllers/admin/data_exports_controller.rb
meltheadorable: If it works! This can't really be tested properly until we start subclassing.

app/models/data_export.rb
DevneyHamilton: ah, thank you

PR #677, opened 2019-08-30
10012: adds to_csv

meltheadorable:


app/models/data_export.rb
```diff
@@ -26,14 +26,30 @@ class DataExport < ApplicationRecord
...
 
   before_save :set_name
 
   DATA_EXPORT_TYPES = {
     "data_export" => "DataExport"
   }
 
+  def to_csv!
+    raise ArgumentError, "No data found" unless data.present?
+    raise TypeError, "Data invalid" unless (data.is_a?(Array) && data.first.is_a?(Array))
```
DevneyHamilton: Should the message specify that data should be a 2D array? or is it expected a developer will look here and see that from the code?

spec/models/data_export_spec.rb
```diff
@@ -24,8 +24,45 @@
 
 require 'rails_helper'
 
-RSpec.describe DataExport, type: :model do
+describe DataExport, type: :model do
   it "has a valid factory" do
     expect(build(:data_export)).to be_valid
   end
+
```
DevneyHamilton, line 31: thanks for these specs!

DevneyHamilton (approved):
This looks good to merge unless you want to address the hound suggestions and/or make the 'invalid data' error more specific.

app/models/data_export.rb
meltheadorable: that `data` needs to be a 2D array is part of the assumptions of this API, that should probably be documented somewhere, but in practice nobody should be running into this error because nobody should be writing code to add non-2d-array data to that field, so the message is not a huge deal

PR #681, opened 2019-09-24
10169: Remove hidden spacing from translations

emitche:
- Fix bug where OBJ boxes showed in business model canvas

### DIFF
--- a/config/locales/en/loans.en.yml
+++ b/config/locales/en/loans.en.yml
@@ -92,14 +92,14 @@ en:
     transaction_schedule: "Schedule"
 
     business_model:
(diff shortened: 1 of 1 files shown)


meltheadorable:
Was this some weird transifex side effect or something? Looks good regardless, just curious how these ended up like this.

meltheadorable (approved):

emitche:
It seems like whoever copied in text into the yml file copied in some text with invisible indents/spacing. I don't think it was a Transifex issue because the Spanish version of the content was fine. It was a strange bug.

PR #682, opened 2019-09-24
10016 new data export (includes `10017 create data export)

DevneyHamilton:


config/routes.rb
```diff
@@ -14,6 +14,9 @@
     resources :basic_projects, path: 'basic-projects'
     resources :calendar, only: [:index]
     resources :calendar_events, only: [:index]
+    scope '(:type)' do # parens lets this param be optional
```
DevneyHamilton: I could not find a way to have the param come after data_exports AND have path helpers etc work

app/controllers/admin/data_exports_controller.rb
```diff
@@ -1,13 +1,47 @@
...
+  def set_export_class_on_new
+    @export_class = DataExport::DATA_EXPORT_TYPES[data_export_new_params[:type]].constantize
+  end
+
+  def data_export_new_params
+    params.permit(:type)
+  end
+
+  def data_export_create_params
+    params.require(:data_export).permit(:type, :division_id, :locale_code, :name)
```
DevneyHamilton: form helper creates this nexted structure for params, where as the named url param for :type in routes, has it as a first-level param. Happy for suggestions about how to streamline this.

app/models/data_export.rb
```diff
@@ -57,15 +58,18 @@ def to_csv!
     save!
   end
 
-  private
-
-  def set_name
+  def set_default_name
```
DevneyHamilton: Name had been set in a before_save. The goal of moving it to this public method is for it to be set and available to show in the `new` form.

app/views/admin/data_exports/_form.html.slim
```diff
@@ -0,0 +1,32 @@
...
+    defaults: { \
+      input_html: {class: 'form-element form-control'},
+    },
+  ) do |f|
+
+  = error_notification(f)
+  = f.input :type, as: :hidden
+  = f.input :division_id, as: :hidden
+  = f.input :name
+  = f.input :start_date
```
DevneyHamilton, line 19: I have an issue where the top of the calendar is hidden behind the header bar. @emitche have you run into this and do you have any suggestions for getting the calendar to be on top?
Screen Shot 2019-09-24 at 11 54 18 AM

app/models/accounting/quickbooks/updater.rb
```diff
@@ -39,6 +39,7 @@ def update(loans = nil)
       #
       def qb_sync_for_loan_update
         raise NotConnectedError unless qb_connection
+        raise AccountsNotSelectedError unless Division.root.qb_accounts_selected?
```
DevneyHamilton: in `division.rb` the method `qb_division` seems to be set up to handle the qb connection being on a non-root division. Hopefully we can sort this out with some refactoring soon?

DevneyHamilton:
Screen Shot 2019-09-24 at 11 54 00 AM
Screen Shot 2019-09-24 at 11 54 10 AM

app/controllers/admin/data_exports_controller.rb
meltheadorable: this is probably fine

app/models/data_export.rb
meltheadorable: Yeah, in my original design the form included a preset dropdown or something for type, so parts of the name weren't resolvable when the form was rendered. I think I'd prefer a `default_name` method here that's public, show it as the placeholder for the input (instead of prefilling the input) and then have `set_name` remain in the `before_save` as `self.name ||= default_name`

config/routes.rb
meltheadorable: We discussed this and I sent you an alternate route definition we can use here. 

```ruby
resources :data_exports, path_names: { new: '(/:type)/new' }
```

meltheadorable (approved):
Looking good!

app/models/data_export.rb
```diff
@@ -57,15 +59,22 @@ def to_csv!
...
       "data_exports.default_name",
       type: I18n.t("data_exports.types.#{export_type_key}"),
       current_time: I18n.l(Time.zone.now, format: :long)
     )
   end
+
+  private
+
+  def set_name
+    self.name = default_name if self.name.blank?
```
DevneyHamilton, line 80: Have to check blank because

a = ""
a ||= "alternative"
puts a

results in ""

DevneyHamilton:
For QA: 
Setup -  sidekiq running and quickbooks connected
go to admin/data_exports/new
Click "Standard Loan Data Export Link"
Should see placeholder name, start and end dates with datepickers, and a prefilled dropdown of locale code options. 
Filling out the form and clicking "create data export" should result in no error in the console. However, since the show view has no content at this point (its in issue 10018), the page will remain as is.

DevneyHamilton:
Currently start date is the 3rd field on the form, after name and locale_code. 

Notes on date-picker placement for start date when the start date is the second field on the form:

the default vertical orientation puts it 'up' and the top of the calendar is covered by the header bar. 
Bottom orientation (defaults to bottom-left) has the calendar overlapped by the 'previously entered values' drop down, at least in chrome. 
Bottom-right orientation is functional, but inconsistent with other date-picker experiences in Madeline. Here's a screenshot: 
Screen Shot 2019-09-25 at 9 32 02 AM

emitche:
Start and end date do not work yet on the report.

DevneyHamilton:
I updated the spec and code to save the dates. I also merged 10018 in so I could update the data_export_flow_spec to check for rendering after create. This makes the branch comparisons a little weird.

emitche:
@DevneyHamilton I need to make revisions to the work from 10018 to update the naming to use "data exports" and not "reports." I'll make the changes on the 10018 branch and then we can integrate the changes.

emitche:
I need to make small 10018 revisions, but otherwise this one is good to go.

DevneyHamilton:
QA notes:

Prerequisites: loans are updated. To update loans manually, make sure quickbooks is connected and run:
`bundle exec rake madeline:enqueue_update_loans_task` QB has to be connected for the update loan task to run.

Running the data export should NOT require that qb be connected. 

Need to open CSV file to confirm:
- coop and loan data make sense
- sum of disbursements, sum of repayments, change in interest, change in repayments have values for most US loans that are active, completed or liquidated
- try different date ranges and check that are different sum of disbursements, sum of repayments, change in interest, change in repayments

At this point, with real prod data, it is not possible to get a successful report because some loans have errors.

DevneyHamilton:
I updated the messages
From  “Data export created successfully” to "Data export initiated" as the 'create success' message for data exports. I think users will assume the data export is the file, and the file is not created yet. 

From  “Loan has transactions missing data. Do you need to run a QB update?” to "Loan has transactions missing data. Either there were errors updating the loan, or the loan needs to be updated." This is more accurate and it would be annoying to see "do you need to run a qb update?" when you already have but the error is still there.

emitche:
Thanks for the updates and additional information, @DevneyHamilton.

emitche:
@DevneyHamilton There is 1 failure on the specs for this. There may be other things that need be fixed once QA is completed.

emitche:
**Question 1:** For all loans with transaction data, for a report over all of time, I am seeing the following. Is this expected?

* change in interest is 0
* change in repayments is 0

**Question 2:** If a loan was not given a custom name, it does not have a name in the loan name field. This might be expected. Is this expected?

All other fields appear to be getting populated correctly from the data and make sense.

emitche:
Using different date ranges on reports shows different amounts in the sum of disbursements and sum of repayments fields. Change in interest and change in repayments remain at 0.

emitche (commented):
QA Review:

@DevneyHamilton 

I have some questions about what I am seeing in the data export. Let me know if what is there is expected.

There is a failing test that needs to be addressed.

DevneyHamilton:
> **Question 1:** For all loans with transaction data, for a report over all of time, I am seeing the following. Is this expected?
> 
> * change in interest is 0
> * change in repayments is 0
> 
> **Question 2:** If a loan was not given a custom name, it does not have a name in the loan name field. This might be expected. Is this expected?
> 
> All other fields appear to be getting populated correctly from the data and make sense.

Q1: no, it's not expected that change in interest and change in repayments is 0. I got various values for US loans that are active, liquidated or complete on my local. We should look at this tomorrow. 

Q2: Hmm I didn't think about that at all - I assumed loans have names. It's an accurate export of the data, so I think it's fine. TWW can rely on ID, and can add names if they want, or request we make a default name if they need that?

DevneyHamilton:
> > **Question 1:** For all loans with transaction data, for a report over all of time, I am seeing the following. Is this expected?
> > 
> > * change in interest is 0
> > * change in repayments is 0
> > 
> > **Question 2:** If a loan was not given a custom name, it does not have a name in the loan name field. This might be expected. Is this expected?
> > All other fields appear to be getting populated correctly from the data and make sense.
> 
> Q1: no, it's not expected that change in interest and change in repayments is 0. I got various values for US loans that are active, liquidated or complete on my local. We should look at this tomorrow.
> 
> Q2: Hmm I didn't think about that at all - I assumed loans have names. It's an accurate export of the data, so I think it's fine. TWW can rely on ID, and can add names if they want, or request we make a default name if they need that?

UPDATE ON QUESTION  1 for @emitche :
I tried running the data export with for all time on staging after both connecting to qb, and manually running the update_all_loans task. This time, it DID have values for change_in fields. (The file is at https://ms-staging.theworkingworld.org/admin/data-exports/8) Theoretically, being connected to qb or not should not affect the data export so we should definitely investigate that. But for now, it would make sense to double-check that you can also get values for the change_in fields for several US coops on staging. If so, then I think this PR is set. Thanks so much for the QA!!!

PR #683, opened 2019-09-25
8369 gem updates

meltheadorable:


emitche (commented):
The updated gems look good.

@meltheadorable Do you know why the tests failed? Is that expected?

PR #684, opened 2019-09-27
10019: add and pass spec for allowing root admins to do anything with…

DevneyHamilton:
… reports, and disallow anyone else from doing anything with reports

### DIFF
--- a/app/policies/data_export_policy.rb
+++ b/app/policies/data_export_policy.rb
@@ -1,9 +1,17 @@
 class DataExportPolicy < ApplicationPolicy
+  def index?
+    division_admin(division: Division.root)
+  end
+  
+  def show?
+    division_admin(division: Division.root)
+  end
+
   def new?
-    true
+    division_admin(division: Division.root)
   end
 
   def create?
-    true
+    division_admin(division: Division.root)
   end
 end
--- a/app/policies/data_export_policy.rb
+++ b/app/policies/data_export_policy.rb
@@ -2,7 +2,7 @@ class DataExportPolicy < ApplicationPolicy
   def index?
     division_admin(division: Division.root)
   end
-  
+
   def show?
     division_admin(division: Division.root)
   end
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -23,6 +23,8 @@
 #
 
 class DataExport < ApplicationRecord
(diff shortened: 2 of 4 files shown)


app/policies/data_export_policy.rb
```diff
@@ -1,9 +1,17 @@
 class DataExportPolicy < ApplicationPolicy
+  def index?
+    division_admin(division: Division.root)
+  end
+
```
DevneyHamilton: To pass the specs below, I only need to define ` index?`  here. I put the other actions in too so the intention is visible in the policy, and they don't depend on other code. The other specs pass without the code because Application Policy has `show`, `create`, and `new` all checking `division_member_or_admin` and the data_export factory has the division set to root_division.

app/policies/data_export_policy.rb
```diff
@@ -1,9 +1,17 @@
...
```
DevneyHamilton, line 5: To pass the specs below, I only need to define `index?` here. I put the other actions in too so the intention is visible in the policy, and they don't depend on other code. The other specs pass without the code because `Application Policy` has `show`, `create,` and `new` all checking `division_member_or_admin` and the `data_export` factory has the division set to `root_division`.

DevneyHamilton:
(also asked in slack but putting here to go with 'question' label) Issue 10019's description covers making sure that any links to reports index/show/new/create etc check authorization before displaying, which I haven't done because I think those links will be added in 10018. The other parts (the comment, policy) are done and ready for review. Should I mark 10019 ready for review, or wait til 10018 is in and I can update 10019 on Monday?

DevneyHamilton:
I can merge 10018 into this once its reviewed, and then add a check on the site menu link to only display the reports link if the user is a root division admin.

DevneyHamilton:
> I can merge 10018 into this once its reviewed, and then add a check on the site menu link to only display the reports link if the user is a root division admin.

Elizabeth and I worked together on this - 10018 is updated with the check, which should take effect when 10019 is merged.

meltheadorable (approved):

emitche:
Tests were all passing before adding in most recent 10016 branch.

emitche:
Waiting on tests and then will merge this into 10016.

emitche:
The same tests as 10016 failed.

PR #685, opened 2019-09-27
10018 data export index

emitche:


emitche:
Let me know if I should remove commented out content. I've included it, as we will likely use it later.

emitche:
Visually, the changes look like this:

image

A menu link has been added:

image

meltheadorable (approved):

config/locales/en/data_exports.en.yml
```diff
@@ -1,6 +1,14 @@
 en:
+  data_export:
+    attachments: "Attachments"
+    file: "File %{n}"
+    status: "Export Status"
+    task: "Associated Task"
   data_exports:
     default_name: "%{type} on %{current_time}"
+    index_heading: "Reports"
```
meltheadorable, line 9: Can we call these data exports? I want to be consistent, and I think "reports" sets expectations about more sophisticated functionality than we're providing.

PR #686, opened 2019-09-30
10020 data export show

emitche:


config/locales/en/data_exports.en.yml
```diff
@@ -1,9 +1,13 @@
 en:
   data_export:
     attachments: "Attachments"
+    info: "Report Information"
```
meltheadorable: Like in the other one, let's consistently call these "Data Exports"

meltheadorable (approved):

emitche:
The same tests are failing on the 10016 branch.

PR #687, opened 2019-09-30
10191 better data export errors

DevneyHamilton:
These are tweaks related to 10015 that I did later seperately.

### DIFF
--- a/app/jobs/data_export_task_job.rb
+++ b/app/jobs/data_export_task_job.rb
@@ -5,8 +5,7 @@ class DataExportTaskJob < TaskJob
       task_for_job(self).set_activity_message("data_export_in_progress")
       ::DataExportService.run(data_export)
     rescue DataExportError => e
-      task_for_job(self).update(custom_error_data: e.child_errors)
-      task_for_job(self).set_activity_message("finished_with_custom_error_data")
+      task_for_job(self).add_errors(e.child_errors)
     end
   end
 end
--- a/app/jobs/data_export_task_job.rb
+++ b/app/jobs/data_export_task_job.rb
@@ -2,11 +2,11 @@ class DataExportTaskJob < TaskJob
   def perform(job_params)
     data_export = DataExport.find(job_params[:data_export_id])
     begin
+      task_for_job(self).set_activity_message("data_export_in_progress")
       ::DataExportService.run(data_export)
     rescue DataExportError => e
       task_for_job(self).update(custom_error_data: e.child_errors)
       task_for_job(self).set_activity_message("finished_with_custom_error_data")
-      raise e
     end
   end
 end
--- a/app/jobs/update_all_loans_job.rb
+++ b/app/jobs/update_all_loans_job.rb
@@ -15,8 +15,7 @@ class UpdateAllLoansJob < TaskJob
         next
       end
     end
-    task.update(custom_error_data: errors_by_loan)
-    task.set_activity_message("finished_with_custom_error_data")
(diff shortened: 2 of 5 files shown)


app/jobs/data_export_task_job.rb
```diff
@@ -2,11 +2,11 @@ class DataExportTaskJob < TaskJob
   def perform(job_params)
     data_export = DataExport.find(job_params[:data_export_id])
     begin
+      task_for_job(self).set_activity_message("data_export_in_progress")
       ::DataExportService.run(data_export)
     rescue DataExportError => e
       task_for_job(self).update(custom_error_data: e.child_errors)
       task_for_job(self).set_activity_message("finished_with_custom_error_data")
-      raise e
```
DevneyHamilton, line 9: raising this error makes the task appear to have failed, when in fact the csv export is available.

meltheadorable: I'm still not sure if that's desirable behavior or not. Should we create a new status for tasks, something like "succeeded with errors" or "partially succeeded"? I want to make sure we also don't call exports that have problems unambiguous successes.

meltheadorable (approved):

DevneyHamilton:
"Completed with errors" status:
Screen Shot 2019-09-30 at 5 02 10 PM

DevneyHamilton:
I think these updates are an improvement, but the usage of "custom_error_data" overall might be something to think about doing a clean-up pass on next sprint?

DevneyHamilton:
If I remember right, in the last conversation we had about it, @meltheadorable and I agreed that we needed to take a more planned, thoughtful approach to 1) representing a 'partial failure' and 2) reporting errors that are currently in the custom_error_data field on Tasks. I expect this to also include Mel's fix for making sure accounting errors in data exports only show up when accounting is hooked up.  Is the next step for Melody to do planning on this? I'm also happy to plan together, or take a stab at it, just want to get on the same page about what our next move is.

meltheadorable:
> If I remember right, in the last conversation we had about it, @meltheadorable and I agreed that we needed to take a more planned, thoughtful approach to 1) representing a 'partial failure' and 2) reporting errors that are currently in the custom_error_data field on Tasks. I expect this to also include Mel's fix for making sure accounting errors in data exports only show up when accounting is hooked up. Is the next step for Melody to do planning on this? I'm also happy to plan together, or take a stab at it, just want to get on the same page about what our next move is.

I think it might make sense to have you try to take a stab at planning this one

First step would be to enumerate some of the properties of our ideal solution, so for instance, we know we want

1) Someplace that multiple different types of tasks can find and report errors on a loan
2) To not double-report the same error for the same loan when it's encountered by two different tasks

I'd try to identify any other requirements and constraints, we can chat about them a bit, and then maybe pair on ways it might be possible to make that happen. It's okay to produce a laundry list in this first step.

My other patch so that errors only show up when accounting is connected is possibly better solved another way, and we can ignore it if we aren't just going to shove it into a patch to make production less noisy.

DevneyHamilton:
Ok, thanks @meltheadorable for the direction there. I will work on that this afternoon, like 3ish (or 2 if I don't do our T2Coach mtg), and can talk with you more then, tomorrow, or Friday before noon ET

UPDATE I am meeting with Lena and pairing with Tom at 3, so it'll be 3:30 or 4 before I'm around. We can look at it together tomorrow or Friday, or early next wk.

DevneyHamilton:
This led to additional planning and addressing errors rather than improving reporting

PR #689, opened 2019-10-03
10208 fix status and activity messages

DevneyHamilton:
This PR covers the following corrections:

1) Before this PR, the quickbooks import task showed activity message "Fetching Quickbooks data " even after it succeeded. Now it's activity message is "completed" if the task succeeds. 

2) Before this PR, the data export task still showed its activity message as "This task has been enqueued" after it had succeeded. Now its activity message is "complete" if the task succeeds. 

3) UpdateAllLoansTask was showing its status as "succeeded" when it had child errors for individual loans. Now it shows its status as "Failed" in this case.

### DIFF
--- a/app/jobs/data_export_task_job.rb
+++ b/app/jobs/data_export_task_job.rb
@@ -2,11 +2,13 @@ class DataExportTaskJob < TaskJob
   def perform(job_params)
     data_export = DataExport.find(job_params[:data_export_id])
     begin
+      task_for_job(self).set_activity_message("in_progress")
       ::DataExportService.run(data_export)
     rescue DataExportError => e
       task_for_job(self).update(custom_error_data: e.child_errors)
       task_for_job(self).set_activity_message("finished_with_custom_error_data")
       raise e
     end
+    task_for_job(self).set_activity_message("completed")
   end
 end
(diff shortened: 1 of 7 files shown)


DevneyHamilton:
Status value fixes: 
Screen Shot 2019-10-03 at 12 15 11 PM

DevneyHamilton:
The goal of this PR is to correct confusing or misleading statuses and activity messages. 

The reason I added a spec for the data export task job is that I think some the status and activity messages that I'd manually tested for one task got out of sync in later issues for other tasks. Rather than set up another manual test for the 'succeed' case for the data export task, which would require some thinking and rejiggering on my local env, I went with a unit test. 

We'll want to refactor the way the TaskJobs are handling errors. In this PR I focused on trying not to introduce new weirdness. I did continue patterns that we'll probably want to refactor.

meltheadorable (approved):

emitche:
QA:

I can QA item 2 on my local environment. That looks fine.

For item 1, I'll need to check that out on staging.

For item 3, I need to check that out on staging. I can't run that task on my local as it is QB dependent.

* When the Unknown company was connected, error received was "Task failed with an internal error. An administrator has been notified."
* When I disconnected QB and ran the task, I got: "Error: Quickbooks not connected"

PR #692, opened 2019-10-09
10193: update accounting setup instructions to reflect changes in int…

DevneyHamilton:
…uit and madeline UIs

### DIFF
--- a/documentation/accounting/SETUP.md
+++ b/documentation/accounting/SETUP.md
@@ -30,7 +30,7 @@ If any documentation is missing in this guide for setting up QuickBooks with Mad
 1. A dashboard for your project will appear after creating the app.
 
 ### Open QuickBooks sandbox
-1. Inside your project dashboard, click on `QuickBooks sandbox` under the `Resources` section.
+1. Go to Sandbox (under Account dropdown). It may take a moment to load.
 1. You will see a sandbox company. Click `Go to company`.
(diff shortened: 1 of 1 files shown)


DevneyHamilton:
@emitche I reset my local db (according to accounting setup instructions, and tested with Sandbox Company 1 under the madelineqb@sassafras.coop Intuit Developer account

I was able to
- connect to the sandbox company
- create transactions for a loan that was created in the sample data
- edit a transaction in qb and get the changes back when refreshing Transactions page for the loan
- manually run the Update All Loans Task. This reported a success, and no errors in the sidekiq log. It has a 'message' reporting errors, but that's a bug that I can fix in 10191. 

Should I work on a more elaborate set up for testing accounting on local? Right now, we have to recreate any accounting data manually when we reset the db. If the sandbox co already has loans txns, they will only come through to Madeline local if the loan transactions in qb happen to match (by loan ID, I'm pretty sure) loans in the reset db.

emitche:
@DevneyHamilton Sorry I am just seeing this. I will take a look.

emitche:
@DevneyHamilton From what you have described here, this sounds good. I will be sending you a Slack thread with things I encountered when trying to test a couple of weeks ago. Thank you so much!

emitche (commented):
Changes look good here.

PR #701, opened 2019-11-06
10263: add rake task to make one-time changes on loan date fields per…

DevneyHamilton:
… client request

### DIFF
--- a/lib/tasks/one_time_changes.rake
+++ b/lib/tasks/one_time_changes.rake
@@ -1,0 +1,13 @@
+namespace :one_time_changes do
+  desc "A one time task responding to Nov 2019 request to update loan date fields."
+  task adjust_loan_dates: :environment do
+    Loan.all.each do |l|
+      if l.actual_end_date.nil? && l.projected_end_date && l.projected_end_date < Time.zone.today
+        new_actual = l.projected_end_date
+        new_projected = nil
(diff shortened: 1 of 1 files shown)


DevneyHamilton:
Wrote this with confused future devs in mind; happy for any feedback.

DevneyHamilton:
manually tested on local copy of prod db.

lib/tasks/one_time_changes.rake
```diff
@@ -0,0 +1,13 @@
+namespace :one_time_changes do
+  desc "A one time task responding to Nov 2019 request to update loan date fields."
+  task adjust_loan_dates: :environment do
+    Loan.all.each do |l|
```
meltheadorable: I like to use `find_each` in cases like this, IIRC there's some kind of performance benefit to doing so.

meltheadorable (approved):

lib/tasks/one_time_changes.rake
DevneyHamilton: thanks!

DevneyHamilton:
After this change is deployed to staging, we'll need to manually run `bundle exec rake one_time_changes:adjust_loan_dates RAILS_ENV=staging`. It will also need to be run manually on production after we verify it works on staging.

PR #703, opened 2019-11-22
10268 qb customer table

DevneyHamilton:


DevneyHamilton:
Most of the time on this was setting up the manual testing loop. I done a minimal approach so far, matching and reusing the existing fetcher logic for other quickbooks models, and coding was very quick. If this is a good opportunity to do some refactoring on fetchers, I'm happy to do that as part of this issue. 

The Accounting::Customer model created here appears to be redundant of the pre-existing Accounting::Quickbooks:Customer model. It was faster to create a parallel from scratch. The old Accounting::Quickbooks:Customer can go away in the following issue (10276), because it will not longer be used when building qb transactions.

DevneyHamilton:
Need to also update in updater (on task and via txns page)

DevneyHamilton:
This can be qa'ed thru the UI once the customer drop-down is added to the transaction creation in an upcoming issue. I manually tested by checking the value of `Accounting::Customer.count` via the rails console.  I checked the current value, then created a new customer in the sandbox co, then reconnected via the settings page to test the `full_fetcher` and confirmed the value had increased by one.  I did the same thing with reloading a loan transactions page (intead of reconnecting on the settings page,) to test,  since that uses the `updater`.

app/models/accounting/quickbooks/customer_old.rb
```diff
@@ -4,7 +4,7 @@
 # for a link to this object in a transaction or other QB object.
 module Accounting
   module Quickbooks
-    class Customer
+    class CustomerOld
```
meltheadorable: Is this still needed?

meltheadorable (approved):

app/models/accounting/quickbooks/customer_old.rb
DevneyHamilton: no, thanks for catching!

PR #704, opened 2019-11-25
9943: Adds OAuth2 Support, removes legacy OAuth

meltheadorable:


app/models/accounting/quickbooks/consumer.rb
```diff
@@ -6,33 +6,23 @@ module Quickbooks
...
-          site: 'https://oauth.intuit.com',
-          request_token_path: '/oauth/v1/get_request_token',
-          authorize_url: 'https://appcenter.intuit.com/Connect/Begin',
-          access_token_path: '/oauth/v1/get_access_token'
-        )
+        @oauth_consumer = OAuth2::Client.new(oauth_consumer_key, oauth_consumer_secret,
+          site: "https://appcenter.intuit.com/connect/oauth2",
+          authorize_url: "https://appcenter.intuit.com/connect/oauth2",
+          token_url: "https://oauth.platform.intuit.com/oauth2/v1/tokens/bearer")
       end
```
DevneyHamilton, line 20: In general I try to steer clear of hardcoding urls like this, but I'm assuming there's a good reason to do it in this case.

app/models/accounting/quickbooks/updater.rb
```diff
@@ -38,7 +39,7 @@ def update(loans = nil)
       # too far in the past for the `since` method to access.
       #
       def qb_sync_for_loan_update
-        raise NotConnectedError unless qb_connection
+        raise NotConnectedError unless @connected
```
DevneyHamilton, line 42: Thank you for changing this to check if connected!

DevneyHamilton (approved):
Thank you so much @meltheadorable ! I know a GREAT GREAT DEAL of work went into this behind the scenes, and it's sooo awesome that the code here is clear & concise :D

emitche:
Thanks for the code review, @DevneyHamilton.

PR #708, opened 2019-12-03
9887 loan memo summary page

emitche:
- Display loan summary questions and groups in print loan memo page


app/models/question_set.rb
```diff
@@ -78,6 +78,10 @@ def question(question_identifier, required: true)
     result
   end
 
+  def summary_questions
```
emitche: @meltheadorable I don't like this particular method. If you have a suggested improvement here, please let me know. The code only needs to check if there are questions with the display_in_summary attribute set to `true`. The code is not using that list of questions to populate any content.

app/views/admin/loans/questionnaires/_questionnaire_toc.html.slim
```diff
@@ -1,4 +1,6 @@
 ul
+  - if @response_set.question_set.summary_questions.length > 0
```
emitche: @meltheadorable See earlier comment about refactoring `summary_questions`.

app/models/question_set.rb
meltheadorable: There may be another option here involving scopes, but if this is really only checking whether they exist, you should be able to do

```ruby
def summary_questions?
  questions.where(display_in_summary: true).count > 0
end
```

which should do a much smaller/faster query, and would avoid you needing to do the length check later in the view.

app/views/admin/loans/questionnaires/_questionnaire_toc.html.slim
meltheadorable: See other comment.

meltheadorable (approved):

app/models/question_set.rb
emitche: Thank you!

emitche:
QA completed in 9886

PR #714, opened 2019-12-20
10347: Only display summary once inside table of contents

emitche:


emitche:
Before change:

image

After change:

image

PR #716, opened 2019-12-20
10157 remove monthly interest accrual job

DevneyHamilton:


DevneyHamilton:
The monthly interest accrual task was missing a translation and failing (issue 10157). It did almost the same thing as the update all loans task. The UpdateAllLoans job updated ALL loans including inactive ones, no matter the division and no matter if the division was connected to qb. the MonthlyInterestAccrual job updated all active loans whose division is the division connected to qb or their descendants. This PR removes the MonthlyInterestAccrual job entirely, and updates the UpdateAllLoans job, which is run daily, to update all active loans whose division is the division connected to qb or their descendants.

meltheadorable (approved):

PR #717, opened 2019-12-20
10228: check for existing qb connection in rake task before enqueuing…

DevneyHamilton:
… task. Keep check that connection is current within the job.


DevneyHamilton:
This issue goes with 10157 in that it does NOT make any changes to the MonthlyInterestAccrual rake task because that task is removed in 10157. 

This PR just checks for the existance of a qb connection for a division in the madeline task. The more detailed check that the connection has current credentials etc is still in the UpdateAllLoansJob inside the updater. This way the rake task won't enqueue the  won't run on prod while accounting isn't set up, but we'll get a nice loud error in the Tasks panel if accounting SHOULD be set up and isn't. This is also holding in mind we may be refactoring for multiple qb connections in the future, and not wanting to mess a lot with connection-related code before that.

meltheadorable (approved):

PR #720, opened 2020-01-01
10360: raise negative balance error only on managed (admin created in…

DevneyHamilton:
… madeline) txns


DevneyHamilton:
In working on this, I found that it looks like the negative balance error was originally designed to only appear when creating a txn in Madeline (in the modal). It was being raised anytime txns were updated, however, and the message  "the balance on this date is __" doesn't make any sense when loading the transactions page for a loan or when updating transactions in the background (as a user, you don't know what date or txn it's referring to). The change in this PR is to only raise the negative balance error on managed transactions, in the modal, where the error message saying "the balance on this date is __" makes sense. Theoretically, a person could create a managed txn in Madeline, and then change it in quickbooks such that it would raise a negative balance error when updating txns another way, but that seems like an edge case, and probably not worth designing around until we get the transactions updating disentangled from the transactions controller.

DevneyHamilton:
I investigated the negative balance errors on active loans that had been showing up in the loan updates (loans 2589 and 2894). The negative balances were coming from txns created in QB. I have detailed notes (https://docs.google.com/document/d/1f4q5dmtu-10O8aM3EX85x8kB8eIQGHJchRU6L1jx5TU/edit) to share with TWW if they have questions about these two loans, but concluded that these should be allowed for now unless/until TWW requests we detect those cases and report them.

DevneyHamilton:
Negative Balance error showing when user is creating a txn in madeline that would result in negative balance: Screen Shot 2020-01-01 at 4 27 21 PM

DevneyHamilton:
There is still one loan on staging that is erring with a negative balance on an ms-managed txn (2894) when updating from qb. I need to look into that one more, but I think this little change is still an improvement.

meltheadorable (approved):

PR #721, opened 2020-01-05
10368: check for found line item when looking for credit or zero debi…

DevneyHamilton:
…t amount


DevneyHamilton:
There were loans failing to import correctly because an unexpected account was used, and the credit line item was not findable. The je extractor should assign :other in this case, not error. 

Detailed notes about loans that had this error on staging: https://docs.google.com/spreadsheets/d/14Fqrl_dMVhw7tgO996VFOiJSayXUNBGvX3dcUMF9JVk/edit#gid=0

DevneyHamilton:
The journal entry extractor was failing on these loans when it didn't find a credit to the expected account when in fact it should just assign :other type  to the txn: 
985
1259
1372
2507
2524
on staging.

The error was in line 45 of the je extractor.

See https://docs.google.com/spreadsheets/d/14Fqrl_dMVhw7tgO996VFOiJSayXUNBGvX3dcUMF9JVk/edit#gid=0 for details about which txns and accounts were not processable due to the error.

meltheadorable (approved):

emitche:
This can be QAed on staging, if needed.

DevneyHamilton:
QA: check txn page for  a couple of (loans 985
1259
1372
2507
2524) and confirm they do not error.

emitche:
Thank you. I will do this for QA.

PR #722, opened 2020-01-09
10297 timeline batch actions

emitche:
- Fix broken actions on smaller screens for timeline header
- Fixes timeline batch actions on smaller screens


emitche:
Small CSS change to fix a bug. There's an additional ticket to address UX improvements that will be added as a pull request.

PR #723, opened 2020-01-09
10373 timeline header

emitche:
- Adjust timeline header so that it is more responsive across screen sizes


emitche:
TODO: 

Tweak the visual here:

image

Additionally, switch over to col-n-12 for items that will wrap onto the next line because the previous item will take up more than 6 columns.

meltheadorable (approved):

PR #725, opened 2020-01-10
10298 duplicate step modal

emitche:
- Adjust duplicate step modal styling


emitche:
Updated duplicate step modal:

image

emitche:
Current duplicate step modal:

image

meltheadorable (approved):

emitche:
It is possible for one of the inputs to cause a line that should be one line to break to two lines:

image

emitche:
Related bug: If "Repeat" "End" option is selected, but not filled out, the form can be submitted. The step will not duplicate, as is the correct action.

image

PR #729, opened 2020-02-07
10438 respect closed books date

DevneyHamilton:


app/models/accounting/interest_calculator.rb
```diff
@@ -205,10 +234,20 @@ def month_boundaries(d1, d2)
     end
 
     def add_int_tx?(txs, prev_tx)
-      return true if txs.nil?
-      if txs
-       return true if prev_tx && prev_tx.principal_balance > 0 && txs.none?(&:interest?)
+      if txs.nil? # this is an end of month day with no txns
```
DevneyHamilton, line 236: This method seems to have too much nesting. I think we will want to know the difference between "no end of month int txn before closed books date" vs "there was a txn that wasn't paired with an interest txn before closed books date." I think the "no end of month int txn" could get really annoying and we could remove it if so; need to try it out with real data and see how common it is.

DevneyHamilton:
@meltheadorable Question about testing here. The interest calculator spec has an elaborate set up (probably necessary) and elaborate expectations. The bluntest way to test closed book dates would be to copy all those existing specs, set a closed books date after those txns dates, and they are not updated. Should I try to wrangle the existing specs into a 'shared example' approach, and run them once with a closed book date before, and once with a closed book date after the txns' dates? or should I take this oppty to try to simplify the specs overall? or something else?

DevneyHamilton:
@meltheadorable also could you please review the changes so far and let me know if they seem to be on track? 

resolved - this is ready for full review now

DevneyHamilton:
> @meltheadorable Question about testing here. The interest calculator spec has an elaborate set up (probably necessary) and elaborate expectations. The bluntest way to test closed book dates would be to copy all those existing specs, set a closed books date after those txns dates, and they are not updated. Should I try to wrangle the existing specs into a 'shared example' approach, and run them once with a closed book date before, and once with a closed book date after the txns' dates? or should I take this oppty to try to simplify the specs overall? or something else?

I think I have this figured out.

DevneyHamilton:
> @meltheadorable Question about testing here. The interest calculator spec has an elaborate set up (probably necessary) and elaborate expectations. The bluntest way to test closed book dates would be to copy all those existing specs, set a closed books date after those txns dates, and they are not updated. Should I try to wrangle the existing specs into a 'shared example' approach, and run them once with a closed book date before, and once with a closed book date after the txns' dates? or should I take this oppty to try to simplify the specs overall? or something else?

resolved - I added specs

DevneyHamilton:
For QA:  I should try this with staging data on local first - it could blow up with a lot of errors showing for loan update etc.

meltheadorable (approved):
This looks great

emitche:
image

emitche:
Adding a closed book date and removing it results in an error on the update all loans task.

image

`if prev_tx.txn_date > @closed_books_date`

Update: This is a valid bug that will need to be addressed. This can be addressed outside of this pull request.

emitche:
Update: Resolved

After adding back in a closed book date, I get: 

`unknown attribute 'custom_data' for Accounting::ProblemLoanTransaction.`

related to this line:

`          ::Accounting::ProblemLoanTransaction.create(loan: loan, accounting_transaction: prev_tx, message: :attempted_new_int_txn_before_closed_books_date, custom_data: {cbd: @closed_books_date, txn_date: prev_tx.txn_date}, level: :warning)
`

PR #730, opened 2020-02-13
10455: fix incorrect method call in customer creation

DevneyHamilton:


DevneyHamilton:
For QA: should be able to create a first txn for a coop that does NOT have an Accounting::Customer record in Madeline that matches the organization name yet, and not get error about expecting 0 arguments and getting 2. 

QA w sandbox: find (or create) a coop/organization whose name does not match any of the Customers in the Customer drop-down in the txn form. Create a new loan for this coop. Add a first disbursement txn. It should succeed. After it's saved, look at it in read-only view and make sure the Customer has been set to one matching the coop name (Madeline should create the new qb customer when it creates the txn). 

QA on staging: loans 2955 and 2947 had this error when running loan updates Feb 23  (https://ms-staging.theworkingworld.org/admin/tasks/175). They should not have this error after this is deployed to staging.

Attn: @emitche

emitche (approved):

emitche:
QA: 

With real data, I was able to duplicate the original error. A transaction could not be created when no customer was set. The result would be an error in the console.

I tested the update. The transaction cannot be created, but the flow is cleaner. You see an error in the add transaction modal instead of the modal closing this time.

image

PR #731, opened 2020-02-23
10454 problem loan txn warnings

DevneyHamilton:


DevneyHamilton:
replaced with pr 732, that has `10438` as base  instead of `develop`

PR #732, opened 2020-02-23
10454 problem loan txn warnings

DevneyHamilton:


app/controllers/admin/accounting/transactions_controller.rb
```diff
@@ -29,6 +29,18 @@ def create
     @loan = Loan.find(transaction_params[:project_id])
     authorize(@loan, :update?)
     @transaction = ::Accounting::Transaction.new(transaction_params)
+
+    # the txn model must be able to save managed txns before
```
DevneyHamilton: I didn't love adding this logic to a controller action. I tried to figure out a way at the model validation level to distinguish between an imported qb txn and a txn that had been created via the Madeline txn form. I didn't find a way that always works (e.g. we have to be able to create 'Managed' txns before the closed book date when running the full fetcher). I also looked at limiting in the browser, but our date-picker is customized text field, rather than a :date field that has `min` built in, so I didn't go that route. Happy to try something with the datepicker, or something else if it you think it would be worth it.

app/controllers/admin/accounting/transactions_controller.rb
```diff
@@ -29,6 +29,20 @@ def create
...
     authorize(@loan, :update?)
     @transaction = ::Accounting::Transaction.new(transaction_params)
+
+    # the txn model must be able to create managed txns before
+    # the closed books date that are imported from qb (e.g. from full fetcher)
+    # but user should not be able to create
+    # a txn before the closed book date via the Madeline interface.
+    # bc there is no way to distinguish Madeline-created from qb-imported
+    # at model level, we do this check here.
+    unless respects_closed_books_date(@transaction)
```
DevneyHamilton: I didn't love adding this logic to a controller action. I tried to figure out a way at the model validation level to distinguish between an imported qb txn and a txn that had been created via the Madeline txn form. I didn't find a way that always works (e.g. we have to be able to create 'Managed' txns before the closed book date when running the full fetcher). I also looked at limiting in the browser, but our date-picker is customized text field, rather than a :date field that has min built in, so I didn't go that route. Happy to try something with the datepicker, or something else if it you think it would be worth it.

DevneyHamilton:
For QA: If this works, a user should 1) see an error if they try to create a txn with a date before the closed books date (which is set in Accounting Settings, introduced in issue 10438) 2) see warnings on loan's transaction page if Madeline 'would have' created an interest transaction before the closed books date. 

For #2 in sandbox mode, would need to create a disbursement and repayment with dates before the closed books date on a new loan IN Quickbooks (not in Madeline). Then visit the txns page for the loan in Madeline, and there should be yellow warning that Madeline wanted to create an int txn.

app/controllers/admin/accounting/transactions_controller.rb
meltheadorable: I think we can detect this at the model level - the full fetcher creates txns through `create_or_update_from_qb_object` rather than via `new` and validations at the model level should run here while `qb_object` is still `nil` for freshly created madeline txns (the qb object won't be created until it passes through the reconciler and txn builder)

I don't necessarily love "test if it has a qb object" as a condition for doing this in the model layer, but I do think I'd prefer that code smell & still doing it via a validation than putting this logic in the controller, at least until we are able to get through a workflow refactor of some kind that may make all these cases easier to distinguish with explicit state tracking.

meltheadorable (approved):
It's okayish to do this in the controller, but I pointed out a way we might be able to move this back into a validation-level concern, but I do have a fever and could be missing something obvious

app/controllers/admin/accounting/transactions_controller.rb
DevneyHamilton: yeah, unfortunately by the time we get to :create, a user-created txn has qb data. I'm trying an ephemeral 'user-created' attr at Tom's suggestion, which is still not super clean but better than what I have I think.

DevneyHamilton: (in general I think it's good that it's hard to tell the diff between a user-created and qb-created txn).

emitche:
When trying to create a transaction before the closed book date, I get this error:

image

The disbursement loads just fine when the transaction is after the date.

The error displayed may not be very helpful to the user. But the code does work as expected: an error is shown when trying to create a transaction before the closed books date.

emitche:
Failing spec appears unrelated to code change.

PR #735, opened 2020-03-09
10495: parse fallback closed book date in interest calculator

DevneyHamilton:


DevneyHamilton:
I will ask Melody if we should handle a fallback/default differently, but this works for now. 

Loan update task WITHOUT this change:
Screen Shot 2020-03-09 at 9 26 07 AM

Loan update task WITH this change:
Screen Shot 2020-03-09 at 9 26 33 AM

emitche (approved):

PR #736, opened 2020-03-10
10304 & 10416: task button and loan update performance

meltheadorable:


app/models/accounting/quickbooks/updater.rb
```diff
@@ -52,15 +60,6 @@ def qb_sync_for_loan_update
             end
           end
         end
-
-        # We record last_updated_at as the time the update *finishes* because last_updated_at
```
DevneyHamilton, line 56: this is wonderful - thank you

app/models/accounting/transaction.rb
```diff
@@ -123,6 +123,7 @@ def self.create_or_update_from_qb_object!(qb_object_type:, qb_object:)
       end
       loan_classes = loan_classes.map { |lc| lc&.match(QB_LOAN_CLASS_REGEX)&.captures&.first }
       associated_loans = Loan.select(:id).where(id: loan_classes)
+      associated_loans.map(&:touch)
```
DevneyHamilton, line 126: cool

app/views/admin/loans/index.html.slim
```diff
@@ -2,6 +2,8 @@
 
 - content_for :with_title
   = link_to t("loan.shared.new"), new_admin_loan_path, class: 'btn btn-default new-button'
+  - if policy(::Accounting::Transaction).update?
```
DevneyHamilton, line 5: Since the updater is currently only updating active loans, should we also only show this link if the loan is active?

app/jobs/update_all_loans_job.rb
```diff
@@ -3,9 +3,10 @@ def perform(_job_params)
     task = task_for_job(self)
     errors_by_loan = []
     divisions = Division.qb_accessible_divisions
-    loans = divisions.map { |i| i.loans.active }.flatten.compact
     updater = Accounting::Quickbooks::Updater.new
     updater.qb_sync_for_loan_update
+
+    loans = divisions.map { |i| i.loans.changed_since(updater.qb_connection.last_updated_at).active }.flatten.compact
```
DevneyHamilton: One thing I missed when I updated this task to do active loans, rather than all loans, is renaming the task to "Update Active Loans." Should we do that so we don't get confused later?

app/models/accounting/quickbooks/full_fetcher.rb
```diff
@@ -16,7 +16,7 @@ def fetch_all
         fetch_qb_data
         restore_accounts!(accounts)
       rescue StandardError => error
-        @qb_connection.destroy
+        @qb_connection&.destroy
```
DevneyHamilton, line 19: super curious what this change does, if you have a moment to explain.

DevneyHamilton (commented):
This is awesome, so excited! I put a couple qs about 'active' vs 'all' loans. These aren't blocking. We do need to decide if we want people to be able to update non-active loans. If not, we need to make that clear in the UI. If so, we need to think about how to support that.

app/views/admin/loans/index.html.slim
meltheadorable: This is on the loan index, so it isn't per-loan.

app/models/accounting/quickbooks/full_fetcher.rb
meltheadorable: It prevents the crash handler from crashing if there is no quickbooks connection

app/views/admin/loans/index.html.slim
DevneyHamilton: ah ok, thanks.

DevneyHamilton:
@emitche  for QA using the Madeline Staging Company NEW, can you please QA this WITH 10526 merged, rather than this branch on it's own? This is because 10416 would theoretically add or change transactions on non-active loans to the QB company data, which we don't want. 10526 is designed to prevent adding or changing qb txns on non-active loans.  If we need to, we can revert the Madeline Staging Company NEW to a backup version of itself, but there's no handy way to test 10416 and then 'reset' the qb company to re-test it with 10526. Happy to talk more about this if this is not making sense.

emitche:
@DevneyHamilton That QA plan makes sense. Thank you for the update. If I have questions, I will let you know.

PR #738, opened 2020-03-11
10526: do not run interest calculation on active loans

DevneyHamilton:
This is a one-line change in the interest calculator to not run `recalculate` unless the loan is active. 

I reviewed the updater to confirm that the updater will take care of saving new QB information and updating balances for txns on non-active loans. 

I briefly tried adding a loud error to the reconciler if it is asked to update a txn whose loan is not active, but the spec situation wasn't easy to work with and it seemed not important enough to force this morning.

Things to note for the future: the `recalculate` method is where a NEW transaction created in Madeline get their line items set up. This is not ideal, and means if a user tries to create a txn on a non-active loan, it will not go to qb and fail silently. we could address this with by prohibiting transaction creation in the UI on non-active loans, refactoring madeline-prompted txn creation to be independent of the interest calculator, or raising an error in the interest calculator that will reach the user. @emitche is there a way to record this in the backlog? It's not on the critical path to accounting but don't want to lose track of it.

### DIFF
--- a/app/models/accounting/interest_calculator.rb
+++ b/app/models/accounting/interest_calculator.rb
@@ -59,6 +59,7 @@ module Accounting
     end
 
     def recalculate
+      return unless @loan.active?
       prev_tx = nil
 
       txns_by_date = transactions.group_by(&:txn_date)
--- a/app/models/loan.rb
+++ b/app/models/loan.rb
@@ -204,7 +204,7 @@ class Loan < Project
   end
 
   def active?
-    status_value == 'active'
+    status_value == STATUS_ACTIVE_VALUE
   end
 
   def healthy?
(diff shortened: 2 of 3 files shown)


DevneyHamilton:
QA: the specific logic here is covered in a unit test. I manually tested via `be rake madeline:enqueue_update_loans_task` on my local to make sure we didn't blow up with a bunch of errors. Theoretically, if we set a recent Closed Books Date before running `madeline:enqueue_update_loans_task` since the 10416 changes, then we'd want to see closed book date warnings on active loans, but not on non-active loans. But this is not reliable because if we run `madeline:enqueue_update_loans_task` on 10416 but without 10526, those non-active loans will have had interest calculated on them. 

I'm not sure of a reliable way to QA this though. @meltheadorable should we log when we are recalculating on a loan, with the loan's status?

PR #739, opened 2020-03-16
10537: hide update txns button on loan index  if current division is …

DevneyHamilton:
…not connected to qb

In order to release 3.x to production before we 'turn on' accounting, we should hide this 'update transactions' button on the loans page so that TWW admins are less likely to be confused. 

I manually tested this on my local. The button shows when I am connected qb. It is gone when I disconnect. This is just the minimal change needed for right now. If we want to do more, we could visibly disable the button with a message about how qb needs to be connected to update txns.


emitche (approved):
This code and approach looks good.

meltheadorable (approved):

PR #740, opened 2020-03-16
10536: add more details and warnings to closed book date help text.

DevneyHamilton:
In demo w/ Brendan he requested more details in the closed books date help text, including what the current one is, and when it was set. We'd need to add database fields to handle when the cbd was set, so I didn't include that for now. This is what the new help text is when the cbd is not set: 
Screen Shot 2020-03-16 at 4 36 59 PM

and when it is:
Screen Shot 2020-03-16 at 4 36 00 PM

### DIFF
--- a/app/views/admin/settings/index.html.slim
+++ b/app/views/admin/settings/index.html.slim
@@ -75,7 +75,7 @@ section.block
           = documentation_popover(@documentations_by_html,
             html_identifier: "#{form_identifier}_closed_books_date", options: popover_options)
         - help_note_type = @division.closed_books_date.present? ? "note_set" : "note_not_set"
-        p.help-block.note = t("quickbooks.closed_books_date.#{help_note_type}", current_date: @division.closed_books_date )
+        p.help-block.note = t("quickbooks.closed_books_date.#{help_note_type}", current_date: @division.closed_books_date)
 
         = f.input :closed_books_date, as: :date_picker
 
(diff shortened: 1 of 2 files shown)


app/views/admin/settings/index.html.slim
```diff
@@ -74,7 +74,8 @@ section.block
           h2 = t("division.closed_books_date")
           = documentation_popover(@documentations_by_html,
             html_identifier: "#{form_identifier}_closed_books_date", options: popover_options)
-        p.help-block.note = t("quickbooks.closed_books_date.note")
+        - help_note_type = @division.closed_books_date.present? ? "note_set" : "note_not_set"
```
DevneyHamilton, line 77: is there a better way to choose between two different translation entries?

emitche:
The warning seems pretty important and should be called out. 

If we did something like this, would that indicate an appropriate amount of warning?

image

Quick mockup of this is at: https://github.com/sassafrastech/madeline/tree/10536_ux

emitche:
We have original start date and original duration in the project step modal (represented as another field, as shown below). But I think the approach you took here is a better approach for clarity.

image

app/views/admin/settings/index.html.slim
emitche: This approach seems fine.

app/views/admin/settings/index.html.slim
```diff
@@ -74,7 +74,8 @@ section.block
...
+        p.help-block.note = t("quickbooks.closed_books_date.#{help_note_type}", current_date: @division.closed_books_date )
```
emitche: Extra space here at the end.

emitche:
For when there is no closed books date set, I might actually call that out for clarity. Something like:

"Madeline will not create transactions or modify Quickbooks transactions before the closed books date. There is currently no closed books date set."

emitche:
If desired, these changes look good and I can make the UX changes on a different branch/pull request. Thank you!

emitche:
The UX tweaks, which I will take care of, can go in a separate pull request.

PR #741, opened 2020-03-17
10538 fix loan health job explosion on qb import

DevneyHamilton:


DevneyHamilton:
For QA: The point of this issue is that before it, Madeline enqueues a `RecalculateLoanHealthJob` for every single transaction it imports from QB. After this issue, it should not enqueue a `RecalculateLoanHealthJob` when a transaction is imported from QB. 

To manually test this, I watched  the /admin/jobs/busy view while running the FullFetcher job (that runs when I re-connect to QB). It is a slow way, since the full fetch from qb takes a few minutes. There may be better ways to test this.  

WITH this change working, about 5 minutes into the FullFetcher Job, that is still the only job I see at /admin/jobs/busy:
Screen Shot 2020-03-17 at 3 11 11 PM

WITHOUT this change working, about 5 minutes into the FullFetcher Job, I see lots of "RecalculateLoanHealthJob" entries at /admin/jobs/busy (and my computer sounds like it's about to explode :)).
Screen Shot 2020-03-17 at 3 03 21 PM

. . .  I also see thousands of jobs enqueued at  /admin/jobs/queues. 
Screen Shot 2020-03-17 at 3 04 26 PM
Screen Shot 2020-03-17 at 3 04 10 PM

meltheadorable (approved):

emitche:
QA: I'm not going to try this without the fix. But I am confirming that only the job we want to see shows up with this fix.

emitche:
One of the tests recently failed. I'm rechecking to see if the tests fail again.

PR #742, opened 2020-03-17
10422: Make autocomplete reusable throughout the application

emitche:
- Add a class that can be used to turn a select into an autocomplete select.


emitche:
@meltheadorable Code review here is not needed. In terms of approach, we already were using the `select2-rails` gem inside the application. But we were only using it only one form. This code change allows any select to easily be changed to an autocomplete select by simply adding a class.

PR #743, opened 2020-03-17
10501 accounting settings

emitche:
- Make account selection use autocomplete select for easier searching through accounts
- Adjust styling of autocomplete select


DevneyHamilton (approved):
So refreshing! Thank you so much Elizabeth!

PR #747, opened 2020-03-20
10552 closed books date warning

emitche:
- Add warning to call attention to impact of closed book date change


emitche:
Testing this with fake data and a QuickBooks sandbox.

If a closed books date is set:
image

emitche:
If a closed book date is not set, the administrator sees:

image

PR #749, opened 2020-03-23
10555: show warning if accounting settings have qb_read_only set to true

DevneyHamilton:


DevneyHamilton:
Closing this because I think the extra info text will be misleading (admins should NOT change the accounting settings read-only setting just because they want to update a loan; it's a major decision that the admin needs to be fully informed about).

PR #755, opened 2020-04-01
10576: Raise error if QB tries to write in read-only mode

meltheadorable:


DevneyHamilton (approved):
I manually tested this on my local. It's good to go. 
A note: because the update_all_loans_job catches StandardError PER loan (line 15 in update_all_loans_job.rb), the overall update will still happen. Worst case is we get a bunch of errors on the update task, but it will error before the reconciler sends data to QB. I manually tested this by undoing the read-only check in the interest calculator, and this is what happened.

PR #756, opened 2020-04-01
10529 show admin users

emitche:


emitche:
With fake data:

The Access Role filter is a column-level filter instead of a top-level filter before the grid.

Ideally, both filters would be available, but wice grid wants filters to refer to items inside tables. Access role is a method and though there is an access role table, it's not currently directly tied to the people table. @meltheadorable If there is a better way to do this, please let me know.

image

emitche:
For review, please let me know if you think this is enough of an admin filter (that the filter is usable and understandable). Otherwise, we could just make a page that lists all admins and have that page listed in the Manage menu. That way the admins can see which admins belong to which division in a page only accessible to admins.

DevneyHamilton (approved):
Thanks! I think this'll be helpful to them.

DevneyHamilton:
> For review, please let me know if you think this is enough of an admin filter (that the filter is usable and understandable). Otherwise, we could just make a page that lists all admins and have that page listed in the Manage menu. That way the admins can see which admins belong to which division in a page only accessible to admins.

Yeah, I think this is good to start and might meet their needs without further work or adding a page. Am I understanding right that  sorting and/or filtering by admin status harder than other fields?

DevneyHamilton:
Update: what I said above was just based on having the column displaying member or admin or none access role. I think displaying that column will be helpful.  I didn't see the eye icon on the far right until now. When I open it up, I get 
Screen Shot 2020-04-14 at 10 21 59 AM. I don't quite understand how to use it.

emitche:
@DevneyHamilton Thank you for looking this over. 

> Am I understanding right that sorting and/or filtering by admin status harder than other fields?

Yes, admin role is not in the database. It's a method on the model. Wice grid expects to use fields that are saved in the database. Perhaps with `has_one` we could adjust access role to work better with Wice Grid. I'm not sure.


> Update: what I said above was just based on having the column displaying member or admin or none access role. I think displaying that column will be helpful. I didn't see the eye icon on the far right until now. When I open it up, I get
> Screen Shot 2020-04-14 at 10 21 59 AM. I don't quite understand how to use it.

Thank you for catching this. In the code, I have access role mapped to whether or not the user has system access. That's why it shows yes/no. I can remove the option of showing the yes/no filter.

The way I was using the filter was just to click on the access role heading, which then sorts by name. Then you see the users listed by "Admin" or "Member".

However, in the system there are 150+ users. I think it might be better to create an admin page that lists all the admins, given that this filter is not working out well.

DevneyHamilton:
Ah, @emitche I'm really sorry, I was rushed and I missed the comment where you had explained the difficulty with this. Sorry for taking more of your time & energy on re-explaining. Thanks for finding a couple possible ways to make it work.  If you can remove the option of showing the yes/no filter and still show the column with "admin" or "member", that seems like it might meet TWW's need.  But I'd defer to you and Melody if either/both of you have a good sense of how to proceed, like with a page instead.

emitche:
@DevneyHamilton Thanks so much! I'll look into hiding the yes/no filter.

emitche:
I just tested this with real data. Though I can hide the true/false, the filtering of users is not actually working correctly. So I think the best approach is a list of admin users.

emitche:
@meltheadorable I've set up pairing for this tomorrow. But if you have any feedback before then, please feel free to leave it. I'm putting this on hold until we pair.

emitche:
Access Role is visible inside the admin panel. It is not filterable. But the admins can export the list and sort by role, if needed.

image

emitche:
I spoke with @meltheadorable about this and based on the code, this approach makes sense. The code change is small and @DevneyHamilton already approved it.

@DevneyHamilton can you give this a look again and make sure it make sense to you from a user perspective? You caught the filter issue, which was a big help. Thank you. There's no filter now, but they can use CSV to export the content and sort.

DevneyHamilton:
QA notes: this worked just fine, looks goods, and sounds like it will meet their needs. Thanks @emitche!

emitche:
@DevneyHamilton Thank you!

PR #759, opened 2020-05-14
10684: running this task on staging and prod, learned we need to load…

DevneyHamilton:
… paths in all envs.

 I ran the resave_all_records task on staging and prod. To do so,  I commented out the "if env is development' line manually on the server. Since this is a lib task, not part of the actual app, that seemed best in terms of time/complexity. This change is just so we have an accurate record in the codebase of the task I actually ran.


DevneyHamilton:
I don't think this needs QA etc

PR #760, opened 2020-05-19
10766 add error warning addr fields to data export

DevneyHamilton:


DevneyHamilton:
Rearranged columns to help user see that data export attachment is available even if task has 'failed' status:
Screen Shot 2020-05-27 at 9 20 05 AM

and in tasks view:
Screen Shot 2020-05-27 at 9 20 24 AM

meltheadorable (approved):

meltheadorable:
I think I'd prefer the attachments to move back to the end of the column -- it's _really_ weird to just be hovering in the middle like that, but I like duplicating the task status information on the data export page.

DevneyHamilton:
Updated screenshot:
Screen Shot 2020-05-29 at 8 41 57 AM

I left the task link to the right of the attachment bc its the least important. When I was testing this, I thought there was a bug in terms of status handling or in the task because I got the failure message, and gave up before reading across the row to see the attachment link was available. It wasn't til I looked into the code that I remembered we'd made the UX compromise of marking 'failed' even if attachment is available. It was really easy to miss that the attachment was, in fact, available for download. I hope the activity message with "Completed" in it will help.

emitche:
QA review:
All data exports as expected.

UX:
The "Failed" message is confusing. I would check in with TWW about this during the demo to discuss what they would prefer to see. The activity message helps, but this could be clearer.

PR #767, opened 2020-06-23
10806: question numbering

meltheadorable:


app/models/question.rb
```diff
@@ -151,18 +152,29 @@ def build_complete_requirements
   end
 
   def full_number
```
DevneyHamilton, line 156: Is this still named `full_number` to avoid renaming in code using this, or some other reason? It seems like what this function returns is no longer actually 'full_number.' The full number would look like III.2.A, for example. If we want this function to do what it's doing here, could we rename it to something like outline_head (I looked this up in wikipedia  . .. )? I can imagine a request from tww to do a 'full_number' on leaf nodes that would actually have III.2.A.ii etc. I think  that updating the name would help us adjust as we get feedback and with onboarding others.

lib/romanize.rb
```diff
@@ -0,0 +1,30 @@
+module Romanize
+  NUMERAL_MAP = {
```
DevneyHamilton, line 2: thanks for doing this and putting it in lib!

app/assets/stylesheets/admin/loans/_questionnaires.scss
```diff
@@ -277,3 +277,7 @@ section.questionnaires {
   font-style: italic;
   font-weight: bold;
 }
+
+ul.toc {
```
DevneyHamilton, line 282: thanks for adding class. took me a few to realize toc was table of contents. does that merit adding a quick comment?

DevneyHamilton (changes requested):
Thank you for fitting this in this week! The Romanize lib is nice :) Looks good on a branch I tried. I requested a naming update and asked a question about whether the toc abbreviation should be explained.

app/models/question.rb
meltheadorable: I tried renaming this and it caused issues in unexpected parts of the codebase for reasons it didn't seem time-efficient to diagnose and try to resolve. I don't want to have to push the release back to accommodate this.

app/assets/stylesheets/admin/loans/_questionnaires.scss
meltheadorable: `toc` is used for table of contents most other places in the code that refer to the table of contents, but I can add a comment if you think it will help

PR #768, opened 2020-06-24
10858 error monitoring

meltheadorable:


app/controllers/application_controller.rb
```diff
@@ -71,4 +73,9 @@ def locale_from_header
       lang_header.scan(/^[a-z]{2}/).first.to_sym
     end
   end
+
+  def set_raven_context
+    Raven.user_context(id: current_user.id, email: current_user.email) # or anything else in session
```
DevneyHamilton: adding `if current_user`to the end of line 78 fixed specs. Don't love changing application code to fix specs, so feel free to do something else.

DevneyHamilton (changes requested):
Thanks for finding something that is so light in terms of additional code here! I'm looking forward to trying it. I'm requesting changes to fix specs, in case you want to do something different than what I tried real fast (adding `if current_user` in the application controller).

PR #772, opened 2020-06-26
10876: add version number to page footer

meltheadorable:
So we can check the version while logged in


DevneyHamilton (approved):
Thank you! this will be nice :)

PR #773, opened 2020-06-26
10877: Adds division public status on index

meltheadorable:


DevneyHamilton (approved):
Thank you! Did a quick QA pass to makes sure I saw yeses and nos, looks good.

PR #774, opened 2020-06-26
10506: fix some n+1 queries

meltheadorable:
This might further improve loan memo performance


DevneyHamilton (approved):
Thank you for finding & fixing this @meltheadorable !

DevneyHamilton:
When I load the questions page on my development environment, I get a notice that the application_policy.rb is triggering a query that needs revision. If this only shows up in development mode, then it might be ok to deploy as is. What do you think? Screen Shot 2020-07-10 at 10 22 31 AM

PR #777, opened 2020-08-03
10978: display all loans for all divisions if division not set, inste…

DevneyHamilton:
…ad of using country in url to find division by shortname

### DIFF
--- a/app/controllers/public/loans_controller.rb
+++ b/app/controllers/public/loans_controller.rb
@@ -3,7 +3,7 @@ class Public::LoansController < Public::PublicController
   after_action :verify_authorized
 
   def index
-    params[:division] ||= layout_site
+    params[:division] ||= Division.root.short_name
     @params = { status: params[:status], pg: params[:pg], division: params[:division] }
     @loans = policy_scope(Loan.filter_by_params(params).
           includes(:organization, division: :parent).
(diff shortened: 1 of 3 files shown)


DevneyHamilton:
I went with just displaying all loans because
the previous logic was relying on the url country matching the short_name of the division. 
I considered creating a mapping, until I found that the nicaragua shortname is "la-base-nicaragua-efbd2fd6-611c-4bd5-af17-506ee9ab68a2" which doesn't seem stable. So I think that should wait til the wordpress redo. 

I don't know what to do UX wise for the "now displaying all loans" text that would be helpful and not make the page look worse. I created an issue for that you or Jill could pick up when you have time https://redmine.sassafras.coop/issues/10984.

DevneyHamilton:
With no division: 
Screen Shot 2020-08-03 at 4 44 29 PM

WIth division selected:
Screen Shot 2020-08-03 at 4 55 16 PM

DevneyHamilton:
For QA:
- ensure a lot of loans display when no division set (/us/loans, /argentina/loans, /nicaragua/loans)
- ensure only appropriate loans when a division IS set

DevneyHamilton:
the failing ci seems to be flapping. I may just be really tired, but for the life of me I can't find the "rerun build" button.

PR #783, opened 2020-09-02
11047: show qb division in txn modal, always read-only

DevneyHamilton:
Display the qb dept in the txn modal, always read-only. 
If qb dept not present, show warning. This case will not happen after issue 11046 in this sprint is merged. 

dept set in edit view:
Screen Shot 2020-09-02 at 10 27 45 AM

dept set in read-only view:
Screen Shot 2020-09-02 at 10 56 14 AM

dept not set:
Screen Shot 2020-09-02 at 10 23 16 AM


app/views/admin/accounting/transactions/_form.slim
```diff
@@ -30,6 +30,13 @@
       - if @transaction.customer
         = @transaction.customer.name
 
+  = f.input :qb_department
+    .form-element
+      = (@transaction.qb_department ? @transaction.qb_department.name : t('quickbooks.department_not_set'))
```
meltheadorable, line 35: I'm not seeing a "department not set" message in the translation files, and the message in the screenshots no longer aligns with the behavior we want (prevent creating txns without a department set instead of allowing them with a blank department)

meltheadorable: I see the message now, it just wasn't flagged as a change, my bad - and I see now that 11046 is probably the one that's going to change the message/behavior related to creating txns. Ignore me.

meltheadorable (approved):
This looks good overall -- I still want to see us figure something we can do with departments that have identical names to each other (there must be a way to differentiate them?) and ideally whatever we decide works both here and in the other places we show those names, like dropdowns. No changes needed here though, just flagging that concern again.

app/views/admin/accounting/transactions/_form.slim
DevneyHamilton: yes - that is correct, 11046 has the change to require dept to make a txn. Thanks for keeping eye out!

PR #785, opened 2020-09-02
11046 require qb div to make txn

DevneyHamilton:


DevneyHamilton:
Warning and button hidden if Madeline division on the loan does not have qb division set. 
I did not link to the division in the warning bc only admins can edit divisions, but more users can see the txns page. 
Screen Shot 2020-09-04 at 11 04 32 AM

meltheadorable (approved):
discussed in slack - looks good, with possible UX adjustments in future

PR #798, opened 2020-10-13
Fix for flapping media spec

DevneyHamilton:
Looks like a spec about sorting media has been passing 'by accident' for a long time because the db happened to return items in the right order.


DevneyHamilton:
Merging w.out review so other branches can pass CI

PR #799, opened 2020-10-14
11153: Added email notification when jobs not running

meltheadorable:


DevneyHamilton (approved):
Thanks for adding this!

PR #806, opened 2020-10-21
11169: add questionnaire numeric answers to standard data export

DevneyHamilton:
This issue adds numeric business planning criteria answers to the standard loan data export. It adds number, rating ("range"), percentage, and currency answer values. The headers for these columns are the question's translated text. The second row of the csv are question ids. 

It currently takes ~30 mins to run this on my local machine, with each of 1715 loans taking about a second to run. 

Tech debt: this task relies on the logic creating a Response model object from the json blob. 

We may want to consider moving the question data to a separate data export for a few reasons:
1) for managing the codebase, if we move it out, then standard data export will not be touched by refactoring questions and responses
2) the question part of this data export is not tested (due to tech debt w/ questions/answers)
3) with question data, the export takes much longer, and TWW could have a standard_data_export that takes a few minutes on its own. 

We'll want to get feedback from TWW about how using this goes, and expect followup work. E.g. they may want us to show progress (num loans added to export) which will require some work to expose that info to the task running the export. 

Here is an export: 
[Standard_Loan_Data_Export_on_Oct_29__2020__2_55_PM20201029-42667-1r0txtn 3.pdf](https://github.com/sassafrastech/madeline/files/5465188/Standard_Loan_Data_Export_on_Oct_29__2020__2_55_PM20201029-42667-1r0txtn.3.pdf)

Note: some loans, esp argentina ones, have alphanumeric answers - this is a problem with the underlying response data, not with the export (the responses also have these values in the browser view of the response). E.g. loan 720 (not active, argentina) has a value "HwGYE" for "Workers in the business at present." We should point this out, and perhaps they can use this report to remedy that data issue.

### DIFF
--- a/spec/services/data_export_service_spec.rb
+++ b/spec/services/data_export_service_spec.rb
@@ -47,7 +47,8 @@ describe DataExportService do
         data = data_export.reload.data
         headers = data[0]
         status_index = headers.index("Estado")
-        expect(data[1][status_index]).to eq "Activo"
+        # use second row because second row of standard loan data export has question ids
+        expect(data[2][status_index]).to eq "Activo"
       end
     end
   end
--- a/spec/services/data_export_service_spec.rb
+++ b/spec/services/data_export_service_spec.rb
@@ -47,8 +47,7 @@ describe DataExportService do
         data = data_export.reload.data
         headers = data[0]
         status_index = headers.index("Estado")
-        # use second row because second row of standard loan data export has question ids
-        expect(data[2][status_index]).to eq "Activo"
+        expect(data[1][status_index]).to eq "Activo"
       end
     end
   end
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -35,7 +35,8 @@ class DataExport < ApplicationRecord
   before_save :set_name
 
   DATA_EXPORT_TYPES = {
-    "standard_loan_data_export" => "StandardLoanDataExport"
+    "standard_loan_data_export" => "StandardLoanDataExport",
+    "enhanced_loan_data_export" => "EnhancedLoanDataExport"
   }
 
   # Process data should be defined on subclasses,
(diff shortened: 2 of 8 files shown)


spec/services/data_export_service_spec.rb
```diff
@@ -47,7 +47,8 @@
         data = data_export.reload.data
         headers = data[0]
         status_index = headers.index("Estado")
-        expect(data[1][status_index]).to eq "Activo"
+        # use second row because second row of standard loan data export has question ids
+        expect(data[2][status_index]).to eq "Activo"
       end
```
DevneyHamilton: this particularity coming thru in the service spec is making the 'standard loan data export' not feel very 'standard' to me. I think it may be easier to manage changes or bugs in the future if we break the question data out into a separate data export, which can be another issue next sprint. I put more about that in the PR comment. What do you think @meltheadorable ?

meltheadorable (commented):
This looks great - this will obviously need to be QA'ed on staging, I don't have the environment to do a local test, but the implementation work looks clean.

I'd like to see us restore the original standard export in addition to this, but otherwise I'm ready to approve this.

DevneyHamilton:
> This looks great - this will obviously need to be QA'ed on staging, I don't have the environment to do a local test, but the implementation work looks clean.
> 
> I'd like to see us restore the original standard export in addition to this, but otherwise I'm ready to approve this.

thank you @meltheadorable !

DevneyHamilton:
Current view of screen after clicking "New Data Export", where admin user chooses the type of export they want. 
Screen Shot 2020-11-03 at 11 52 18 AM

meltheadorable (approved):
Work splitting these out into separate reports looks great! I think we should plan to create a more robust "choose type" page that can provide some additional information and a nicer UI for selecting report types, but that'd need design and should definitely be a separate ticket. Excellent work! 💯

meltheadorable:
Marked "needs qa" because this is one of those tickets we will have to QA on staging, my local env can't support it right now.

PR #807, opened 2020-11-04
11255: ensure doc number always updated when writing to qb

DevneyHamilton:
This is a small fix, that I manually tested in my dev environment. In the txn builder, the doc number (also known as journal number or check number) was only being set on new txns, not if the txn had a qb id. Now it's always set.

### DIFF
--- a/app/models/accounting/qb/transaction_builder.rb
+++ b/app/models/accounting/qb/transaction_builder.rb
@@ -34,10 +34,8 @@ module Accounting
         if transaction.qb_id
           p.id = transaction.qb_id
           p.sync_token = transaction.sync_token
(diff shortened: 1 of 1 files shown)


meltheadorable:
I feel like I've seen this change before - it looks fine but I'm just wondering why this code is giving me deja vu

meltheadorable (approved):

DevneyHamilton:
> I feel like I've seen this change before - it looks fine but I'm just wondering why this code is giving me deja vu

oh, I wish I knew - I'll let you know if I think of anything. 

it is possible you reviewed the change and it got lost in a merge

PR #808, opened 2020-11-04
11208: documentation index

meltheadorable:


app/views/admin/documentations/index.html.slim
```diff
@@ -0,0 +1,16 @@
...
+- @documentations.each do |calling_controller, documentation_list|
+  h2= t("controllers.admin.#{calling_controller}")
+  - documentation_list.each do |documentation|
+    table.table
+      thead
+        tr
+          td= documentation_title(documentation)
+          td= documentation_location(documentation)
+          td= documentation_division(documentation)
+          td.text-right= documentation_edit(documentation)
```
DevneyHamilton: Do we need to check here if the user has permission to edit before showing the edit link? in QA, when I clicked to edit, I was allowed to get to the edit page and use the form, but when i clicked submit button ("Update Documentation") I was redirected to https://ms.tww.org/users/sign_in . ..

DevneyHamilton (changes requested):
This is looking so nice! Found a QA issue where I can start, but not finish editing a documentation item. As a root division admin, when I hit 'Update Documentation' I got redirected to ttps://ms.tww.org/users/sign_in . . . @meltheadorable can you pls confirm if you can go thru the whole flow of updating a documentation from this index view?

app/views/admin/documentations/index.html.slim
DevneyHamilton: update - issue is with legacy redirect based on url saved to db. For now, hide edit button; change redirect behavior in issue next sprint

PR #809, opened 2020-11-13
11285 enhanced data export crash

DevneyHamilton:
This issue should be released in 3.15.1, to staging and production today (Fri Nov 13, 2020)

Updated to remove lines that were triggering unnecessary recursive db activity (see details in the comment in the file). This task runs in <2 mins on staging :D. See "updated version of enhanced data export" at https://ms-staging.theworkingworld.org/admin/data-exports for example output

### DIFF
--- a/app/models/enhanced_loan_data_export.rb
+++ b/app/models/enhanced_loan_data_export.rb
@@ -23,7 +23,6 @@
 #
 
 class EnhancedLoanDataExport < DataExport
-
   Q_DATA_TYPES = ['number', 'percentage', 'rating', 'currency']
   BASE_HEADERS = [
     'loan_id',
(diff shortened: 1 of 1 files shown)


app/models/enhanced_loan_data_export.rb
```diff
@@ -112,12 +113,13 @@ def loan_data_as_hash(loan)
   end
 
   def response_hash(loan)
-    return {} if loan.criteria.blank?
     result = {}
-    response_set = loan.criteria
```
DevneyHamilton, line 117: lines 115 and 117 were the culprit - they trigger recursive building of the question tree from the database, which was happening twice per loan. we don't need question data here (on purpose) - we rely on the question ids in the raw response json. the new logic in lines 117, 118, and 119 let us access that response json w/out triggering the question-related db queries.

meltheadorable (approved):

PR #810, opened 2020-11-16
11048: Move update button

meltheadorable:


app/controllers/admin/accounting/quickbooks_controller.rb
```diff
@@ -53,6 +53,19 @@ def connected
     render json: @quickbooks_connected
   end
 
+  def update_changed
+    authorize :'accounting/quickbooks', :update?
+    @task = Task.create(
+      job_class: FetchQuickbooksChangesJob,
```
DevneyHamilton, line 59: thanks for making the name of the job more accurate!

config/routes.rb
```diff
@@ -84,14 +84,11 @@
...
-        end
-      end
-
-      resources :transactions do
-        collection do
           patch :update_changed
         end
       end
+
+      resources :transactions
```
DevneyHamilton, line 91: oh it's nice this ends up cleaner too :)

DevneyHamilton (approved):
This seems like a really nice improvement for them! I tested on local and it worked for me. Thank you!

app/models/accounting/qb/updater.rb
```diff
@@ -113,6 +113,8 @@ def types
       def ar_model_for(qb_object_type)
         return Account if Account::QB_OBJECT_TYPE == qb_object_type
         return Customer if Customer::QB_OBJECT_TYPE == qb_object_type
+        return Vendor if Vendor::QB_OBJECT_TYPE == qb_object_type
+        return Department if Department::QB_OBJECT_TYPE == qb_object_type
```
DevneyHamilton, line 117: @meltheadorable in qa, I found this bug from my earlier work on importing vendors and departments. We need these lines for the objects to actually be created on import. hoping it's ok to slip the fix in here, but I can also revert the branch and open separate pr if you'd prefer that.

meltheadorable: This is fine

PR #811, opened 2020-11-17
11251 all disb as purchases

DevneyHamilton:
This PR refactors the txn creation flow to make all new Madeline disbursements QB Purchases. Previously, a madeline disbursement could be a JE or a Check Purchase. This change is motivated by all TWW needs all disbursements, not just check disbursements, to have the vendor field visible in QB. The vendor field doesn't show up for JEs in QB.


app/controllers/concerns/transaction_controllable.rb
```diff
@@ -24,7 +24,7 @@ def prep_transaction_form
     @loan_transaction_types = Accounting::Transaction.loan_transaction_type_options.select do |option|
       Accounting::Transaction::AVAILABLE_LOAN_TRANSACTION_TYPES.include?(option[1].to_sym)
     end
-    @qb_subtypes = [["Check", "Check"], ["None", nil]]
+    @qb_subtypes = Accounting::Transaction::QB_PAYMENT_TYPES.map { |t| [I18n.t("transactions.qb_object_subtype.#{t.downcase}"), t] }
```
DevneyHamilton, line 28: AVAILABLE_LOAN_TRANSACTION_TYPES are option sets in madeline. for QB_PAYMENT_TYPES 
I used i18n instead, since these types are determined by QB and cannot be user created in any case.

app/models/accounting/qb/transaction_reconciler.rb
```diff
@@ -35,7 +35,7 @@ def builder
       end
 
       def service(m_txn)
-        m_txn.subtype?("Check") ? purchase_service : je_service
+        m_txn.qb_object_type == "Purchase" ? purchase_service : je_service
```
DevneyHamilton, line 38: It is unlikely but possible there will be legacy disbursements that are JEs, so we decide based on qb object type, not txn type

app/models/accounting/qb/transaction_builder.rb
```diff
@@ -18,16 +18,16 @@ def initialize(qb_division = Division.root)
       end
 
       def build_for_qb(transaction)
-        if transaction.subtype?("Check")
-          build_check_for_qb(transaction)
-        else
+        if transaction.qb_object_type == "JournalEntry"
```
DevneyHamilton, line 21: It is unlikely but possible there will be legacy disbursements that are JEs, so we decide based on qb object type, not txn type

meltheadorable (approved):
This all looks good to me

meltheadorable (approved):

PR #812, opened 2020-11-17
11253 balances not calced for some loans

DevneyHamilton:
So we learned when we need to call calculate balances in the updater - if a loan shouldn't have interest calculation run, we still need to calculate balances for it. This has a first pass at those changes. I wanted to get your feedback before going farther. . .

### DIFF
--- a/app/models/accounting/qb/updater.rb
+++ b/app/models/accounting/qb/updater.rb
@@ -81,7 +81,7 @@ module Accounting
       # Creates/deletes LineItems as needed.
       def extract_qb_data(loan)
         if loan.transactions.present?
-          loan.transactions.each do |txn|
+          loan.transactions.standard_order.each do |txn|
             if txn.quickbooks_data.present?
               Accounting::QB::DataExtractor.new(txn).extract!
               txn.save!
--- a/app/models/accounting/qb/updater.rb
+++ b/app/models/accounting/qb/updater.rb
@@ -61,7 +61,13 @@ module Accounting
         end
       end
 
-      # updates single loan
+      # To update a loan, we first must extract any new qb data on its txns.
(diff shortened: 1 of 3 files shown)


app/models/accounting/qb/updater.rb
```diff
@@ -61,32 +61,49 @@ def qb_sync_for_loan_update
...
+      private
+
+      # Extracts data for txns from the JSON in `quickbooks_data`
+      # into each Transaction's attributes and associated LineItems.
+      # Creates/deletes LineItems as needed.
+      def extract_qb_data(loan)
         if loan.transactions.present?
-          loan.transactions.standard_order.each do |txn|
-            extract_qb_data(txn)
+          loan.transactions.each do |txn|
```
DevneyHamilton: we do not need standard order for qb data extraction, since each txn's data extraction should be indep of other txns.

app/models/accounting/qb/updater.rb
```diff
@@ -61,32 +61,49 @@ def qb_sync_for_loan_update
...
+      # Though interest calculation calculates balances on any loans whose interest
+      # is calculated, we call calculate balances again after the interest calculation
+      # because not all loans have their interest calculated in 'recalculate', but all loans need
+      # their balances calculated. Note: attempting to calculate balances
+      # on all loans before interest calculation has caused problems in the past.
       def update_loan(loan)
+        extract_qb_data(loan)
+        InterestCalculator.new(loan).recalculate
+        calculate_balances(loan)
+      end
```
DevneyHamilton, line 75: iterating over txns twice (once in extract qb data, once in calculate balances) is unavoidable, but we only need standard order in calculate balances

PR #816, opened 2020-12-04
11254 txn modal edit new bug

DevneyHamilton:
Diagnosis: pre-existing code called loadContent for a new loan in `initialize.` Since `initialize` is called in the constructor, if the modal had not yet been initialized when the user clicks an existing txn there was a race condition between two calls to loadContent completing - one for 'new' and one for the existing txn.

Fix: move loadContent call out of initialize. Call `showModal` for both new and existing txns, and handle conditional logic for args to loadContent there.

### DIFF
--- a/app/assets/javascripts/admin/transaction_modal_view.coffee
+++ b/app/assets/javascripts/admin/transaction_modal_view.coffee
@@ -11,8 +11,7 @@ class MS.Views.TransactionModalView extends Backbone.View
     @loanId = params.loanId
     @locale = params.locale # need this later for description, keep
 
-  show: (txn_id, action) ->
-    console.log("refactored_show")
+  showModal: (txn_id, action) ->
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/admin/transaction_modal_view.coffee
```diff
@@ -10,13 +10,13 @@ class MS.Views.TransactionModalView extends Backbone.View
   initialize: (params) ->
     @loanId = params.loanId
     @locale = params.locale
-    url = "/admin/accounting/transactions/new"
-    @loadContent(url, @loanId, 'new')
```
DevneyHamilton, line 14: this is called thru constructor even if action is 'show,' kicking off race with the other correct call to loadContent in 'show'

meltheadorable (approved):

PR #820, opened 2020-12-18
11395: Add first/last page to pagination

meltheadorable:


DevneyHamilton (approved):
Thank you for doing this! tested on staging with loan 1344 that has > 1000 txns and it's working well.

PR #823, opened 2020-12-30
11350 biz plan outline AND 11369 optional qs biz plan tree

DevneyHamilton:
Handles optional questions in the outline links; adds styling. Makes 11350 releasable. 

Screen Shot 2020-12-30 at 2 31 55 PM
Screen Shot 2020-12-30 at 2 17 10 PM

### DIFF
--- a/app/assets/javascripts/admin/loan_questionnaires_view.coffee
+++ b/app/assets/javascripts/admin/loan_questionnaires_view.coffee
@@ -59,8 +59,7 @@ class MS.Views.LoanQuestionnairesView extends Backbone.View
         $question = @$(".question[data-id=#{node.id}]")
         $li.attr('data-id', node.id)
             .addClass($question.attr('class'))
-
-        if node.id == 'optional_group'
+        if node.id.toString().includes('optional_group')
           $li.addClass('optional-group')
         else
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/admin/loan_questionnaires_view.coffee
```diff
@@ -58,31 +59,31 @@ class MS.Views.LoanQuestionnairesView extends Backbone.View
...
     # Recurse, depth first
     for node in nodes
       if node.children?.length
-        node.children = @groupOptional(node.children)
+        node.children = @groupOptional(node.children, node.id)
 
     if nodes.some( (el) -> el.optional ) && !nodes.every( (el) -> el.optional )
       # Add optional group to this level
-      nodes.push { id: 'optional_group', name: optionalGroupName, children: [] }
+      console.log(parentId)
```
meltheadorable: should probably remove the debugging statements, or add more text to them so that the console isn't just filling with random numbers 😉

meltheadorable (approved):
This all looks good, I'll QA it on staging when we merge & push it so that we can get any bugfixes into tickets for next sprint, but I don't have time to get an environment set up to really test this

PR #824, opened 2020-12-30
11368: use 'user_created' to determine if txn coming from m…

DevneyHamilton:
Some historical qb disbursements in TWW QB don't have the qb vendor set. That is ok. Eden requested that we require vendor to be set on disbursements created in the Madeline txn form, however. I used user_created, an attr set in the controller #create and #update methods, to distinguish these in the validation step. After QA on staging, I added the same 'user_created' condition for validating presence of  check number.


meltheadorable:
The description of this PR and the code I'm looking at are not the same, so I'm hesitant to approve this in case the wrong thing got pushed, otherwise it's fine for now

DevneyHamilton:
> The description of this PR and the code I'm looking at are not the same, so I'm hesitant to approve this in case the wrong thing got pushed, otherwise it's fine for now

@meltheadorable I'm sorry I forgot to update descriptions after I changed strategy here. The description is updated now and matches the code now.

PR #825, opened 2020-12-30
11378: fix respect_closed_book_date validation and set user_created i…

DevneyHamilton:
…n update method in controller so called when editing txns. 

before this, if I edited a txn to have a date before closed books date, i'd see an error from the txn builder about missing line items (because we don't update txns before closed books date). Now i'll see the correct notice that the date can't be before closed books date. 
Screen Shot 2020-12-30 at 5 21 33 PM

### DIFF
--- a/app/models/accounting/transaction.rb
+++ b/app/models/accounting/transaction.rb
@@ -246,7 +246,8 @@ class Accounting::Transaction < ApplicationRecord
   end
 
   def respect_closed_books_date
-    return if division.closed_books_date.nil? || txn_date > division.closed_books_date
-    errors.add(:txn_date, :closed_books_date, date: division.closed_books_date)
+    cbd = Loan.find(project_id).qb_division.closed_books_date
+    return if cbd.nil? || txn_date > cbd
+    errors.add(:txn_date, :closed_books_date, date: cbd)
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/accounting/transaction.rb
```diff
@@ -246,7 +246,8 @@ def net_debit_for_account(account_id)
   end
 
   def respect_closed_books_date
-    return if division.closed_books_date.nil? || txn_date > division.closed_books_date
-    errors.add(:txn_date, :closed_books_date, date: division.closed_books_date)
+    cbd = Loan.find(project_id).qb_division.closed_books_date
```
meltheadorable: why are we doing the explicit find here?

DevneyHamilton: closed_books_date was coming back nil if I just use `loan`. I don't know why but I assume its because its in the validation step and the record hasn't been saved yet. The project_id attr is available, but ActiveRecord associations don't seem to be.

meltheadorable: there should be a way to do this without the explicit find but i don't have the setup to test this locally right now to validate my theory so this is fine, though a TODO: comment reflecting we should try removing that would be good just so that doesn't get lost until i've got time to muck around with it

meltheadorable (approved):

app/models/accounting/transaction.rb
DevneyHamilton: @meltheadorable I tried this again and I am able to use the association, so I'm doing that. I must have gotten mixed up when I was testing. Thanks for catching this!

PR #830, opened 2021-01-22
11349: use filename not identifer on media item

DevneyHamilton:
Second PR from this branch to fix "media#index (ActionView::Template::Error) "undefined method `identifier' for #<CarrierWave::Storage::" error. Use 'filename' instead of 'identifier.


DevneyHamilton:
Merging based on local rspec pass

PR #831, opened 2021-01-25
11349: make num cols responsive in media gallery

DevneyHamilton:


DevneyHamilton:
merging after Melody qa'ed and confirmed via slack

PR #832, opened 2021-01-28
11330 AND 11405 edit mode logic 2

DevneyHamilton:
Changes in outline (11405):
- include all levels
- collapse/expand outline
Screen Shot 2021-01-28 at 4 36 15 PM

Changes to edit UI:

- remove top-level "Edit Responses" link
- hide top-level "Delete All Responses" link
- Screen Shot 2021-01-28 at 4 35 54 PM

- have a pencil icon with "Enter Edit Mode" hover text on every group & question
- Screen Shot 2021-01-28 at 4 36 28 PM

- in edit mode,sticky edit bar has "! Now Editing'
- in edit mode,  text questions that have a rich text modal have "Edit Text" link to the right
- Screen Shot 2021-01-28 at 4 36 56 PM
- after making a change to any question and before saving, see 'pending changes' in sticky edit bar
Screen Shot 2021-01-28 at 4 37 08 PM


app/views/admin/loans/_questions.html.slim
```diff
@@ -15,17 +15,13 @@ section.questionnaires
...
-            i.fa.fa-pencil.fa-large>
-            = t(".edit")
-          a.show-action.form-element href="javascript:void(0)"
-            = t("cancel_edit")
-          = link_to [:admin, @response_set], method: :delete,
-            data: { confirm: t("loan.confirm_response_deletion") }
-            i.fa.fa-trash.fa-large>
-            = t(".delete")
+      //hide delete link and see if there are any complaints before removing entirely
+      / .show-actions
```
meltheadorable, line 19: I think we should leave the method of entering edit mode  at the top here, even if we're getting rid of delete, I think people may be used to this method.

app/assets/stylesheets/admin/questions/_questions.scss
```diff
@@ -1,3 +1,22 @@
+.now-editing {
+  color: $red2;
```
DevneyHamilton: I think I need to update these $red2s to actually be a division custom color

meltheadorable: There are probably a few other places that's a thing, if this is an easy change to do now go ahead, otherwise I think we can take it on during a general pass to find these sorts of inconsistencies

meltheadorable (approved):

DevneyHamilton:
merged after running tests on local

PR #833, opened 2021-01-29
11433 improve biz plan loading time

DevneyHamilton:
This doesn't improve biz planning loading time per se, but will keep it from getting worse as updates to the business planning page make more calls to `children`, `visible?`, and `required` methods for questions.


DevneyHamilton:
rspec is passing on local

meltheadorable (approved):

PR #835, opened 2021-02-04
11406 biz planning viz updates

DevneyHamilton:
This issue has a few visual updates designed in https://redmine.sassafras.coop/issues/11224. The biggest one is making question instructions collapsible. The original design doesn't work well with html instructions, so in consultation with mariam, I implemented a version where 'long' instructions (defined by a number of characters) are collapsible with a left-hand caret. This is a compromise that minimizes the number of carets on the page, keeps at least the first line of instructions visible, and keeps any one set of instructions from taking up too much space unless it's expanded. Note that 'short' instructions that are not collapsible may wrap to two lines on small screens or three lines in a very narrow window

edit mode, all collapsibles collapsed . . .
Screen Shot 2021-02-03 at 4 36 07 PM

edit mode, collapsibles expanded
Screen Shot 2021-02-03 at 4 37 01 PM


app/assets/javascripts/admin/loan_questionnaires_view.coffee
```diff
@@ -32,6 +32,16 @@ class MS.Views.LoanQuestionnairesView extends Backbone.View
     'click .edit-rt-response': 'openRichTextModal'
     'click a.deep-link': 'scrollElementToView'
     'change input': 'formUpdated'
+    'click .help-block .expansion-control': 'toggleExpansion'
+
```
DevneyHamilton, line 36: expanding/collapsing instructions is different than other expand/collapse in madeline. The difference is that rather than hiding/showing and element, we are changing the styles on the element. We change the styles on the element because we're previewing the top line, and putting 'the rest' into a separate element is extra, messy work.

meltheadorable (approved):

DevneyHamilton:
specs passing on local

PR #838, opened 2021-02-10
11413: Logs Landing Page - First Draft

meltheadorable:


DevneyHamilton (commented):
Cool! One issue I found in QA is that if I edit the log from the log landing page, then after the update saves, I see the title of the log twice. Should we fix that before deploying? If so, do you want to do that @meltheadorable or should I? I'm happy to on Thursday or Friday.  Screen Shot 2021-02-10 at 10 34 20 AM

PR #849, opened 2021-03-23
11665: Fix reflected url param issue

smoyte:
I have not QA'd this. Both Basic Projects and Loans various tabs should be tested.


DevneyHamilton:
qa'ed on local, will again on staging

PR #853, opened 2021-04-01
11676 add specs for biz plan edit

DevneyHamilton:
This PR adds a high level spec to questionnaire (existing spec had been marked 'skip' for years, maybe since summernote added?).  

It fixes two issues:
1) in edit mode, an unanswered question now shows "This is a required question" instead of 'no answer.' This was an oversight on my part implementing the original story
2) re-does custom_dark_accent, in custom colors, which was merge and somehow got undone. 

This issue also renames "Edit Text" to "Edit Answer" which Brendan said would be good in our meeting, and it was very quick to change. 

See  mock https://redmine.sassafras.coop/attachments/791 for reference

### DIFF
--- a/app/views/admin/loans/questionnaires/_answer.html.slim
+++ b/app/views/admin/loans/questionnaires/_answer.html.slim
@@ -54,7 +54,9 @@
     .view-element.answer = sanitize(response.text, tags: tags, attributes: attrs)
 
   .form-element.answer.rt-answer = sanitize(response.text, tags: tags, attributes: attrs)
-  .form-element.answer.no-response class=(!response.text.present? ? "" : "hidden") = t("loan.no_answer")
+  .form-element.answer.blank.no-response.custom_dark_accent class=(!response.text.present? ? "" : "hidden")
+      = fa_icon "exclamation-circle"
+      = t("questions.missing_required_answer")
 
   = f.input_field :"#{question.attribute_sym}[text]",
     value: sanitize(response.text, tags: tags, attributes: attrs), as: :hidden, class: "rt-response"
--- a/app/views/admin/loans/questionnaires/_answer.html.slim
+++ b/app/views/admin/loans/questionnaires/_answer.html.slim
@@ -54,9 +54,8 @@
     .view-element.answer = sanitize(response.text, tags: tags, attributes: attrs)
 
   .form-element.answer.rt-answer = sanitize(response.text, tags: tags, attributes: attrs)
-  .form-element.answer.blank.no-response.custom_dark_accent class=(!response.text.present? ? "" : "hidden")
-      = fa_icon "exclamation-circle"
-      = t("questions.missing_required_answer")
+  .form-element.answer.no-response class=(!response.text.present? ? "" : "hidden")
+     = t("loan.no_answer")
 
   = f.input_field :"#{question.attribute_sym}[text]",
     value: sanitize(response.text, tags: tags, attributes: attrs), as: :hidden, class: "rt-response"
--- a/app/views/layouts/admin/_custom_colors.html.slim
+++ b/app/views/layouts/admin/_custom_colors.html.slim
@@ -7,7 +7,13 @@
 
 style type="text/css"
 
-  | /* General links */
+
+  | /* For customizing */
+    .custom_dark_accent {
+      color: #{colors[:accent_darkened]};
+    }
+    
+   /* General links */
     a,
     .btn-link,
     .table.rowlink td:not(.rowlink-skip) a,
--- a/config/locales/en/questions.en.yml
+++ b/config/locales/en/questions.en.yml
@@ -8,7 +8,7 @@ en:
     currently_editing: "Now editing"
     delete_error: "The item could not be deleted due to an error"
     edit_all: "Enter Edit Mode"
-    edit_rich_text: "Edit Text"
+    edit_rich_text: "Edit Answer"
     explanation: "Explanation"
     manage: "Manage Loan Questions"
     missing_required_answer: "This is a required question."
--- a/spec/features/admin/questionnaire_spec.rb
+++ b/spec/features/admin/questionnaire_spec.rb
@@ -18,26 +18,24 @@ feature 'questionnaire', js: true do
 
       # check that edit-all button turns on edit mode
       expect(page).not_to have_content("Now editing")
-      page.find(".edit-all", match: :first).click
+      first(".edit-all").click()
       expect(page).to have_content("Now editing")
 
       # cancel button is visible in edit mode
-      edit_bar = page.find("#editBar")
-      edit_bar.find(".cancel-edit").click()
+      first("#editBar .cancel-edit").click
       expect(page).not_to have_content("Now editing")
 
       # save changes button is visible in edit mode
-      page.find(".edit-all", match: :first).click
+      first(".edit-all").click()
       expect(page).to have_content("Now editing")
-      click_button 'Save Changes'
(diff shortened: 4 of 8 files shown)


app/views/admin/loans/questionnaires/_answer.html.slim
```diff
@@ -54,7 +54,9 @@
     .view-element.answer = sanitize(response.text, tags: tags, attributes: attrs)
 
   .form-element.answer.rt-answer = sanitize(response.text, tags: tags, attributes: attrs)
-  .form-element.answer.no-response class=(!response.text.present? ? "" : "hidden") = t("loan.no_answer")
+  .form-element.answer.blank.no-response.custom_dark_accent class=(!response.text.present? ? "" : "hidden")
+      = fa_icon "exclamation-circle"
+      = t("questions.missing_required_answer")
```
DevneyHamilton, line 59: show the "This is a required question" instead of 'no answer' according to these mocks: https://redmine.sassafras.coop/attachments/791

app/views/layouts/admin/_custom_colors.html.slim
```diff
@@ -7,7 +7,13 @@
 
 style type="text/css"
 
-  | /* General links */
+
+  | /* For customizing */
+    .custom_dark_accent {
+      color: #{colors[:accent_darkened]};
+    }
+
```
DevneyHamilton, line 15: these had been merged before, but then got lost from develop somehow, which is why some things were gray instead of red as in the mocks here: https://redmine.sassafras.coop/attachments/791

config/locales/en/questions.en.yml
```diff
@@ -8,7 +8,7 @@ en:
     currently_editing: "Now editing"
     delete_error: "The item could not be deleted due to an error"
     edit_all: "Enter Edit Mode"
-    edit_rich_text: "Edit Text"
+    edit_rich_text: "Edit Answer"
```
DevneyHamilton, line 11: in-meeting request from brendan

app/views/layouts/admin/_custom_colors.html.slim
smoyte: Hmm, strange. Is it worth doing some forensic work to see what happened or no?

spec/features/admin/questionnaire_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 # TODO: look for a way around targeting summernote textarea
-xfeature 'questionnaire', js: true do
+feature 'questionnaire', js: true do
```
smoyte, line 4: Hooray for unskipping specs!

spec/features/admin/questionnaire_spec.rb
```diff
@@ -15,66 +15,38 @@
...
-      expect(criteria.summary.text).to eq 'dragon'
-    end
-
-    it "overwrite button works" do
-      find('input[type="submit"][name="overwrite"]').click
-      expect(page).to have_content 'successfully updated'
-      expect(page).to have_content 'gnashing teeth'
-      expect(criteria.summary.text).to eq 'gnashing teeth'
-    end
+    # TODO reimplement
```
smoyte, line 44: Are we cutting this code out b/c the underlying feature is no longer there or b/c the test doesn't work or some other reason?

spec/features/admin/questionnaire_spec.rb
```diff
@@ -15,66 +15,38 @@
...
-      # delete
-      accept_confirm { click_link('Delete All Responses') }
-      expect(page).to have_content 'successfully deleted'
-      expect(page).not_to have_content 'sexy unicorns'
-      # After deletion, should be in edit mode
-      expect(page).not_to have_content 'Edit Responses'
-      expect(page).to have_selector 'input[value="Save Responses"]'
+      # check that edit-all button turns on edit mode
+      expect(page).not_to have_content("Now editing")
+      page.find(".edit-all", match: :first).click
```
smoyte: I think you can write `first(".edit-all").click` to achieve the same thing, which is rather less verbose. Have you tried that? If not can you see if it works?

If it does, same simplifications throughout please.

spec/features/admin/questionnaire_spec.rb
```diff
@@ -15,66 +15,38 @@
...
+      expect(page).to have_content("Now editing")
+
+      # cancel button is visible in edit mode
+      edit_bar = page.find("#editBar")
+      edit_bar.find(".cancel-edit").click()
```
smoyte: These two lines can be `first("#editBar .cancel-edit").click` I believe.

spec/features/admin/questionnaire_spec.rb
```diff
@@ -15,66 +15,38 @@
...
+      expect(page).not_to have_content("Now editing")
+
+      # save changes button is visible in edit mode
+      page.find(".edit-all", match: :first).click
+      expect(page).to have_content("Now editing")
+      click_button 'Save Changes'
```
smoyte: I'd like us to be consistent about parenthesis usage please. Would you mind adding some here. Later on I'll likely propose we add some rubocop rules for this so it can be caught while we are all coding and not have to take up time in PRs.

spec/features/admin/questionnaire_spec.rb
```diff
@@ -15,66 +15,38 @@
...
+      expect(page).not_to have_content("Now editing")
+
+      # test outline expansion
+      outline = page.find(".outline")
+      expect(page).not_to have_content("Outline")
+      outline.find(".expander").click()
```
smoyte: Please remove the `()`. Rubocop will be useful here too.

spec/features/admin/questionnaire_spec.rb
```diff
@@ -15,66 +15,38 @@
...
+      expect(page).to have_content("Outline")
+      outline.find(".hider").click()
+      expect(page).not_to have_content("Outline")
```
smoyte, line 39: Nice job on the spec! Reads very clear.

smoyte (commented):
Nice work! Clean and easy to read spec. Bonus points for throwing in the fix Brendan requested. I love the 'get it done' energy! I have a few questions and some linting requests. Thank you!

spec/features/admin/questionnaire_spec.rb
DevneyHamilton: this spec was also skipped, and doesn't work. we could make an issue to do the work to reinstate it . . .

spec/features/admin/questionnaire_spec.rb
DevneyHamilton: cool, thanks for this pointer

app/views/layouts/admin/_custom_colors.html.slim
DevneyHamilton: yeah, i found the commit where it got lost. i was manually merging a release branch that's outside of our normal release flow, back in early feb.  So getting our release process smoothed out like we're talking about on wed next wk will help make that not happen again.

smoyte:
The "Edit Answer" change looks good. Color change looks good too.

I don't think we should change "No Answer" to "This is a required question". The result is inconsistent. See below, in view mode:

image

and in edit mode:

image

Both are required questions, and one says 'required' in show mode, while the other says 'required' in edit mode.

I think we need to rethink the design here holistically. 

For the meantime, it might help to keep "No Answer" but add square brackets "[No Answer]" and perhaps make it a bit grayer to give it less visual weight? Do we have a de-emphasized text color variable in the stylesheets?

Thoughts?

PR #855, opened 2021-04-07
11706: Fix NoMethodError during validation when country not selected

meltheadorable:


app/models/organization.rb
```diff
@@ -57,7 +57,7 @@ class Organization < ApplicationRecord
   validates :name, :division, :country, presence: true
 
   validate :primary_contact_is_member
-  with_options if: ->(org) { org.country.iso_code == "US" } do |us_org|
+  with_options if: ->(org) { org&.country&.iso_code == "US" } do |us_org|
```
smoyte, line 60: ~~I don't believe the first `&` is needed, is it?~~

~~In general I am leery of two &.'s in a row due to Law of Demeter. In fact, two .'s is typically frowned upon, isn't it? Here it seems ok since `org` really just means `self` since it's a callback thing and really `country` is an attribute of Organization. So LoD doesn't really feel violated here.~~

~~But still, the fewer &.'s, the better in my book. And `org` should never be nil, right?~~

spec/models/organization_spec.rb
```diff
@@ -73,6 +73,13 @@
     let!(:country_us) { create(:country, iso_code: 'US') }
     let!(:country_not_us) { create(:country, iso_code: 'AR') }
 
+    describe "country" do
+      it "should be required" do
+        org = build(:organization, country: nil)
+        expect(org.valid?).to be false
```
smoyte: This can be shortened to `expect(org).to be_invalid`.

app/models/organization.rb
smoyte: Can we actually just add a non-null constraint to country_id? I see this kind of thing a lot in Madeline where safe navigation is being used on columns that shouldn't be null but are missing a non-null constraint just due to them not being included in the data model when it was first built out. I'd like us to start adding these bit by bit because they reduce errors like this and make code simpler. What do you think @meltheadorable and also @DevneyHamilton?

If we add the constraint, the &.s shouldn't be needed anymore.

smoyte: I checked the prod database and there was only 4 orgs with missing country IDs so I did a bit of sleuthing and set them manually. (3 mexico, 1 US). So the constraint shouldn't fail, at least on prod.

smoyte (commented):
Thanks Mel for the quick response and the quick spec!

I think this would fix the immediate problem but I want to explore a different approach here and in general. Please let me know your thoughts.

Thanks!

DevneyHamilton (commented):
Nothing to add here. I remember when we talked about it sync @meltheadorable and @smoyth you all said this'll have both the validation change and the db level constraint, right?

meltheadorable:
I added the not null constraint but just wanted to comment as a reminder that the safe navigation is still needed because this code is run during validation and we don't want it causing 500 errors instead of telling the user why validation failed

smoyte:
@meltheadorable I'm with you on that for the second `&.`. But how can `org` ever be nil? Wouldn't that be a bug in Rails?

The fewer &.'s the better in my book.

smoyte:
And thanks for adding the not-null!

PR #858, opened 2021-04-09
11608 redirect existing public pages

DevneyHamilton:
This PR covers multiple issues:
11614 allow admins to update division URL
11613 add link to public page to admin view
11611 use division custom colors
11610 Use division shortname instead of ID (in the url)
11609 change permissions (for public division pages)
11608 redirect existing public pages
11606 make public layout without WP
The page currently shows all active and completed loans under the heading "Loans"

TODOs within the scope of these issues:
merge with gem updates on develop
Improve specs on loan and division policies
Have a fallback image for loans that do not have a featured image

Outstanding questions about getting this ready for release:
- does TWW & La Base Argentina & La Base Nicaragua actually need more work on these new pages before we deploy & get rid of existing public pages?  [will depend on brendan's rxn to demo]
- should we make a task to make shortnames for all divisions based on their division name? or leave that to admins (all divisions in prod have a shrotname but most have a UUID as part of it)
- should we do a simple version of a loans page


app/helpers/application_helper.rb
```diff
@@ -93,13 +93,40 @@ def app_version_number
     @app_version_number ||= version
   end
 
-  def division_select_options(include_root: true, include_all: false, public_only: true)
+  def division_select_options(include_root: true, include_all: false, public_only: false)
```
DevneyHamilton, line 96: this flag was set to false everywhere except the division dropdown on public pages. we may want to bring in the division dropdown based on what we learn from brendan in demo tmrw.

DevneyHamilton:
Closed in favor of https://github.com/sassafrastech/madeline/pull/901 which built off of and replaced this

PR #865, opened 2021-05-01
11760 admin cant see problem loan txns

DevneyHamilton:
The transaction policy refactor had a regression where no one could see problem loan transaction information. Problem loan transactions are not a subclass of transactions. PLTs reports errors madeline had processing txns imported from QB. The primary audience for these are us (the development team), and highly skilled madeline admins (e.g. Eden and Zahuna). Making them useful to other people would need design work.  An earlier commit of this branch is deployed on staging so that Eden & Zahuna could test on Thurs & Friday. This branch is a small improvement, with root division admins able to see  spec coverage on the separate problem loan transaction policy. In this version, root admins can see the index and show for problem loan transactions and that is now documented in the specs. 

For accounting, we should show non root admins an alternate version of the Screen Shot 2021-04-30 at 8 20 42 PM "There are n warnings associated with transactions for this loan. Your loan balances may not be correct. Please review the transactions with warnings, and only proceed once you have confirmed they are correct. " warning and similar error. I would love to chat w Tom or Melody about the best way to do that in the loan controller. 

For division-based accounting we should revisit problem loan transactions. 

Note: if this branch is not merged before release, Zahuna and Eden will experience a regression on staging where they cannot click through to transaction warnings & errors.

### DIFF
--- a/app/controllers/admin/accounting/problem_loan_transactions_controller.rb
+++ b/app/controllers/admin/accounting/problem_loan_transactions_controller.rb
@@ -14,8 +14,8 @@ module Admin
       end
 
       def show
-        authorize :'accounting/problem_loan_transaction', :show?
         @plt = ::Accounting::ProblemLoanTransaction.find(params[:id])
+        authorize @plt
       end
     end
   end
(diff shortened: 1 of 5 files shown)


meltheadorable (approved):
This looks fantastic, thanks.

PR #869, opened 2021-05-18
10317 starting balance not imported

DevneyHamilton:
We needed to upgrade quickbooks_ruby, which was breaking on a certain txn. This branch currently on staging so I could continue looking into issues with that loan (3293).


DevneyHamilton:
renamed branch to 11781_error_in_qb_ruby

PR #870, opened 2021-05-19
11781 error in qb ruby

DevneyHamilton:
We needed to upgrade quickbooks_ruby, which was breaking on certain txns where older version of quickbooks_ruby was erring on a call to BigDecimal.new in dependency, roxml. (see https://github.com/Empact/roxml/releases/tag/v4.1.0) This branch currently on staging so I could continue looking into issues with that loan (3293).

### DIFF
--- a/app/views/admin/project_logs/_log.html.slim
+++ b/app/views/admin/project_logs/_log.html.slim
@@ -52,7 +52,7 @@
   - if log.summary
     .summary
       h5.sr-only = t('common.summary')
-      span = strip_tags(log.summary)
+      span = strip_tags(log.summary.to_s)
 
   - if log.has_more? && @context != 'dashboard'
     div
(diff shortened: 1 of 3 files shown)


app/views/admin/project_logs/_log.html.slim
```diff
@@ -52,7 +52,7 @@
   - if log.summary
     .summary
       h5.sr-only = t('common.summary')
-      span = strip_tags(log.summary)
+      span = strip_tags(log.summary.to_s)
```
smoyte: What's this little sneaker doing stowing away in here? :D Can we clean up the commit history quickly so this isn't part of the PR? I can help with this if you like.

smoyte:
This looks fine otherwise! Thanks for fixing!

PR #872, opened 2021-05-21
11798: Change to flexbox to fix questionnaire positioning issue

smoyte:
Includes a reorg of the questionnaire styles and some comments about what is in which file.


DevneyHamilton:
I tested this in firefox & chrome on mac and it's looking good. Thank you @smoyth

PR #873, opened 2021-05-21
11799 notify on bg error

smoyte:
NOTE: Best to review this one commit-by-commit to see what actually changed in the job.

* Removed Sentry for now b/c the gem was deprecated, we weren't actually using it in the code properly yet, and I am a bit concerned about the extra complexity it's adding at a time when we are trying to stabilize our error handling flows
* Unified two very similar QB related job classes
* Removed explicit error message from Task UI. For now this means it will always show "system error. devs notified" for all loans in the list. I think we need to rethink the tasks page error display along with the problem loan transactions thing. It could be that in the future that area could hold error information potentially of use to the user that is not related to a particular transaction. So I didn't want to rip it out completely at this point.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,7 +22,6 @@ export QB_SANDBOX_OAUTH2_CLIENT_SECRET=xxx
 # Set this to 1 or 0 for manual override.
 export QB_SANDBOX_MODE=1
 
-export SENTRY_DSN=
 export SCOUT_KEY=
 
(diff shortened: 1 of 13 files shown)


DevneyHamilton (approved):
Thank you @smoyth! this'll make testing these jobs easier too. Do you want to add specs for those (we'd latest assigned that task to me, but I would need help on how to go about it in any case)? Manually testing these jobs is a massive time sink.

PR #874, opened 2021-05-24
Arg migration support

smoyte:
These are some small things that are worth keeping from the arg. migration, and some of them (schema changes) will be needed to do the full migration. So not a huge rush to deploy this but I'd like to get it in the next release. Thanks!

### DIFF
--- a/app/models/concerns/notable.rb
+++ b/app/models/concerns/notable.rb
@@ -3,7 +3,7 @@ module Notable
 
   included do
     # notes that are added to organizations or people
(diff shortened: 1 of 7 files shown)


DevneyHamilton (approved):
thank you for these - they look like they won't cause issues for the rest of the code base. I didn't follow "So not a huge rush to deploy this but I'd like to get it in the next release. " - is there any reason not to just merge this into develop?

smoyte:
Nope!

PR #883, opened 2021-06-07
11826: allow members to see transactions index

DevneyHamilton:
members (non admins) should be able to see the transactions tab for any loan they can see.

### DIFF
--- a/app/policies/accounting/transaction_policy.rb
+++ b/app/policies/accounting/transaction_policy.rb
@@ -1,6 +1,7 @@
 class Accounting::TransactionPolicy < ApplicationPolicy
   def index?
-    machine_user_or_appropriate_division_admin?
+    member_level_access = user && user != :machine && user.accessible_division_ids.include?(division.id)
+    member_level_access || machine_user_or_appropriate_division_admin?
   end
 
   def show?
--- a/spec/policies/accounting/transaction_policy_spec.rb
+++ b/spec/policies/accounting/transaction_policy_spec.rb
@@ -22,7 +22,7 @@ describe Accounting::TransactionPolicy do
 
   context "with non-admin" do
     context "user has access to this division" do
-      let(:user) { create_member(division) }
+      let(:user) { create(:user, :member, division: division) }
       forbid_all_but_index
       it_behaves_like "returns no reasons even if issues other than user role"
     end
--- a/spec/policies/accounting/transaction_policy_spec.rb
+++ b/spec/policies/accounting/transaction_policy_spec.rb
@@ -21,9 +21,17 @@ describe Accounting::TransactionPolicy do
   end
 
   context 'with non-admin' do
-    let(:user) { create(:user) }
-    forbid_all
-    it_behaves_like 'returns no reasons even if issues other than user role'
+    context 'user has access to this division' do
+      let(:user) { create_member(division) }
+      forbid_all_but_index
+      it_behaves_like 'returns no reasons even if issues other than user role'
+    end
+
+    context 'user does not have access to this division' do
+      let(:user) { create(:user) }
+      forbid_all
+      it_behaves_like 'returns no reasons even if issues other than user role'
+    end
   end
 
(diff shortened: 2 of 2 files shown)


spec/policies/accounting/transaction_policy_spec.rb
```diff
@@ -11,83 +11,91 @@
...
-    it_behaves_like 'returns no reasons even if issues other than user role'
+    it_behaves_like "returns no reasons even if issues other than user role"
   end
 
-  context 'with admin of division but not parent_division' do
+  context "with admin of division but not parent_division" do
     let(:user) { create(:user, :admin, division: division) }
-    forbid_all
-    it_behaves_like 'returns no reasons even if issues other than user role'
+    forbid_all_but_index
```
DevneyHamilton, line 45: logic change here: an admin on a division should be able to see the loan index for loans on that division

spec/policies/accounting/transaction_policy_spec.rb
```diff
@@ -11,83 +11,91 @@
...
       let(:loan_trait) { :completed }
       it { expect(policy.read_only_reasons).to be_empty }
     end
   end
 
-  context 'with non-admin' do
-    let(:user) { create(:user) }
-    forbid_all
-    it_behaves_like 'returns no reasons even if issues other than user role'
+  context "with non-admin" do
```
DevneyHamilton, line 23: logic change: allow members to see txn indexes for loans they have access to

app/policies/accounting/transaction_policy.rb
```diff
@@ -1,6 +1,7 @@
 class Accounting::TransactionPolicy < ApplicationPolicy
   def index?
-    machine_user_or_appropriate_division_admin?
+    member_level_access = user && user != :machine && user.accessible_division_ids.include?(division.id)
```
smoyte, line 3: Two things here:

1. It seems a bit off to me to have the User model responsible for determining which divisions it has access to. Seems like a violation of SRP. One approach would be to just spell out `DivisionPolicy::Scope.new(user, Division).accessible_ids`. But even that, in a way, seems a bit odd—get a list of IDs of divisions the user has access to and then check if this division is in it. Should that logic just be encapsulated within DivisionPolicy? I'd expect to see something link `DivisionPolicy.new(user, division).access?`. (It appears `access?` is different than `show?`, please confirm this is correct.) Could we try that, and later (not now) try to phase out the `accessible_division` methods from `User`?

2. To match the general style of Policy classes, I think you could put this line in a private method just like `machine_user_or_appropriate_division_admin?` and then you wouldn't need the local var.

spec/policies/accounting/transaction_policy_spec.rb
```diff
@@ -11,83 +11,91 @@
...
+    context "user has access to this division" do
+      let(:user) { create_member(division) }
```
smoyte: I see there are `:member` and `:admin` traits on the `User` factory. Is there a reason not to use those and perhaps phase out `create_member` and `create_admin`?

spec/policies/accounting/transaction_policy_spec.rb
```diff
@@ -11,83 +11,91 @@
...
+      forbid_all_but_index
+      it_behaves_like "returns no reasons even if issues other than user role"
+    end
+
+    context "user does not have access to this division" do
+      let(:user) { create(:user) }
+      forbid_all
+      it_behaves_like "returns no reasons even if issues other than user role"
+    end
```
smoyte, line 34: Very nice use of contexts! Reads really clearly.

spec/policies/accounting/transaction_policy_spec.rb
smoyte: Thanks for pointing these out.

smoyte (commented):
Looks great to me logic-wise. Just a few tweaks recommended in the policy. Let me know if you have questions or concerns. Thank you!

PR #884, opened 2021-06-07
11825: remove active scope from fetch qb changes job, rely on txn pol…

DevneyHamilton:
There is a non-active loan that Eden updated in QB. She needs to be able to see those changes in Madeline. 

The quickest way to support this is to remove the active scope on the fetch changes job. We want to do this in general, since Madeline should know what QB knows about any loans with transactions, regardless of whether they are active. That was there as a precaution before the big transactions policy refactor which does the job of making sure we don't run interest calculation on non-active loans. 

We need to add a spec for the job, and I'm glad to pair on that this week. I do want to get this up on prod today so Eden can fix this last thing.

UPDATE: realized the specific qb change should have come in thru a nightly update, so I am also looking into that.

### DIFF
--- a/app/jobs/fetch_quickbooks_changes_job.rb
+++ b/app/jobs/fetch_quickbooks_changes_job.rb
@@ -3,7 +3,7 @@ class FetchQuickbooksChangesJob < QuickbooksUpdateJob
 
   def loans
     @loans ||= divisions.map do |division|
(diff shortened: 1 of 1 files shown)


DevneyHamilton:
closing in favor of https://github.com/sassafrastech/madeline/pull/885

PR #885, opened 2021-06-07
11825 new qb data not coming in

DevneyHamilton:
There is a non-active loan that Eden updated in QB. She needs to be able to see those changes in Madeline. There are two ways these changes should be available, but there are problems with each. This pr has immediate fixes for these.

fetch changes from Quickbooks link: This link had an 'active' scope on it.  That was there as a precaution before the big transactions policy refactor which does the job of making sure we don't run interest calculation on non-active loans. We do want to pull QB changes on non-active loans, since Madeline should know what QB knows about any loans with transactions, regardless of whether they are active.  [we still need to do a spec on these jobs, I can work on that tmrw with some pairing]

nightly update all loans job: This job has not been running on production. This is most likely because we failed to undo a workaround in the schedule.rb file that was waiting for the production server upgrade which has happened -  https://github.com/sassafrastech/madeline/pull/885

### DIFF
--- a/app/jobs/fetch_quickbooks_changes_job.rb
+++ b/app/jobs/fetch_quickbooks_changes_job.rb
@@ -3,7 +3,7 @@ class FetchQuickbooksChangesJob < QuickbooksUpdateJob
 
   def loans
     @loans ||= divisions.map do |division|
-      division.loans.changed_since(updater.qb_connection.last_updated_at).active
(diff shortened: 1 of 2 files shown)


smoyte (commented):
Looks good to me! Thanks for the explanation and for continuing to think about tests! I look forward to pairing with you on that.

PR #889, opened 2021-07-07
11831 fix error updating non madeline txn

DevneyHamilton:
In response to bug 11831, we reviewed spec coverage and the use of 'managed.' The root cause of bug  11831 was that the interest calculator assumes only managed txns are edited or written, but we were allowing admins to edit ANY disbursement or repayment, including non-managed ones. 
This PR restricts editing txn to only managed repayments and disbursements (not unmanaged repayments and disbursements, and no interest txns whether managed or not)

### DIFF
--- a/app/controllers/admin/accounting/transactions_controller.rb
+++ b/app/controllers/admin/accounting/transactions_controller.rb
@@ -1,7 +1,7 @@
 class Admin::Accounting::TransactionsController < Admin::AdminController
   def new
     @loan = Loan.find_by(id: params[:project_id])
-    @transaction = sample_transaction
+    @transaction = ::Accounting::Transaction.new(project: @loan, txn_date: Time.zone.today, managed: true)
     authorize(@transaction, :new?)
     prep_transaction_form
     render_modal_partial
@@ -17,14 +17,16 @@ class Admin::Accounting::TransactionsController < Admin::AdminController
 
   def sync
     @loan = Loan.find(params[:project_id])
-    @transaction = sample_transaction
+    @transaction = ::Accounting::Transaction.new(project: @loan, txn_date: Time.zone.today)
     authorize(@transaction, :show?)
     sync_and_handle_errors
     redirect_to(admin_loan_tab_path(@loan, tab: "transactions"))
   end
 
   def create
-    @transaction = sample_transaction(transaction_params.merge(project_id: params[:project_id]))
+    @transaction = ::Accounting::Transaction.new(
+      transaction_params.merge(project_id: params[:project_id], managed: true)
+    )
     authorize(@transaction, :create?)
     @loan = @transaction.project
     @transaction.user_created = true
@@ -167,10 +169,5 @@ class Admin::Accounting::TransactionsController < Admin::AdminController
 
   def settings_link
     view_context.link_to(t('menu.accounting_settings'), admin_accounting_settings_path)
-  end
-
-  def sample_transaction(attrs = {})
-    most_likely_editable_attrs = {project: @loan, txn_date: Time.zone.today, managed: true}
-    ::Accounting::Transaction.new(most_likely_editable_attrs.merge(attrs))
   end
 end
--- a/app/controllers/admin/accounting/transactions_controller.rb
+++ b/app/controllers/admin/accounting/transactions_controller.rb
@@ -24,7 +24,7 @@ class Admin::Accounting::TransactionsController < Admin::AdminController
   end
 
   def create
-    @transaction = ::Accounting::Transaction.new(transaction_params.merge(project_id: params[:project_id]))
+    @transaction = sample_transaction(transaction_params.merge(project_id: params[:project_id]))
     authorize(@transaction, :create?)
     @loan = @transaction.project
     @transaction.user_created = true
@@ -169,7 +169,8 @@ class Admin::Accounting::TransactionsController < Admin::AdminController
     view_context.link_to(t('menu.accounting_settings'), admin_accounting_settings_path)
   end
 
-  def sample_transaction
-    ::Accounting::Transaction.new(project: @loan, txn_date: Time.zone.today)
+  def sample_transaction(attrs = {})
+    most_likely_editable_attrs = {project: @loan, txn_date: Time.zone.today, managed: true, loan_transaction_type_value: :disbursement}
+    ::Accounting::Transaction.new(most_likely_editable_attrs.merge(attrs))
   end
 end
--- a/app/controllers/admin/accounting/transactions_controller.rb
+++ b/app/controllers/admin/accounting/transactions_controller.rb
@@ -170,7 +170,7 @@ class Admin::Accounting::TransactionsController < Admin::AdminController
   end
 
   def sample_transaction(attrs = {})
-    most_likely_editable_attrs = {project: @loan, txn_date: Time.zone.today, managed: true, loan_transaction_type_value: :disbursement}
+    most_likely_editable_attrs = {project: @loan, txn_date: Time.zone.today, managed: true}
     ::Accounting::Transaction.new(most_likely_editable_attrs.merge(attrs))
   end
 end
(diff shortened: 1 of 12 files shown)


app/controllers/admin/accounting/transactions_controller.rb
```diff
@@ -169,7 +169,8 @@ def settings_link
     view_context.link_to(t('menu.accounting_settings'), admin_accounting_settings_path)
   end
 
-  def sample_transaction
-    ::Accounting::Transaction.new(project: @loan, txn_date: Time.zone.today)
+  def sample_transaction(attrs = {})
+    most_likely_editable_attrs = {project: @loan, txn_date: Time.zone.today, managed: true}
```
smoyte: Above we are passing in `project_id` and supposedly expecting it to override `project`, but it's not clear to me if that is reliable. Can we make both use `project`?

app/controllers/admin/loans_controller.rb
```diff
@@ -54,7 +54,7 @@ def show
         prep_logs(@loan)
       when "transactions"
         # requires additional level of validation beyond approrpiate loan access
-        @sample_transaction = ::Accounting::Transaction.new(project: @loan)
+        @sample_transaction = ::Accounting::Transaction.new(project: @loan, managed: true)
```
smoyte, line 57: Nice

app/models/accounting/interest_calculator.rb
```diff
@@ -59,7 +59,8 @@ def initialize(loan)
     end
 
     def recalculate
-      return unless TransactionPolicy.new(:machine, Transaction.new(project: @loan)).create?
+      # check if a generic non-interest, managed txn for this loan can be written
+      return unless TransactionPolicy.new(:machine, Transaction.new(project: @loan, managed: true)).create?
```
smoyte, line 63: Nice comment.

app/models/accounting/qb/transaction_class_finder.rb
```diff
@@ -2,8 +2,6 @@ module Accounting
   module QB
     class FindError < StandardError; end
 
-    # This class is responsible for batching up Quickbooks API calls into separate types.
-    # The API does support batch requests for queries, but quickbooks-ruby does not.
```
smoyte, line 6: Thanks!

app/policies/accounting/transaction_policy.rb
```diff
@@ -9,7 +9,7 @@ def show?
   end
 
   def create?
-    machine_user_or_appropriate_division_admin? && read_only_reasons.none?
+    machine_user_or_appropriate_division_admin? && record.managed? && !record.interest? && read_only_reasons.none?
```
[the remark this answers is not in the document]
smoyte: Please fix

app/views/admin/accounting/transactions/_modal_content.slim
```diff
@@ -8,7 +8,7 @@ div.modal-header
       = @transaction.description
 
 div.modal-body
-  - unless @transaction.new_record? || @transaction.type?("interest")
+  - if !@transaction.new_record? && policy(@transaction).update?
```
smoyte: Nice improvement! You can also do `persisted?` to avoid the negation, but this is fine.

spec/features/admin/accounting/transaction_flow_spec.rb
```diff
@@ -250,6 +251,32 @@
...
+        visit admin_loan_tab_path(loan, tab: "transactions")
+        click_on txn.txn_date.strftime("%B %-d, %Y")
+        expect(page).to have_content("Edit")
+      end
+    end
+
+    describe "an admin cannot edit a non-managed txn" do
+      let!(:txn) do
+        create(:accounting_transaction,
+               project_id: loan.id, description: "I love icecream", division: division, managed: false, loan_transaction_type_value: :disbursement)
```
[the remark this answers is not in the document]
smoyte: Please fix. I would do it like this:

```
create(:accounting_transaction, project_id: loan.id, description: "I love icecream", division: division, 
                                managed: false, loan_transaction_type_value: :disbursement)
```

since I prefer not having super tall & narrow blocks. But whatever passes the linter is fine with me.

spec/features/admin/accounting/transaction_flow_spec.rb
```diff
@@ -250,6 +250,45 @@
...
+      end
+
+      scenario do
+        visit admin_loan_tab_path(loan, tab: "transactions")
+        click_on txn.txn_date.strftime("%B %-d, %Y")
+        expect(page).not_to have_content("Edit")
+      end
+    end
+
+    describe "an admin cannot edit a managed interest txn" do
```
smoyte, line 288: I'm not sure if this spec is necessary since the above two already demonstrate that the policy is being applied correctly. The details of the policy fall to the policy spec. Or am I missing something?

spec/models/accounting/qb/purchase_extractor_spec.rb
```diff
@@ -17,7 +17,10 @@
       credit_accounts: [txn_acct],
       type: "Purchase",
       total: 12345.67,
-      doc_number: "from qb")
+      doc_number: "from qb",
+      sync_token: "abc",
+      payment_type: "Check"
+    )
```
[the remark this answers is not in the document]
smoyte: Please fix.

spec/policies/accounting/transaction_policy_spec.rb
```diff
@@ -95,6 +95,18 @@
...
     end
 
+    describe "individual transaction level rules" do
+      context "non-ms-managed transaction" do
+        let(:described_transaction) { Accounting::Transaction.new(project: loan, managed: false) }
+        forbid_all_but_read
+      end
+
+      context "ms-managed interest transaction" do
+        let(:described_transaction) { Accounting::Transaction.new(project: loan, managed: false, loan_transaction_type_value: :interest) }
```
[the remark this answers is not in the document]
smoyte: Please fix. Thanks!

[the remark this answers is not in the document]
smoyte: Also, `managed` should be true above?

spec/models/accounting/qb/purchase_extractor_spec.rb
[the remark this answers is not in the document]
smoyte: Also not sure why this is in this PR.. are we sure it's meant to be here? If it's just an incidental fix/tidy that's ok, just doublechecking.

app/controllers/admin/accounting/transactions_controller.rb
smoyte: Separately, I'm wondering what the purpose of using the sample method here is, as opposed to just adding `managed: true` to the `merge` in `create`. Thoughts?

smoyte (commented):
Very nice use of the policy! The changes ended up being simple but all the hard work you did makes us feel good that they're correct! Just a few small things in my review!

app/controllers/admin/accounting/transactions_controller.rb
DevneyHamilton: it looks like sample_transaction was there to DRY things up between new and sync. I'm updating it to not use sample_transaction, and have each endpoint method have their own 'sample transaction' so they are each responsible for knowing what their sample txn should be. I agree the merge overwrite is confusing (I looked up the docs to make sure the overwrite is in the right order, but putting that into comments seems clunky).

spec/features/admin/accounting/transaction_flow_spec.rb
DevneyHamilton: i'd rather have this redundancy here than not, since it's been broken, and the relationship between managed and transaction types is undertested throughout and underdocumented in the code, and we'll be changing things.

spec/models/accounting/qb/purchase_extractor_spec.rb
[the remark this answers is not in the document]
DevneyHamilton: yes, these were some tiny cleanups that were fast & safe to do when i was doign that spec review - it seemed just as fast to just do them as write them into the spreadsheet & i didn't want to lose the work

spec/policies/accounting/transaction_policy_spec.rb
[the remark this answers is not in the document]
DevneyHamilton: woops! thanks!

PR #892, opened 2021-07-09
11859: Switch to github actions and chrome headless

smoyte:


DevneyHamilton (approved):
Thank you! I am very excited to be off of phantomjs (RIP) 

What is the download files about? 

I agree with leaving the single quotes & updating the files when we work in them. 

The check annotations are a lot to either deal with all at once or ignore - are we going to handle them right now, or pare them down as we go?

smoyte:
Download files thing is some code I had in Gather that I tossed in but I don't think we need it so I'll take it out for now.

I think we should do a global rubocop --fix run soon to clear up many of these things. I think that is what you were asking about?

PR #894, opened 2021-07-12
11854 unset payment type

DevneyHamilton:
replace qb_object_subtype with disbursement_type, and then set qb field payment_type in txn builder. Convert the other direction in the purchase extractor

### DIFF
--- a/app/models/accounting/qb/purchase_extractor.rb
+++ b/app/models/accounting/qb/purchase_extractor.rb
@@ -34,7 +34,6 @@ module Accounting
       end
 
       def extract_subtype
-        #TODO UPDATE
         payment_type = txn.quickbooks_data["payment_type"]
         txn.disbursement_type = (payment_type.present? && payment_type.to_s == "Check") ? :check : :other
       end
--- a/app/models/accounting/qb/purchase_extractor.rb
+++ b/app/models/accounting/qb/purchase_extractor.rb
@@ -34,8 +34,9 @@ module Accounting
       end
 
       def extract_subtype
+        #TODO UPDATE
         payment_type = txn.quickbooks_data["payment_type"]
-        txn.qb_object_subtype = payment_type.to_s if payment_type.present?
+        txn.disbursement_type = payment_type.to_s if payment_type.present?
       end
 
       def extract_check_number
--- a/app/models/accounting/qb/purchase_extractor.rb
+++ b/app/models/accounting/qb/purchase_extractor.rb
@@ -36,11 +36,11 @@ module Accounting
       def extract_subtype
         #TODO UPDATE
         payment_type = txn.quickbooks_data["payment_type"]
-        txn.disbursement_type = payment_type.to_s if payment_type.present?
+        txn.disbursement_type = (payment_type.present? && payment_type.to_s == "Check") ? :check : :other
       end
 
       def extract_check_number
-        if txn.subtype?("Check")
+        if txn.check?
           doc_number = txn.quickbooks_data["doc_number"]
           txn.check_number = doc_number.remove("MS-Managed").strip if doc_number.present?
         end
--- a/app/models/accounting/qb/purchase_extractor.rb
+++ b/app/models/accounting/qb/purchase_extractor.rb
@@ -35,7 +35,7 @@ module Accounting
 
       def extract_subtype
         payment_type = txn.quickbooks_data["payment_type"]
-        txn.disbursement_type = (payment_type.present? && payment_type.to_s == "Check") ? :check : :other
+        txn.disbursement_type = (payment_type == "Check") ? :check : :other
       end
 
       def extract_check_number
--- a/app/models/accounting/qb/transaction_builder.rb
+++ b/app/models/accounting/qb/transaction_builder.rb
@@ -40,8 +40,7 @@ module Accounting
         p.txn_date = transaction.txn_date if transaction.txn_date.present?
         p.account_ref = transaction.account.try(:reference)
         p.department_ref = department_reference(transaction.project)
-        #TODO: update to convert
-        p.payment_type = transaction.disbursement_type
+        p.payment_type = set_payment_type(transaction)
         p.entity_ref = transaction.vendor.try(:reference) # all check txns have a vendor
         p.total = transaction.amount
 
@@ -197,6 +196,10 @@ module Accounting
         return nil if txn.loan_transaction_type_value == 'other'
         ms_status = txn.loan_transaction_type_value == 'interest' ? 'MS-Automatic' : 'MS-Managed'
         txn.check? ? "#{txn.check_number} #{ms_status}" : ms_status
+      end
+
+      def set_payment_type(disbursement)
+        disbursement.disbursement_type == "check" ? "Check" : "Cash"
       end
     end
   end
--- a/app/models/accounting/qb/transaction_builder.rb
+++ b/app/models/accounting/qb/transaction_builder.rb
@@ -35,12 +35,12 @@ module Accounting
           p.id = transaction.qb_id
           p.sync_token = transaction.sync_token
         end
-        p.doc_number = set_journal_number(transaction)
+        p.doc_number = journal_number(transaction)
         p.private_note = transaction.private_note
         p.txn_date = transaction.txn_date if transaction.txn_date.present?
         p.account_ref = transaction.account.try(:reference)
         p.department_ref = department_reference(transaction.project)
-        p.payment_type = set_payment_type(transaction)
+        p.payment_type = payment_type(transaction)
         p.entity_ref = transaction.vendor.try(:reference) # all check txns have a vendor
         p.total = transaction.amount
 
@@ -83,7 +83,7 @@ module Accounting
           je.id = transaction.qb_id
           je.sync_token = transaction.sync_token
         end
-        je.doc_number = set_journal_number(transaction)
+        je.doc_number = journal_number(transaction)
 
         je.private_note = transaction.private_note
         je.txn_date = transaction.txn_date if transaction.txn_date.present?
@@ -192,14 +192,14 @@ module Accounting
         transaction.update(customer: customer)
       end
 
-      def set_journal_number(txn)
+      def journal_number(txn)
         return nil if txn.loan_transaction_type_value == 'other'
         ms_status = txn.loan_transaction_type_value == 'interest' ? 'MS-Automatic' : 'MS-Managed'
         txn.check? ? "#{txn.check_number} #{ms_status}" : ms_status
       end
 
-      def set_payment_type(disbursement)
-        disbursement.disbursement_type == "check" ? "Check" : "Cash"
+      def payment_type(txn)
+        txn.disbursement_type == "check" ? "Check" : "Cash"
       end
     end
   end
--- a/app/models/accounting/qb/transaction_builder.rb
+++ b/app/models/accounting/qb/transaction_builder.rb
@@ -40,7 +40,8 @@ module Accounting
         p.txn_date = transaction.txn_date if transaction.txn_date.present?
         p.account_ref = transaction.account.try(:reference)
         p.department_ref = department_reference(transaction.project)
-        p.payment_type = transaction.qb_object_subtype
+        #TODO: update to convert
+        p.payment_type = transaction.disbursement_type
         p.entity_ref = transaction.vendor.try(:reference) # all check txns have a vendor
         p.total = transaction.amount
 
@@ -195,7 +196,7 @@ module Accounting
       def set_journal_number(txn)
         return nil if txn.loan_transaction_type_value == 'other'
         ms_status = txn.loan_transaction_type_value == 'interest' ? 'MS-Automatic' : 'MS-Managed'
-        txn.subtype?("Check") ? "#{txn.check_number} #{ms_status}" : ms_status
+        txn.check? ? "#{txn.check_number} #{ms_status}" : ms_status
       end
     end
   end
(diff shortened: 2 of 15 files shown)


app/models/accounting/qb/purchase_extractor.rb
```diff
@@ -35,11 +35,11 @@ def set_type
 
       def extract_subtype
         payment_type = txn.quickbooks_data["payment_type"]
-        txn.qb_object_subtype = payment_type.to_s if payment_type.present?
+        txn.disbursement_type = (payment_type.present? && payment_type.to_s == "Check") ? :check : :other
```
smoyte: I don't think we need the `to_s` here, right? What else could payment_type be besides a string? And then in that case, the present? check is not needed either?

app/models/accounting/qb/purchase_extractor.rb
```diff
@@ -35,11 +35,11 @@ def set_type
...
       end
 
       def extract_check_number
-        if txn.subtype?("Check")
+        if txn.check?
```
smoyte, line 42: This is cleaner!

app/models/accounting/qb/transaction_builder.rb
```diff
@@ -195,7 +195,11 @@ def ensure_accounting_customer_set(transaction)
       def set_journal_number(txn)
         return nil if txn.loan_transaction_type_value == 'other'
         ms_status = txn.loan_transaction_type_value == 'interest' ? 'MS-Automatic' : 'MS-Managed'
-        txn.subtype?("Check") ? "#{txn.check_number} #{ms_status}" : ms_status
+        txn.check? ? "#{txn.check_number} #{ms_status}" : ms_status
+      end
+
+      def set_payment_type(disbursement)
```
smoyte: I wonder if `txn` is a better var name b/c it's consistent with the other setter methods and txn really is a transaction, right? Like there isn't a Disbursement class or anything? Not a big deal but seeing `disbursement.disbursement_type` seems a bit redundant to me as if `disbursement_type` was named improperly, even though it's not.

smoyte: Another quick thing to fix here is to remove the `set_` prefixes from these two methods, since neither of them are really setter methods—they are really pure functions that just compute a value for us. 

Furthermore, if they really were setter methods, as you know, Ruby convention is to use `def whatever=(value)`, attr_writer, etc., rather than `set_...` methods.

 I realize the use of `set_` here is old and you were probably just trying to be consistent but perhaps we can just clean it up?

app/models/accounting/transaction.rb
```diff
@@ -44,12 +44,11 @@ class Accounting::Transaction < ApplicationRecord
   validates :amount, presence: true, unless: :interest?, if: :managed?
   validates :accounting_account_id, presence: true, unless: :interest?, if: :managed?
   validate :respect_closed_books_date, if: :user_created
-  with_options if: ->(txn) { txn.qb_object_subtype == "Check" && txn.user_created } do
-    validates :check_number, presence: true
-  end
-  # validate that all disbursements created in Madeline's transaction form have a vendor
   with_options if: ->(txn) { txn.loan_transaction_type_value == "disbursement" && txn.user_created } do
-    validates :qb_vendor_id, presence: true
+    validate :disbursement_fields
```
smoyte: These `with_options` blocks are I believe intended to be used if you have bunch of validations that should only apply if those conditions are met. It seems odd to use it when there is just one validation in each. 

Could we just do:

```
validate :disbursement_fields, if: ->(txn) { txn.loan_transaction_type_value == "disbursement" && txn.user_created }
```

You could also shorten things by adding an boolean `disbursement?` method, which I think is a good idea since it's best to check against the literal string `"disbursement"` in only one place to avoid errors.

So that would make it

```
validate :disbursement_fields, if: ->(txn) { txn.disbursement? && txn.user_created }
```

Which is nice and neat!

app/models/accounting/transaction.rb
```diff
@@ -44,12 +44,11 @@ class Accounting::Transaction < ApplicationRecord
...
+  end
+  with_options if: ->(txn) { txn.disbursement_type == "check" && txn.user_created } do
+    validates :check_number, presence: true
```
smoyte: Similar idea here but with `check?`

app/models/accounting/transaction.rb
```diff
@@ -189,8 +188,8 @@ def type?(type)
     loan_transaction_type_value == type
   end
 
-  def subtype?(subtype)
-    qb_object_subtype == subtype
+  def check?
+    disbursement_type == "check"
```
smoyte, line 204: Already got this one! Nice!

app/models/accounting/transaction.rb
```diff
@@ -208,6 +207,14 @@ def net_debit_for_account(account_id)
     end
   end
 
+  def disbursement_fields
+    if loan_transaction_type_value == "disbursement"
```
smoyte: Can use `disbursement?` here too!

app/models/accounting/transaction.rb
```diff
@@ -208,6 +207,14 @@ def net_debit_for_account(account_id)
...
+      errors.add(:disbursement_type) unless disbursement_type.present?
+      errors.add(:qb_vendor_id) unless qb_vendor_id.present?
```
smoyte: Not sure what is going on here as there is no error message. Is this meant to be a presence validation, like 'can't be blank'? If so, can we just list those at the top and use `if` to apply them conditionally?

app/models/accounting/transaction.rb
```diff
@@ -208,6 +207,14 @@ def net_debit_for_account(account_id)
...
+      errors.add(:qb_object_type, "must be purchase") unless qb_object_type == "Purchase"
```
smoyte: Typically if a value is forced to be something, we shouldn't really be doing a validation error, but instead:

1. Either hiding or disabling the form field and
2. Forcing the value to be "Purchase" in a before_validation callback. I usually call this a 'normalization' step.

If the client-side JS part of that feels like too much right now, we could just add the normalization bit (cover it with a quick spec that tests both disbursement and non-disbursement flows (I can share some nice idiomatic code for this) and just do it without telling the user. I think that is still better than adding a new validation, translating it, testing it, etc. At least we're part way there..? Curious for your thoughts...

app/views/admin/accounting/transactions/_form.slim
```diff
@@ -26,15 +26,15 @@
   // if not, only display if Purchase because subtype null isn't meaningful to users
   // .disbursement-only class lets js control display when new record
   - if @transaction.new_record? || @transaction.qb_object_type == "Purchase"
```
smoyte, line 28: Can we add a `purchase?` method too and use it here, while we're at it?

app/models/accounting/transaction.rb
smoyte: Oh, hold on, it seems `qb_object_type` is not in the form... Let's talk about this one...

db/migrate/20210709205410_change_qb_object_subtype_to_disbursement_type_on_transaction.rb
```diff
@@ -0,0 +1,15 @@
...
+    change_column_default :accounting_transactions, :disbursement_type, :other
+  end
+
+  def down
+    Accounting::Transaction.where(disbursement_type: :other).update_all(disbursement_type: "Cash")
+    Accounting::Transaction.where(disbursement_type: :check).update_all(disbursement_type: "Check")
+    rename_column :accounting_transactions, :disbursement_type, :qb_object_subtype
+    change_column_default :accounting_transactions, :qb_object_subtype, :nil
+  end
+end
```
smoyte, line 15: Nicely done on the migration!

smoyte: One small thing: the default value should be a string rather than a symbol. And I think on the `down` migration you want `nil` rather than `:nil` or `"nil"`...

smoyte: I am curious why we're adding a default here without adding a null constraint. Does this value apply to all transactions or only those of a certain type? If the latter, I think it's more appropriate to keep allowing nulls and remove the default...

spec/models/accounting/qb/purchase_extractor_spec.rb
```diff
@@ -39,7 +40,24 @@
       expect(txn.account).to eq txn_acct
       expect(txn.amount).to equal_money(12345.67)
       expect(txn.sync_token).to eq "abc"
-      expect(txn.qb_object_subtype).to eq "Check"
+      expect(txn.disbursement_type).to eq "check"
+      expect { txn.save! }.not_to raise_error
```
smoyte, line 44: If we're just trying to test if the object is valid, a more performant and precise way of doing so is `expect(txn).to be_valid`. (More performance b/c it doesn't hit the DB.)

If we are concerned about e.g. a callback raising an error, this works, or we could just do `txn.save!` since an error would fail the spec anyway.

spec/models/accounting/qb/purchase_extractor_spec.rb
```diff
@@ -39,7 +40,24 @@
...
+      Accounting::QB::PurchaseExtractor.new(txn).extract!
+      expect(txn.loan_transaction_type_value).to eq "disbursement"
+      expect(txn.managed).to be false
+      expect(txn.line_items.size).to eq 2
+      expect(txn.line_items[1].account).to eq txn.account
+      expect(txn.line_items[1].credit?).to be true
+      expect(txn.account).to eq txn_acct
+      expect(txn.amount).to equal_money(12345.67)
+      expect(txn.sync_token).to eq "abc"
+      expect(txn.disbursement_type).to eq "other"
```
smoyte, line 60: Thanks for adding spec coverage! Great use of context blocks with the payment_type let var. Textbook!

spec/models/accounting/qb/purchase_extractor_spec.rb
```diff
@@ -28,7 +28,8 @@
     )
   end
 
-  context "extract!" do
+  context "extracting a check" do
```
smoyte, line 31: It might make sense to wrap these next contexts in a `describe "#extract"` block.

spec/models/accounting/qb/transaction_builder_spec.rb
```diff
@@ -1,111 +1,167 @@
...
-    expect(details.map { |i| i.department_ref.value }.uniq).to eq [qb_department_id]
-    expect(details.map { |i| i.account_ref.value }.uniq).to match_array [qb_bank_account_id, qb_principal_account_id, qb_office_account_id]
+  describe "a purchase" do
+    before do
+      transaction.line_items << [prin_line_item, line_item_2]
+    end
+    let(:disbursement_type) { :check }
+    let(:check_number) { "123" }
+    let(:transaction) do
+      Accounting::Transaction.create(
```
smoyte, line 45: Please use `create!` unless you're dealing with a potential validation failure. (IMO create! should have been the default when they designed Rails!)

spec/models/accounting/qb/transaction_builder_spec.rb
```diff
@@ -1,111 +1,167 @@
...
+            amount: 25)
+    }
 
-    it 'creates JournalEntry with date' do
+    let(:line_item_3) {
+      build(:line_item,
+            account: principal_account,
+            posting_type: "Credit",
+            amount: 75)
+    }
```
smoyte, line 142: Can we fit these line items on one line each? If so, it might be a good idea since in my experience it makes it much easier to read the spec if it's not spread over many screens. I will sometimes even shorten var names or use a couple additional `let`s to make things fit on one line.

spec/models/accounting/qb/transaction_builder_spec.rb
```diff
@@ -1,111 +1,167 @@
...
+
+      details = list.map { |i| i.journal_entry_line_detail }
+      expect(details.map { |i| i.posting_type }.uniq).to match_array %w(Debit Credit)
+      expect(details.map { |i| i.entity.entity_ref.value }.uniq).to eq [customer.qb_id]
+      expect(details.map { |i| i.class_ref.value }.uniq).to eq [loan_id]
+      # parent_ref on a qb class is not queryable in qb api, so not tested here; must use qb
+      # to verify that parent class set correctly
+      expect(details.map { |i| i.department_ref.value }.uniq).to eq [qb_department_id]
+      expect(details.map { |i| i.account_ref.value }.uniq).to match_array [qb_bank_account_id, qb_principal_account_id, qb_office_account_id]
+    end
```
smoyte, line 165: Nice added spec coverage!

spec/models/accounting/transaction_spec.rb
```diff
@@ -87,6 +88,37 @@
...
+      it "is not required for non-managed disbursements" do
+        expect do
+          create(:accounting_transaction,
+                 loan_transaction_type_value: :disbursement,
+                 managed: false,
+                 disbursement_type: nil)
+        end.not_to raise_error
+      end
+    end
+  end
```
smoyte, line 120: Not sure if we will need these depending on discussions...

smoyte (commented):
Overall great work! Thanks for doing the refactor to fix naming and improve readability. A few small things and a question about the validation stuff. Let's discuss. Exciting stuff!

app/models/accounting/qb/purchase_extractor.rb
DevneyHamilton: i will keep the presence check. the qb api is stricter than its UI on some things, and we might import something that doesn't have it - if for some reason it's missing we'd want this one to be other.

DevneyHamilton: . . . on second thought, we want a loud error if it's missing, since we'd be writing back the txn as a having a 'cash' payment type if we ever update it . . .

PR #895, opened 2021-07-20
Feature/10344 rails 6

smoyte:
Many of the changes here are new boilerplate generated by Rails via the `rails app:update` script. My general approach is to adjust files touched by that script (e.g. application.rb) to be as close to what the script generates as possible, with the only deviations being things that we actually want to be different from the default. That way it's easy to do a diff between what we have and what it's suggesting. So you'll see a bunch of commented out stuff that we will never need, but I think we should leave it there for this reason.

Thanks!


app/serializers/application_serializer.rb
```diff
@@ -0,0 +1,8 @@
+class ApplicationSerializer < ActiveModel::Serializer
+  # ActiveModel Serializers seems to have a bug in which it tries to determine a root key
+  # even though we are using the :attributes adapter, which doesn't use a root.
+  # This shows up when one tries to `render(json: collection)` where collection is some collection of
+  # objects that has a valid guessable serializer.
+  # Setting type to empty string is a workaround for this.
```
DevneyHamilton, line 6: does it make any sense to add to this comment why setting type to "" is a workaround? or if someone needs to know, we'll find it in the docs?

Gemfile
```diff
@@ -44,19 +43,19 @@ gem "sidekiq"
...
 
 # Tables
 gem "font-awesome-sass", "~> 5.12"
 gem "jquery-ui-rails"
 
-# We are using this fork because:
-# 1. 'tag_options' no longer valid in rails 5
-# 2. AR 'size' method triggers the count query which causes errors in rails 5
-gem "wice_grid", git: "https://github.com/sassafrastech/wice_grid.git", branch: "rails5"
+# We are using this fork because the original gem is no longer maintained. This fork seems to be where
```
DevneyHamilton, line 55: thanks for looking at all the forks & finding the best one for us to use!

app/views/admin/projects/_translatable.html.slim
```diff
@@ -2,7 +2,7 @@
   = f.input :summary, wrapper_html: {class: 'hidden-print'}
     = project.summary
   = f.input :details
-    = simple_format(project.details)
+    = simple_format(project.details.to_s)
```
DevneyHamilton, line 5: a while back i was wondering if there was a way to DRY up the .to_s's on translatables. . and failed. curious if you have any observations about how translatables is working & why? (in terms of software design?)

DevneyHamilton (approved):
Hooray!!! I am so excited we are moving to Rails 6. Thank you, Tom. I saw spec/system/admin/people_flow_spec.rb failed - does that need attention?

app/serializers/application_serializer.rb
smoyte: Yeah I'm hoping future versions will fix this.

Gemfile
smoyte: You're welcome!

app/views/admin/projects/_translatable.html.slim
smoyte: Good question. I'll look into it and get back to you.

smoyte:
> Hooray!!! I am so excited we are moving to Rails 6. Thank you, Tom. I saw spec/system/admin/people_flow_spec.rb failed - does that need attention?

Where do you see that? I'm seeing all checks passed...

PR #896, opened 2021-07-21
11860 sync time stamp seems faulty

DevneyHamilton:
I added logging where last_updated_at is updated, and made it simpler to reason about. 

Main logic changes:
- do not update last_updated_at unless new accounting data is being pulled in (so NOT on authentication updates to teh connection object)
- when updating last_updated_at, always use timestamp from just before the update process (instead of after the update process finishes)

### DIFF
--- a/app/jobs/quickbooks_update_job.rb
+++ b/app/jobs/quickbooks_update_job.rb
@@ -6,7 +6,7 @@ class QuickbooksUpdateJob < QuickbooksJob
     # issues so that if fetch we still hide those loans' txn data appropriately.
     Accounting::SyncIssue.global.delete_all
     self.updater = Accounting::QB::Updater.new
-    started_update_at = Time.zone.current
+    started_update_at = Time.current
     updater.qb_sync_for_loan_update
     task.set_activity_message("syncing_with_quickbooks")
     loans.each_with_index do |loan, index|
--- a/app/jobs/quickbooks_update_job.rb
+++ b/app/jobs/quickbooks_update_job.rb
@@ -37,6 +37,7 @@ class QuickbooksUpdateJob < QuickbooksJob
     # TODO: This is duplicated in Updater#update and needs to be DRYed up.
     # We record last_updated_at as the time this update started. The user-prompted ways
     # the update is started are used by only admins and rarely.
+    Rails.logger.debug("Setting qb cnxn last_updated_at to #{started_update_at}"
     updater.qb_connection.update_attribute(:last_updated_at, started_update_at)
 
     # Even if there have been per-loan errors, we still consider the task completed.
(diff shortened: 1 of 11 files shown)


app/jobs/quickbooks_update_job.rb
```diff
@@ -34,14 +35,10 @@ def perform(_job_params)
...
-    # We record last_updated_at as the time the update *finishes* because last_updated_at
-    # is used to avoid runs that immediately follow each other as in the case of a transaction creation
-    # followed by a transaction listing. If we record the time the update *started* and the update
-    # takes some time, we would need to increase the MIN_TIME_BETWEEN_UPDATES value and that might
-    # make it frustrating for users who want to deliberately re-run the updater.
-    # The other function of last_updated_at is to check if a full sync needs to be run,
-    # but that condition is measured in days, not seconds, so this small a difference shouldn't matter.
-    updater.qb_connection.update_attribute(:last_updated_at, Time.current)
+    # We record last_updated_at as the time this update started. The user-prompted ways
+    # the update is started are used by only admins and rarely.
```
smoyte: I think I follow your logic here! But this comment is written seemingly as a response to the old one, and without the old one as context, the second sentence is kind of hard to grasp. So perhaps some more detail here would be useful, perhaps just saying that it's important to use update start time so as not to miss txns. Simple as that really, isn't it?

Separately: the old comment mentions `MIN_TIME_BETWEEN_UPDATES`. Do we need to increase that? Or perhaps more correctly, do we need some way of preventing multiple updates running at one time and causing havoc? That sounds bad and I'm not convinced we are dealing with it properly right now, especially if we are changing the below... 

Like what if someone clicks the sync button before the previous sync finishes?

app/models/accounting/qb/updater.rb
```diff
@@ -114,10 +112,12 @@ def too_soon_to_run_again?
       end
 
       def changes
+        # assumes that after a new qb connection has been established, the FullFetcher will
+        # retrieve all data from qb and last_updated_at will have a value before updater runs.
         unless last_updated_at && last_updated_at > max_updated_at
           raise DataResetRequiredError
         end
-
+        Rails.logger.debug("Calling to QB API to get changes since #{last_updated_at}")
```
smoyte, line 116: Nice

app/models/accounting/qb/full_fetcher.rb
```diff
@@ -37,6 +37,7 @@ def fetch_qb_data
         ::Accounting::QB::TransactionFetcher.new(division).fetch
         ::Accounting::QB::DepartmentFetcher.new(division).fetch
         ::Accounting::QB::VendorFetcher.new(division).fetch
+        Rails.logger.debug("Setting qb cnxn last_updated_at to #{started_fetch_at}")
```
smoyte: Should we make a method on the Connection class for updating the stamp and move this log statement in there? That way it's not duplicated and we don't accidentally leave it out if we add one of these somewhere else?

smoyte (commented):
Looks like a big improvement! I'm a bit concerned about concurrent updates happening unintentionally. Might it make sense to also add a global "accounting update in progress" flag somewhere for now that would flat out prevent that? And if we can support it later in some cases (e.g. two different loans), we can consider it then?

PR #897, opened 2021-07-23
Feature/11872 division parent restriction

smoyte:
This is one early chunk of the manage questions work. Includes some cleanups.


spec/support/matchers/have_errors.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literals: true
+
```
DevneyHamilton, line 2: thanks for adding this matcher

spec/support/shared_examples/flow.rb
```diff
@@ -0,0 +1,42 @@
+# TS, 7/23/21: I think we should deprecate this file and move the testing logic into each
+# feature spec, since each page is slightly different and there are different things we want to test
+# in each one. I also don't think we need to be testing the change division flow for every module.
```
DevneyHamilton, line 3: I was wondering about this and this also sounds good to me as a direction to go in.

DevneyHamilton (approved):
Thanks! I am going to merge this.

PR #901, opened 2021-08-12
11748 public pages v2

DevneyHamilton:
Removes wp template code. 
Redirects existing urls to new division public pages
Create a landing page for each division with a list of loans for the division

- whether to add separate color config for public pages DONE
- revisit naming & ordering of new fields on admin division page (external website url (e.g. theworkingworld.org/us or seedcommons.org) and public url on madeline (.e.g madeline.theworkingworld.org/divisions/seed-commons))DONE
- work w tww on making sure their logos, descrpitions, and colors are ready for changeover IN PROGRESS


app/assets/stylesheets/admin/loans/_questionnaires.scss
```diff
@@ -99,6 +99,7 @@ section.questionnaires {
   }
 
   .help-block {
+    color: var(--custom-accent-darkened);
```
smoyte, line 102: Neat! This is the first I've seen of custom CSS vars. I see why you're using them too. Good idea!

app/assets/stylesheets/public/_divisions.scss
```diff
@@ -0,0 +1,274 @@
+// TODO: update with safer custom colors
+// Currently assumes custom-accent-main is a dark color,
+// and uses --custom-accent-fg-text as an second accent color.
+
```
smoyte, line 4: I assume most of this CSS came from some other files?

app/assets/stylesheets/public/_divisions.scss
```diff
@@ -0,0 +1,274 @@
...
+    float: left;
+    padding: 0 3% 3% 0;
+  }
+
+  .loan-summary {
+    display: inline-block;
+    float: left;
+  }
+}
+/** LOAN LIST STYLES **/
```
smoyte, line 83: Generally I find if a big comment divider like this is needed in a SCSS file, it is probably better to just split the file into several smaller ones in an aptly named folder.

app/assets/stylesheets/public/_divisions.scss
smoyte: Separate comment for below since I can't comment on an empty file. What is ` app/assets/stylesheets/shared/_division-brand.scss ` for?

app/controllers/application_controller.rb
```diff
@@ -16,17 +16,7 @@ def admin_controller?
...
-    path_action = path[:action]
-
-    # Public loan pages have a different error message
-    if path_controller == "public/loans" && path_action == "show"
-      flash[:error] = t('loan.public.not_authorized')
-    else
-      flash[:error] = t('unauthorized_error')
-    end
-
+    flash[:error] = t('unauthorized_error')
```
smoyte: Yay!

spec/system/public/division_flow_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      context "loan has no image" do
+        let!(:loan_no_image) { create(:loan, :active, :public, division: parent_division) }
+        it "displays seedcommons logo" do
+          visit public_division_path(short_name: parent_division.short_name)
+          expect(page.find('div.coop_pic_sm')['style']).to include("seedcommons")
+        end
+      end
+    end
+  end
+end
```
smoyte, line 105: Might be good to add spec coverage of the custom 403 handling?

app/controllers/application_controller.rb
smoyte: I'm just realizing the error code being thrown below is wrong I think. It should be 403 if it's pundit related. 401 is if they're not logged in at all.

it's confusing b/c 401 is called "unauthorized" but it really should be called "unauthenticated", while 403 is called "forbidden" but really should be "unauthorized".

app/views/public/divisions/show.html.slim
```diff
@@ -0,0 +1,68 @@
...
+                li.disabled #{t :show}:
+                li class=('active' if !@params[:status] || @params[:status] == 'active')
+                  = link_to t(:show_active), @params.merge(status: nil, pg: nil)
+                li class=('active' if @params[:status] == 'completed')
+                  = link_to t(:show_completed), @params.merge(status: 'completed', pg: nil)
+                li class=('active' if @params[:status] == 'all')
+                  = link_to t(:show_all), @params.merge(status: 'all', pg: nil)
+            .country
+              = label_tag(:division, "#{t :division}: ")
+              = select_tag(:division, options_for_select(division_select_options(include_all: true, public_only: true), @params[:division]),
```
smoyte, line 35: It appears `@params[:division]` never gets set?

app/controllers/public/divisions_controller.rb
```diff
@@ -0,0 +1,30 @@
+class Public::DivisionsController < Public::PublicController
+  layout "public/main"
+
+  rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
+
+  def show
+    @params = {status: params[:status], pg: params[:pg]}
+    @division = Division.find_by(short_name: params[:short_name])
+    @status_filter = [:active]
```
smoyte, line 9: It seems `@status_filter` is only used in this method so perhaps doesn't need to be an ivar?

app/controllers/public/divisions_controller.rb
```diff
@@ -0,0 +1,30 @@
+class Public::DivisionsController < Public::PublicController
+  layout "public/main"
+
+  rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
+
+  def show
+    @params = {status: params[:status], pg: params[:pg]}
```
smoyte, line 7: I think calling this `@params` may be a bit confusing b/c of the similarity to the `params` hash which has a special meaning in Rails as you know. Also you could use the `slice` method here.

Furthermore, does `params[:pg]` ever get used for anything? If not, perhaps you could simplify this all quite a bit?

app/controllers/public/divisions_controller.rb
smoyte: Also I think this would read better if the above assignment is made in an `else` block below instead of before the `if`.

app/controllers/public/divisions_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    @status_filter = [:active]
+    if @params[:status] == "all"
+      @status_filter = [:active, :completed]
+    elsif @params[:status]
+      @status_filter = [@params[:status]]
+    end
+    @selected_division = @division
+    authorize @division
+    # loan scope handles filtering loans based on whether their division is public,
+    # only completed or active loans, and their public level value
```
smoyte, line 18: Thanks for this comment. I'm not sure if you want to get into this but I'll offer it. I think the responsibility for determining which divisions' loans to include does lie with the Policy class. I think the way to implement it might look something like:

```
LoanPolicy::Scope.new(nil, Loan).publicly_visible_for_division(@division).where(status_value: ...)...
```

You'd get rid of the implicit call to the `publicly_visible` method and call this new one directly, which would contain the old logic plus the restriction to self and descendants.

The trick here is knowing that you can define custom methods on the Scope object and just bypass the `policy_scope` shortcut, which only knows how to call `resolve` with no arguments.

Again, I think it would be fine to leave it here if you really want to move on but if you have space I'd suggest refactoring and adjusting the LoanPolicy spec.

app/controllers/public/divisions_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    divisions_to_include = @division.self_and_descendants&.map(&:id)
```
smoyte, line 19: For something like this it's better to use `pluck(:id)` since `map` will load up all the Division objects into memory before getting the `id` attrib.

app/controllers/public/loans_controller.rb
```diff
@@ -1,36 +1,25 @@
...
+    redirect_to controller: "divisions", action: "show", short_name: division
   end
 
-  def show
-    @loan = Loan.find(params[:id])
-    authorize @loan
-
-    @pictures = @loan.featured_pictures(limit: 5) # for slideshow
-    @other_loans = policy_scope(Loan.related_loans(@loan).active_or_completed.order('signing_date desc'))
+  def division_for_site(site)
```
smoyte, line 7: Can this be a private method?

app/controllers/public/loans_controller.rb
```diff
@@ -1,36 +1,25 @@
...
-    @loans = policy_scope(Loan.filter_by_params(params).
-          includes(:organization, division: :parent).
-          page(params[:pg]).per(20).
-          order('signing_date DESC'))
-    authorize @loans
-
-    # Set last loan list URL for 'Back to Loan List' link
-    session[:loans_path] = request.fullpath
+    division = params[:division] || division_for_site(params[:site])
+    redirect_to controller: "divisions", action: "show", short_name: division
```
smoyte, line 4: Hooray for nice redirects!

Do we have spec coverage on these? It would be super easy to add if not (using request specs perhaps) and it would pay dividends I think. Redirects like this can be finicky...

smoyte: Also, I'm surprised these work currently given the `public` namespace is not specified in some way. Perhaps it figures it out b/c the current controller is also in `public`?

app/models/division.rb
```diff
@@ -124,9 +125,14 @@ def parent_division_and_name
   end
 
   def generate_short_name
-    return if short_name.present? && Division.pluck(:short_name).exclude?(self.short_name)
+    # no change to short_name that is already saved and therefore uniq
+    return if self.short_name.present? && self.short_name == self.attribute_in_database(:short_name)
```
smoyte, line 129: Could you use `persisted? && !short_name_changed?` Anything around attrib changes should probably used the AR Dirty features.

app/models/division.rb
```diff
@@ -124,9 +125,14 @@ def parent_division_and_name
...
+
+    # accept a short_name that is being provided and is uniq
+    return if self.short_name.present? && Division.pluck(:short_name).exclude?(self.short_name)
```
smoyte, line 132: `self.` not needed here.

smoyte: Separately, a more efficient way to check for uniqueness is `Division.where(short_name: short_name).empty?`, which Rails will turn into an EXISTS query and thus you won't need to load all the existing short names into memory and check them with ruby `exclude?`

app/models/division.rb
```diff
@@ -124,9 +125,14 @@ def parent_division_and_name
...
 
-    self.short_name = name.parameterize
+    # short_name not provided or provided short_name is not uniq
+    self.short_name ||= name.parameterize
```
smoyte, line 135: If we are adding a short name field to the form, might it make sense to actually remove all this and just use Rails' uniqueness validation and have the user pick a unique one, rather than generating a shortname with a nasty UUID on the end?

app/policies/division_policy.rb
```diff
@@ -5,7 +5,7 @@ def index?
   end
 
   def show?
-    division_member_or_admin(division: @record)
+    user ? division_member_or_admin(division: @record) : @record.public
```
smoyte, line 8: Would `public?` work?

app/policies/division_policy.rb
```diff
@@ -33,26 +33,27 @@ def select?
...
+      scope.published
     end
 
     def accessible_divisions(public_only: false)
-      public_only ? Division.published : user.accessible_divisions
+      if public_only
+        resolve_public
+      else
+        resolve_admin
+      end
```
smoyte, line 56: Nice refactors here. I think you could perhaps go a bit further. The similarity of this method to `resolve` got me thinking. I think `public_only` is only ever true in the public loans case. The rest of the usages of this method via `division_select_options` are in the admin module. So I think really this method is just equivalent to `resolve` and it could be removed. Feel free to break my logic!

app/views/admin/divisions/_form.html.slim
```diff
@@ -72,6 +78,19 @@
     = documentation_popover(@documentations_by_html,
       html_identifier: "#{form_identifier}_public", options: popover_options)
 
+  - if @division.public?
```
smoyte, line 75: Putting this in an ERB if block means that when creating a new division you won't be able to set the URL. You'll have to know to save the record and then edit it. Is that ok?

app/views/layouts/public/main.html.slim
```diff
@@ -0,0 +1,32 @@
...
+    = javascript_include_tag "application"
+    = yield :per_page_js
+    = render("layouts/shared/custom_colors") 
+
+    = stylesheet_link_tag "application", media: "all"
+
+
+  body class="#{params[:controller].gsub('/', '-')}
+              action-#{params[:action]}
+              #{devise_controller? ? 'devise' : ''}
```
smoyte, line 24: I don't think this can ever be true for public pages?

app/views/layouts/public/main.html.slim
```diff
@@ -0,0 +1,32 @@
+/ This layout doesn't get used directly by the controller, but instead via
+/ the 'nested layout' design pattern from the Rails docs.
+/ It is called by the devise and signed_in layouts, and then yeilds to custom content blocks
+/ defined by those layouts.
```
smoyte, line 4: Nice job figuring out how to use this pattern!

app/views/public/divisions/show.html.slim
```diff
@@ -0,0 +1,68 @@
...
+                    = loan.location
+                  .sign_date
+                    i.fa.fa-fw.fa-calendar>
+                    = ldate(loan.signing_date, format: :long)
+                  .loan_amount
+                    / i.icon-coinstack>
+                    = format_currency(loan.amount, loan.currency)
+                  / p.progress.progress-danger style="background:#444;"
+                  /   span.progress-bar data-percentage="75" style=("width: 75%;")
+                  / p: button.lend_btn.btn type="button"  Add Loan
```
smoyte: Can we remove commented stuff for now?

config/environments/development.rb
```diff
@@ -76,18 +76,4 @@
...
-      us: 'https://theworkingworld.org',
-      argentina: 'https://labase.org',
-      nicaragua: 'https://labase.org',
-    },
-    template_paths: { # url of blank wordpress page for rails to retrieve
-      us: '/us/rails',
-      argentina: '/argentina/rails',
-      nicaragua: '/nicaragua/rails',
-    },
-  }
```
smoyte, line 92: Thank gosh this is gone!

spec/policies/loan_policy_spec.rb
```diff
@@ -5,30 +5,38 @@
...
+  describe 'without user, as in public usage' do
     let!(:user) { nil }
-    before { user.try(:accessible_division_ids) }
 
     it 'returns the correct loans' do
-      expect(loan_scope(user).resolve).to contain_exactly(loan1, loan2)
+      expect(loan_scope(user).resolve).to contain_exactly(
+        public_div_featured_active_loan,
+        public_div_featured_completed_loan
+      )
```
smoyte, line 39: Nice coverage here.

spec/system/admin/division_flow_spec.rb
```diff
@@ -40,28 +40,60 @@
...
+    expect(page).to have_content(division.name)
+
+
+    # confirm shortname update took effect in public url
+    visit admin_division_path(division)
+    public_url_section = find(".division_public_url a", match: :first)
+      expect(public_url_section).to have_content("newshort")
+    find(".division_public_url a", match: :first).click
+
+    expect(URI.parse(current_url)).to have_content("newshort")
```
smoyte, line 82: Good to cover this too!

spec/system/public/division_flow_spec.rb
```diff
@@ -0,0 +1,105 @@
+require 'rails_helper'
+
+feature 'public division flow' do
```
smoyte, line 3: Very nice coverage here! Great work!

smoyte (commented):
Monumental! Thank you so much for this! Looks great overall. Please take a look at my comments and let me know if you'd like to discuss.

app/assets/stylesheets/admin/loans/_questionnaires.scss
DevneyHamilton: yeah i like them for this :)

app/assets/stylesheets/public/_divisions.scss
DevneyHamilton: correct assumption, and re: _division-brand.scss, thanks for catchng that, i think it's danglign from the merge but i'll confirm that. . .

app/controllers/application_controller.rb
DevneyHamilton: hmm it's coming from Pundit::NotAuthorizedError, and then I can just use status: :unauthorized to keep it consistent. . .

app/views/public/divisions/show.html.slim
DevneyHamilton: yes!

app/controllers/application_controller.rb
smoyte: Yes, it is indeed confusing. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses. I do believe 403/`:forbidden` is still the correct response.

app/assets/stylesheets/public/_divisions.scss
DevneyHamilton: agreed, this is pulled in from the past & i'm going to leave as is for now, but if we do more with styles here i'll clean up the whole thing more

spec/system/public/division_flow_spec.rb
DevneyHamilton: yes . . . will do this as a patch next wk before things go to prod

DevneyHamilton:
thanks for the review @smoyth - very helpful as usual :)

PR #909, opened 2021-08-31
Feature/11891 append only

smoyte:
This is phase 2/3 of manage question work.

Enforces proper ordering of questions based on division in UI and in model. Migrates existing data to achieve same. Adds good feature spec coverage of manage questions page. Improves performance of manage questions page. Tidies a bunch of things.

QA wise, I have tested it pretty thoroughly. You may want to try it out though. It's pretty fun to see where you can and can't drag or add questions!


app/assets/javascripts/admin/questions_view.coffee
```diff
@@ -33,13 +37,36 @@ class MS.Views.QuestionsView extends Backbone.View
...
+        when 'inside' then 0
+        when 'before' then targetIndex
+        else targetIndex + 1
+
+    # Ensure left sibling is in same or ancestor division, right sibling in same or descendant division
+    (insertIndex == 0 || siblings[insertIndex - 1].division_depth <= ownDivisionDepth) &&
+      (insertIndex >= siblings.length || siblings[insertIndex].division_depth >= ownDivisionDepth)
+
   newNode: (e) ->
+    e.preventDefault()
```
smoyte, line 63: Page will no longer scroll to top when editing/creating question!

app/controllers/admin/questions_controller.rb
```diff
@@ -4,16 +4,16 @@ class Admin::QuestionsController < Admin::AdminController
 
   def index
     authorize Question
-    # Hide retired questions for now
-    sets = QuestionSet.where(internal_name: %w(loan_criteria loan_post_analysis)).to_a
-    questions = sets.map { |s| top_level_questions(s) }.flatten
-    @json = ActiveModel::Serializer::CollectionSerializer.new(questions, user: current_user).to_json
+    @set_name = params[:set] || "criteria"
```
smoyte, line 7: Each question set is now a separate page load instead of loading them all and just hiding the ones not in the current set (!). Much faster.

db/migrate/20210730190522_remove_required_from_questions.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveRequiredFromQuestions < ActiveRecord::Migration[6.1]
+  def change
+    remove_column :questions, :required, :boolean
```
smoyte, line 3: This became obsolete when we introduced per-loan requiredness

db/migrate/20210819162038_fix_division_depths_on_questions.rb
```diff
@@ -0,0 +1,55 @@
...
+    # Ensure no groups with children in a higher division
+    groups_with_higher_children_query = <<-SQL
+      SELECT DISTINCT q.id, q.division_id FROM questions q WHERE EXISTS (
+        SELECT id FROM questions cq WHERE cq.parent_id = q.id AND
+          (SELECT MAX(generations) FROM division_hierarchies WHERE descendant_id = cq.division_id) <
+            (SELECT MAX(generations) FROM division_hierarchies WHERE descendant_id = q.division_id)
+      )
+    SQL
+    unless execute(groups_with_higher_children_query).to_a.empty?
+      raise "Group(s) exist with children in higher division. Please correct manually before proceeding."
```
smoyte, line 12: I got rid of these manually on production and staging. This is just to check none reappear between now and deploy.

app/models/question_set.rb
```diff
@@ -7,30 +7,6 @@ class QuestionSet < ApplicationRecord
...
-      question_set_id: id,
-      parent: nil,
-      data_type: "group",
-      internal_name: "root_#{id}",
-      required: true,
-      position: 0,
-      division: Division.root
-    )
-  end
-
```
DevneyHamilton, line 33: this is very nice to have going away!

db/migrate/20210730161610_change_question_status_to_boolean.rb
```diff
@@ -0,0 +1,7 @@
+class ChangeQuestionStatusToBoolean < ActiveRecord::Migration[6.1]
```
DevneyHamilton, line 1: this is also a great change to have - thank you!

db/migrate/20210730190522_remove_required_from_questions.rb
DevneyHamilton: thanks!

app/serializers/filtered_question_serializer.rb
```diff
@@ -31,7 +31,10 @@ def required_loan_types
   end
 
   def can_edit
-    # This is not a policy-level matter, it's a view thing.
+    # This is not a policy-level matter, it's a view thing. Only questions for the currently selected division
+    # can be edited as a matter of good UX. A user may be allowed to edit more questions than that
+    # by the policy, so this rule is more restrictive than the policy. The policy is also checked
+    # at page load. So we don't check it here for efficiency's sake.
```
DevneyHamilton, line 37: thanks for this comment - useful to have for the future.

app/models/question_mover.rb
```diff
@@ -0,0 +1,79 @@
...
+    ensure_current_division
+    ensure_new_parent_is_group
+    ensure_correct_adjacency
+    ensure_parent_of_same_or_ancestor_division
+    method =
+      case relation
+      when :before then :prepend_sibling
+      when :after then :append_sibling
+      when :inside then :prepend_child
+      end
```
DevneyHamilton, line 20: QUESTION: does prepending for inside here mean that as a user, if i have group A with qs A1, A2, and A4, and I want to pull in an existing q A3 from elsewhere, A3 will first be added to A, and they'll be A3, A1, A2, A4. Then I can move A3 after A2 in a second step? that's ok, just seeing if i understand this right.

spec/models/question_mover_spec.rb
```diff
@@ -0,0 +1,316 @@
...
+describe QuestionMover do
+  # Used by create_question function
+  let!(:qset) { create(:question_set, internal_name: "loan_criteria") }
+
+  let!(:div_a) { create(:division, parent: root_division) }
+  let!(:div_b) { create(:division, parent: div_a) }
+  let!(:div_c1) { create(:division, parent: div_b) }
+  let!(:div_c2) { create(:division, parent: div_b) }
+  let!(:div_c3) { create(:division, parent: div_b) }
+  let!(:div_d) { create(:division, parent: div_c1) }
```
DevneyHamilton, line 12: thanks for thorough heirarchy testing here

spec/models/question_mover_spec.rb
```diff
@@ -0,0 +1,316 @@
...
+  let!(:q4) { create_question(division: div_b, parent: qset.root_group, type: "group") }
+    let!(:q40) { create_question(division: div_b, parent: q4, type: "text") }
+    let!(:q41) { create_question(division: div_c1, parent: q4, type: "text") }
+  let!(:q5) { create_question(division: div_c1, parent: qset.root_group, type: "text") }
+  let!(:q6) { create_question(division: div_c1, parent: qset.root_group, type: "group") }
+    let!(:q60) { create_question(division: div_c1, parent: q6, type: "text") }
+    let!(:q61) { create_question(division: div_c1, parent: q6, type: "text") }
+    let!(:q62) { create_question(division: div_c3, parent: q6, type: "text") }
+  let!(:q7) { create_question(division: div_c1, parent: qset.root_group, type: "group") }
+  # rubocop:enable Layout/IndentationConsistency
```
DevneyHamilton, line 50: cool that you can do this w rubocop

spec/models/question_mover_spec.rb
```diff
@@ -0,0 +1,316 @@
...
+            let(:relation) { :after }
+            it_behaves_like "succeeds"
+          end
+
+          context "within existing division block :after target with different division" do
+            let(:current_division) { div_c1 }
+            let(:question) { q41 }
+            let(:target) { q112 } # has div_c3
+            let(:relation) { :after }
+            it_behaves_like "succeeds"
```
DevneyHamilton, line 148: QUESTION: Am I following right that it's ok to have a q for div B be after qs for div C as long as Div B and C are siblings? (re error message: "must be adjacent to questions of same division depth")

spec/models/question_mover_spec.rb
```diff
@@ -0,0 +1,316 @@
...
+            let(:target) { q7 }
+            let(:relation) { :inside }
+            it_behaves_like "succeeds"
+          end
+        end
+      end
+    end
+  end
+
+  context "moving question with ancestor division than current division" do
```
DevneyHamilton: QUESTION:  is there a word missing here? with ancestor div different than curr div?

app/controllers/admin/questions_controller.rb
DevneyHamilton: awesome, thank you!

app/controllers/admin/questions_controller.rb
```diff
@@ -46,15 +47,10 @@ def update
...
-      when 'inside' then :prepend_child
-    end
-
-    target.send(method, @question)
+    QuestionMover.new(current_division: selected_division, question: @question,
+                      target: target, relation: params[:relation].to_sym).move
     render_set_json(@question.question_set)
-  rescue
+  rescue ArgumentError
     flash.now[:error] = I18n.t('questions.move_error') + ": " + $!.to_s
```
DevneyHamilton, line 54: QUESTION: we aren't expecting users to see this error right, it's just a fallback if something in the jqtree restrictions on moving in the browser fails?

app/models/question.rb
```diff
@@ -198,4 +199,14 @@ def ensure_internal_name
       self.update! internal_name: "field_#{id}"
     end
   end
+
+  def parent_division_depth
```
DevneyHamilton: thanks for this!

app/policies/question_policy.rb
```diff
@@ -1,11 +1,23 @@
 class QuestionPolicy < ApplicationPolicy
+  def show?
```
DevneyHamilton, line 2: great

spec/policies/question_policy_spec.rb
```diff
@@ -1,49 +1,35 @@
-require 'rails_helper'
+require "rails_helper"
```
DevneyHamilton, line 2: thanks for making this spec about user role instead of question arrangements.

app/serializers/question_serializer.rb
```diff
@@ -32,10 +33,6 @@ def required_loan_types
   end
 
   def can_edit
-    if @user
-      Pundit.policy!(@user, object).edit?
-    else
-      return false
-    end
+    object.division_id == selected_division.id
```
DevneyHamilton: we should communicate to brendan extra that ppl can only edit qs in the curr division so he can communicate that to tww people really clearly.

app/models/division.rb
```diff
@@ -74,6 +74,18 @@ def division
     self
   end
 
+  # Allows efficient multiple tree comparisons by caching.
+  def self_or_ancestor_of?(other_division)
```
DevneyHamilton, line 78: cool, thank you!

app/controllers/admin/questions_controller.rb
```diff
@@ -73,7 +73,7 @@ def top_level_questions(set)
     root_group = set.root_group_preloaded
     return root_group.children unless selected_division
 
-    FilteredQuestion.new(root_group, division: selected_division).children
+    FilteredQuestion.new(root_group, selected_division: selected_division).children
```
DevneyHamilton: awesome, this will be really helpful in the future

app/assets/javascripts/admin/questions_view.coffee
```diff
@@ -6,7 +6,7 @@ class MS.Views.QuestionsView extends Backbone.View
     new MS.Views.AutoLoadingIndicatorView()
     @tree = @$('.jqtree')
     @tree.tree
-      data: @tree.data('data')
+      data: params.questions
```
DevneyHamilton, line 12: thanks!

app/assets/javascripts/admin/questions_view.coffee
DevneyHamilton: thanks!

spec/models/question_spec.rb
```diff
@@ -16,13 +16,64 @@
...
       expect(f2.position).to eq 1
       expect(f3.position).to eq 2
       expect(f31.position).to eq 0
       expect(f32.position).to eq 1
     end
+
+    context "with manually set position" do
+      let!(:f33) { create_question(set: set, parent: f3, name: "f33", type: "string", position: 0) }
+
+      it "goes to end anyway (position NOT respected)" do
```
DevneyHamilton: SUGGESTED CHANGE: "goes to end anyway (manually set position is overridden to follow ordering rules)" would make more sense to me here. . .

app/assets/javascripts/admin/questions_view.coffee
```diff
@@ -36,6 +37,29 @@ class MS.Views.QuestionsView extends Backbone.View
     'change [name="question[override_associations]"]': 'showHideAssociations'
     'change .require-checkbox': 'changeRequireCheckbox'
 
+  canMove: (node) ->
+    node.can_edit
+
+  canMoveTo: (movedNode, targetNode, position) ->
```
DevneyHamilton, line 43: woohooo!

DevneyHamilton:
Thank you - this is a boatload of work and puts question code in a muuuuch better place!

One thing that needs fixing here is that as a user who is currently in the root division, I need to be able to click the "Questions" link without raising an error (currently get error that division can't be nil). This may be waiting on UI work for what that scenario should look like. If not, we could load the questions page with no questions and instruct user to select a division to see questions. 

Another question and possible todo: is there anywhere enforcing that every question set can have only one root question (parent_id: nil)?

A general question: is override_associations continuing with us, or is that going away as part of this whole refactor? 

There's one [suggested change](https://github.com/sassafrastech/madeline/pull/909#discussion_r701005297) to a spec desc above - the rest is gratitude and clarifying questions!

app/models/question_mover.rb
smoyte: Good question! This comes from the jqtree library. It will let you drag A3 into position directly and send it to the server as `:after`. I think `:inside` is only ever used if the group is empty.

spec/models/question_mover_spec.rb
smoyte: Yes exactly. We just enforce that each division is a contiguous block in that case, so the numbering doesn't get _too_ weird.

spec/models/question_mover_spec.rb
smoyte: Ah yes, I'll fix.

app/controllers/admin/questions_controller.rb
smoyte: It does actually show up on the page due to some pre-existing error display machinery. But the user likely won't see it b/c it shows at top of the page and user is almost always scrolled down. But it could be helpful for helping us diagnose. You are right that it would be unexpected for that error to show up and would mean we have a bug.

spec/models/question_spec.rb
smoyte: Hmm, this is what I meant: `goes to end anyway (closure_tree does not appear to respect explicit position, not sure why)`

smoyte:
>  I need to be able to click the "Questions" link without raising an error

Yes, this is in phase 3, and is why we can't merge phase 2 yet. But now as I write this, I am thinking it would be better to fix that and merge rather than keeping all this code diverged. I just pushed up some changes for that.

> Another question and possible todo: is there anywhere enforcing that every question set can have only one root question (parent_id: nil)?

Great idea! I love the way you are thinking! I can add a special unique index to ensure only one NULL parent_id per question_set_id, something like this: https://www.enterprisedb.com/postgres-tutorials/postgresql-unique-constraint-null-allowing-only-one-null (Note that a regular unique index won't work b/c NULL is not treated like other values).

> A general question: is override_associations continuing with us, or is that going away as part of this whole refactor?

No plans to change any of that required question machinery as of now.

app/controllers/concerns/questionnaire_renderable.rb
```diff
@@ -5,7 +5,7 @@ def prep_questionnaire(json: true)
     @attrib = params[:filter] || "criteria"
     @response_set ||= @loan.send(@attrib) || ResponseSet.new(kind: @attrib, loan: @loan)
     @response_set.current_user = current_user
```
DevneyHamilton: do we still need current user on response_set?

smoyte: Gosh I don't think so! Great catch! I'll remove it.

PR #911, opened 2021-09-01
Remove confusing current_division method

smoyte:
`current_division` is a very confusing method since it returns the user's "owning division" if selected_division is nil, which means the Root Division/'All Divisions' is selected.

In the prod db, there are 3 users who have access to the root division but do not have it as their 'owning division'. So for those three users, if they have 'All Divisions' selected, some strange things would happen.

I think devs generally think `current_division` means you'll get a division object back, including root if you've selected root. So I saw it used in a bunch of places where I think dev intention was not to get the user's owning division when in root div mode. 

So I think it's best to remove the method entirely and replace with selected_division_or_root. I went through and tried to guess developer intent as best I could.

Ultimately I'd like to remove selected_division altogether and rename selected_division_or_root to current_division, as I think it's an anti-pattern to return `nil` in that way. It's much better to use the root division object itself as it leads to fewer nil errors. But for now, given that selected_division still exists, using current_division is always going to be confusing b/c it's not clear what the difference is between it and selected_division.

### DIFF
--- a/app/controllers/admin/divisions_controller.rb
+++ b/app/controllers/admin/divisions_controller.rb
@@ -5,8 +5,8 @@ class Admin::DivisionsController < Admin::AdminController
     redisplay_url = params[:redisplay_url] || root_path
     division_id = params[:division_id]
     save_selected_division_to_session(division_id)
-    division = Division.find_safe(division_id)
-    authorize division || current_division
+    division = Division.find_by(id: division_id) || Division.root
+    authorize division
     redirect_to redisplay_url
   end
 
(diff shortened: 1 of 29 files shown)


app/controllers/admin/divisions_controller.rb
```diff
@@ -5,8 +5,8 @@ def select
     redisplay_url = params[:redisplay_url] || root_path
     division_id = params[:division_id]
     save_selected_division_to_session(division_id)
-    division = Division.find_safe(division_id)
-    authorize division || current_division
```
smoyte, line 9: This one was especially bizarre—"allow selecting this division if user can select? this division OR they've selected 'All Divisions' and they can select? their 'owning division'".

PR #912, opened 2021-09-06
11923 prevent dup response sets

DevneyHamilton:
Patch to bring back duplicating loans as a feature. Root issue was that for at least 5 yrs, we have had 'has_one' associations for 'criteria' and 'post_analysis' on the loan model. (see https://github.com/sassafrastech/madeline/commit/bb6f1a401b84e2548918474ece56c039cd10bc99). It creates a bug with amoeba_dup, which we use for duplicating projects and loans (loans are subclass of project) because amoeba_dup for projects is configured (in project.rb) to include all associations that are not explicitly excluded. So it copies once for the 'has_one' criteria line, once for the 'has_one' post_analysis line, and once for the has_many response_sets line. when it copies for the has_many response_sets line, it copies the criteria and post analysis response sets it already saw and copied.

SOLUTION:
exclude criteria and post analysis associations in the amoeba duplication config for project with an explanation. this keeps the change confined to the duplication code instead of doing a change (like replacing the has_ones with getters and setters) that will affect business planning, and may also conflict more with refactor tom is working on.

### DIFF
--- a/spec/factories/loans_factory.rb
+++ b/spec/factories/loans_factory.rb
@@ -162,20 +162,15 @@ FactoryBot.define do
       end
     end
 
-    trait :with_copies do
-      after(:create) do |loan|
-        create(:loan, original: loan)
-      end
-    end
-
     # Assumes a LoanQuestionSet with name 'loan_criteria' and questions `summary` and `workers` exists.
     trait :with_criteria_responses do |loan|
       after(:create) do |loan|
-        loan.criteria = create(:response_set,
+        create(:response_set,
           kind: 'criteria',
           loan: loan,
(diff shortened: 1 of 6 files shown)


spec/factories/loans_factory.rb
```diff
@@ -162,20 +162,15 @@
       end
     end
 
-    trait :with_copies do
-      after(:create) do |loan|
-        create(:loan, original: loan)
-      end
-    end
```
DevneyHamilton, line 169: removing this because it never happens in the codebase and it circumvents our duplicator code which is what we want to actually test.

PR #913, opened 2021-09-07
11840 catch error improve reporting

DevneyHamilton:
Catch and do some logging to get more info about the Matching Transaction Errors comign from QB. First pass at merging qb error handling so less duplication bt controllers and jobs.


app/controllers/admin/accounting/transactions_controller.rb
```diff
@@ -79,7 +79,8 @@ def process_transaction_and_handle_errors
     # We use it because we want to rollback transaction creation or updates if there are errors.
     ActiveRecord::Base.transaction do
       if @transaction.save
-        if (error_msg = handle_qb_errors { run_updater(project: @transaction.project) })
+        loan = @transaction.loan
+        if (error_msg = Accounting::QB::ErrorHandler.new(loan: loan).handle_qb_errors { run_updater(project: loan) })
```
smoyte, line 83: Neato!

app/jobs/quickbooks_update_job.rb
```diff
@@ -11,28 +11,9 @@ def perform(_job_params)
...
-        # But we don't expose the original error message to the user since it won't be intelligble
-        # and could be a security issue.
-        Accounting::SyncIssue.create!(level: :error, loan: loan, message: :system_error_recalc)
-
-        # We want to receive a loud notification about an unhandled error.
-        # If we find this is often generating a lot of similar errors
-        # then we should really start using Sentry or some other service to group them.
-        notify_of_error(e, data: {context: "Unhandled error during loan update", loan_id: loan.id})
-      end
+      Accounting::QB::ErrorHandler.new(loan:loan, in_background_job: true).handle_qb_errors { updater.update_loan(loan) }
```
smoyte, line 14: Nice optional boolean flag here 😎

app/models/accounting/qb/error_handler.rb
```diff
@@ -0,0 +1,79 @@
...
+               Quickbooks::ThrottleExceeded,
+               Quickbooks::TooManyRequests,
+               Quickbooks::IntuitRequestException => e
+          Rails.logger.error e
+          ExceptionNotifier.notify_exception(e)
+          error_msg =I18n.t("quickbooks.misc", msg: e)
+        rescue Accounting::QB::NegativeBalanceError => e
+          Rails.logger.error e
+          error_msg =I18n.t("quickbooks.negative_balance", amt: e.prev_balance)
+        rescue StandardError => e
```
smoyte: Should you just re-raise the error here if not in bg job? That was the behavior earlier... It might help to include this in your list—for each error, is it logged, sent out via notification, saved in an issue, unhandled/reraised, or all the above?

app/models/accounting/qb/error_handler.rb
```diff
@@ -0,0 +1,79 @@
...
+                          line_item_changes: txn.line_items.map { |li| li.previous_changes }}
+          Rails.logger.error("Accounting::QB::IntuitRequestError #{intuit_error_type} #{e.message}. Txn date: #{txn.txn_date}. Txn qb id: #{txn.qb_id}. Changes: #{changes_hash}")
+          error_msg = I18n.t("quickbooks.#{intuit_error_type}",
+                        date: txn.txn_date,
+                        qb_id: txn.qb_id)
+        rescue Quickbooks::MissingRealmError,
+               Accounting::QB::NotConnectedError,
+               Quickbooks::AuthorizationFailure => e
+          Rails.logger.error e
+          error_msg = I18n.t("quickbooks.authorization_failure", settings: settings_link).html_safe
```
smoyte: In the previous code, in a bg job, these errors would get handled by the "catch all" below, and an issue would get created. Now, as I understand it, these would just get piped to null since the error_msg is ignored in the bg job code, right?

I believe the same is true for the next two `rescue` blocks as well, and for DataResetRequired above.

I think you could unify this behavior if you looked at each of the error types and considered what could work for both scenarios. For instance, for the auth failure ones, you could add a sync issue in both cases. For the throttle, too many requests, etc., I think it would be ok to let them fall to the catch-all.

I think you may want to make a list of each of these and think about what needs to happen. I also am not sure about how to properly do error notification. Right now it's happening in two ways and I think one of them would result in a NoMethodError since `notify_of_error` is not defined here.

app/models/accounting/qb/error_handler.rb
```diff
@@ -0,0 +1,79 @@
...
+        rescue StandardError => e
+          # If there is an unhandled error updating an individual loan, we don't want the whole process to fail.
+          # We let the user know that there was a system error and we've been notified.
+          # But we don't expose the original error message to the user since it won't be intelligble
+          # and could be a security issue.
+          Accounting::SyncIssue.create!(level: :error, loan: @loan, message: :system_error_recalc)
+          # We want to receive a loud notification about an unhandled error.
+          # If we find this is often generating a lot of similar errors
+          # then we should really start using Sentry or some other service to group them.
+          notify_of_error(e, data: {context: "Unhandled error during loan update", loan_id: @loan.id})
```
smoyte: I don't think this method is defined in this class...

app/models/accounting/qb/transaction_reconciler.rb
```diff
@@ -30,6 +33,10 @@ def reconcile(transaction)
 
       attr_reader :qb_connection, :principal_account, :qb_division
 
+      def handle_qb_intuit_request_exception(txn, e)
+        raise IntuitRequestError.new(transaction: txn, message: e.message)
```
smoyte: At first glance it's not clear what the difference is between IntuitRequestError and IntuitRequestException. It seems like the former wraps the latter and adds a transaction object for the error handler to use? Maybe call it WrappedIntuitRequestException or AnnotatedIntuitRequestException?

config/locales/en/en.yml
```diff
@@ -536,12 +538,14 @@ en:
...
     quickbooks_accounts_not_selected: "QuickBooks accounts not set. Please select QuickBooks accounts on the Accounting Settings page."
     quickbooks_unavailable_fetch: "QuickBooks was temporarily unavailable. Your transaction data may be out of date. Please try syncing again later."
     quickbooks_unavailable_recalc: "QuickBooks was temporarily unavailable during calcuations for this loan, so transaction data may be partially incorrect. Please try syncing again later to correct this."
     system_error_fetch: "There was an unexpected system error during fetching of QuickBooks data. The Madeline support team has been notified. Your transaction data may be out of date."
     system_error_recalc: "There was an unexpected system error during calculations for this loan. The Madeline support team has been notified."
+    unknown_intuit_request_exception: "An unexpected error occurred between QB and Madeline on transaction (dated %{date} with QB id %{qb_id}). Due to this error, Madeline is unable to complete your action at this time. The Madeline support team has been notified."
```
smoyte, line 551: This looks like it will be helpful.

smoyte (commented):
Thanks for taking this on Devney! I appreciate that you are tackling this refactor instead of just adding to the duplication!!! I also note that you said this was a 'first pass'.

I do think this is a tricky refactor and we need to be careful here not to lose issue logging functionality, and because we don't have good test coverage of error handling flows. Please see my comments and let me know what you think. Happy to meet with you if you like.

I like where this is headed!

PR #918, opened 2021-09-16
Feature/11890 division question sets (phase 3)

smoyte:


app/models/loan.rb
```diff
@@ -13,8 +12,6 @@ class Loan < Project
   belongs_to :organization
   belongs_to :currency
   belongs_to :representative, class_name: 'Person'
-  has_one :criteria, -> { where("response_sets.kind" => 'criteria') }, class_name: "ResponseSet"
-  has_one :post_analysis, -> { where("response_sets.kind" => 'post_analysis') }, class_name: "ResponseSet"
```
smoyte, line 16: Reducing coupling of Loan to ResponseSet here.

app/models/question_set.rb
```diff
@@ -1,25 +1,28 @@
 class QuestionSet < ApplicationRecord
   include Translatable
 
-  has_closure_tree_root :root_group, class_name: "Question"
+  KINDS = %i[loan_criteria loan_post_analysis].freeze
+
+  belongs_to :division, inverse_of: :question_sets
+  has_many :response_sets, inverse_of: :question_set, dependent: :restrict_with_exception
 
-  translates :label
```
smoyte, line 6: This was not getting used it seems. We may end up adding it back later if QuestionSets become editable but for now it's unnecessary.

app/views/admin/loans/_questions.html.slim
```diff
@@ -1,80 +1,69 @@
...
+    = t("questions.none_for_loan_division", division_name: @loan_division.name)
+  - else
+    - if @question_sets.size > 1
+      / Switch at top allowing toggling between question sets
+      .filter-switch.centered-text
+        .btn-group
+          - @question_sets.each do |qset|
+            = link_to t("question_sets.#{qset.kind}"),
+              admin_loan_tab_path(@loan.id, tab: "questions", qset: qset.id),
+              class: "btn btn-default #{'active' if @question_set == qset}"
```
smoyte, line 14: New here is just the ability to handle zero or one question set, and a loop over question sets for the filter switch instead of hard coding them.

app/views/admin/questions/index.html.slim
```diff
@@ -1,45 +1,50 @@
...
-        class: "btn btn-default #{'active' if @set_name == 'post_analysis'}"
+  - if @question_sets.empty?
+    = t("questions.none_for_current_division")
+  - else
+    .filter-switch
+      .btn-group
+        - @question_sets.each do |qset|
+          = link_to t("question_sets.#{qset.kind}"),
+            admin_questions_path(qset: qset.id),
+            class: "btn btn-default #{'active' if @question_set == qset}"
```
smoyte, line 12: Similar changes here as for biz planning page.

app/views/layouts/admin/main.html.slim
```diff
@@ -25,6 +25,7 @@ html lang="en-US"
           dateFormats: #{json(I18n.t("date.moment_formats"))},
           locale: #{json(I18n.locale)}
         });
+        window.popoverView = new MS.Views.PopoverView({el: "body"});
```
smoyte, line 28: Separated this out for separation of concerns and b/c we need to pass it to QuestionsView.

db/migrate/20210908204646_add_loan_prefix_to_response_set_kinds.rb
```diff
@@ -0,0 +1,9 @@
+class AddLoanPrefixToResponseSetKinds < ActiveRecord::Migration[6.1]
+  def up
+    execute("UPDATE response_sets SET kind = 'loan_' || kind")
```
smoyte, line 3: Goal here is to get rid of all the places in the code where we are either adding on this prefix or stripping it 🤮

db/migrate/20210910164556_convert_to_jsonb.rb
```diff
@@ -0,0 +1,45 @@
...
+    convert_to_jsonb(:accounting_sync_issues, :custom_data)
+    convert_to_jsonb(:accounting_transactions, :quickbooks_data)
+    convert_to_jsonb(:data_exports, :data)
+    convert_to_jsonb(:divisions, :custom_data)
+    convert_to_jsonb(:divisions, :locales)
+    convert_to_jsonb(:organizations, :custom_data)
+    convert_to_jsonb(:projects, :custom_data)
+    convert_to_jsonb(:response_sets, :custom_data)
+    convert_to_jsonb(:tasks, :activity_message_data)
+    convert_to_jsonb(:tasks, :custom_error_data)
```
smoyte, line 16: Had to convert response sets in order to make the migration performant. jsonb is better than json so did it for all json fields at once.

app/views/layouts/admin/_site_menu.html.slim
```diff
@@ -26,7 +26,8 @@ li.nav-item
       li = link_to t('menu.divisions'), admin_divisions_path
     - if policy(Documentation).index?
       li = link_to t('menu.documentations'), admin_documentations_path
-    li = link_to t('menu.questions'), admin_questions_path
+    - unless selected_division.nil?
```
DevneyHamilton, line 29: great, thanks for just hiding this link

spec/system/admin/questions_flow_spec.rb
```diff
@@ -105,20 +105,33 @@
...
     expect(page).not_to have_css(".jqtree-title", text: "Stonks")
   end
 
+  scenario "in 'All Divisions' mode, don't allow" do
+    visit(admin_person_path(actor)) # Don't start at the root path b/c that's where we get redirected.
+
+    within("#site-menu") do
+      find("a", text: "Manage").click
+      expect(page).to have_content("Accounting Settings") # Wait for dropdown to show
+      expect(page).not_to have_content("Questions")
```
DevneyHamilton: thanks for this spec coverage

app/models/question_set.rb
DevneyHamilton: thanks for taking this out

app/models/division.rb
```diff
@@ -11,6 +11,7 @@ class Division < ApplicationRecord
   has_many :people, dependent: :restrict_with_exception
   has_many :organizations, dependent: :restrict_with_exception
 
+  has_many :question_sets, inverse_of: :division, dependent: :restrict_with_exception
```
DevneyHamilton, line 14: thanks for using restrict_with_exception here.

app/models/question_set.rb
```diff
@@ -1,12 +1,22 @@
...
   belongs_to :division, inverse_of: :question_sets
 
   has_closure_tree_root :root_group, class_name: "Question"
 
   after_create :create_root_group!
 
+  def self.find_for_division(division)
+    self_and_ancestor_ids = division.self_and_ancestors.pluck(:id)
+    KINDS.map do |kind|
+      candidates = where(internal_name: kind, division_id: self_and_ancestor_ids).index_by(&:division_id)
```
DevneyHamilton: this may be clear when i keep reviewing, but is 'internal' referring to the name just being in our codebase, or 'Internal' the la base argentina application?

app/controllers/concerns/questionnaire_renderable.rb
```diff
@@ -1,10 +1,14 @@
 module QuestionnaireRenderable
   extend ActiveSupport::Concern
 
+  # Prepares variables for rendering a questionnaire.
```
DevneyHamilton, line 4: thanks for adding this comment

app/models/loan.rb
```diff
@@ -59,6 +57,16 @@ def self.txn_mode_choices
     TXN_MODES
   end
 
+  # These methods should go away soon.
```
DevneyHamilton: i am looking fwd to them going away; is this a recommendation for work after this push on argentina migration?

app/views/admin/loans/_questions.html.slim
```diff
@@ -4,16 +4,12 @@ section.questionnaires
...
     .btn-group
-      = link_to t('loan.criteria'),
-        admin_loan_tab_path(@loan.id, filter: "loan_criteria", tab: "questions"),
-        class: "btn btn-default #{'active' if @attrib == 'loan_criteria'}",
-        data: {filter: "loan_criteria"}
-      = link_to t('loan.post_analysis'),
-        admin_loan_tab_path(@loan.id, filter: "loan_post_analysis", tab: "questions"),
-        class: "btn btn-default #{'active' if @attrib == 'loan_post_analysis'}",
-        data: {filter: "loan_post_analysis"}
+      - @question_sets.each do |qset|
```
DevneyHamilton: great, glad to have one less thing (The attr) that has to be 'threaded through' a request

db/migrate/20210909142344_add_question_set_to_response_sets.rb
```diff
@@ -0,0 +1,12 @@
+class AddQuestionSetToResponseSets < ActiveRecord::Migration[6.1]
+  def change
+    add_reference :response_sets, :question_set, foreign_key: true, index: true
+    reversible do |dir|
+      dir.up do
```
DevneyHamilton: is there a way to write this so the down process is clearer? 

-

db/migrate/20210907201954_add_division_to_question_sets.rb
```diff
@@ -0,0 +1,11 @@
+class AddDivisionToQuestionSets < ActiveRecord::Migration[6.1]
+  def change
+    add_reference :question_sets, :division, foreign_key: true, index: true
+    reversible do |dir|
```
DevneyHamilton: is there a way to write this so the down process is clearer?

db/migrate/20210908204646_add_loan_prefix_to_response_set_kinds.rb
```diff
@@ -0,0 +1,9 @@
...
+  end
+
+  def down
+    execute("UPDATE response_sets SET kind = substring(kind from 6)")
```
DevneyHamilton, line 7: it's been a long time since i've seen something 1 indexed! interesting sql does that.

app/views/admin/loans/print.html.slim
```diff
@@ -35,8 +35,8 @@ section data-container="loan"
           response_set: @response_set
         - if @response_set.question_set.summary_questions?
           = render "admin/loans/questionnaires/questionnaire_form",
-            response_set: @response_set, summary_view: true, attrib: :criteria
+            response_set: @response_set, summary_view: true
```
DevneyHamilton, line 38: happy threading attrib through this process is going away!

db/migrate/20210910164556_convert_to_jsonb.rb
DevneyHamilton: ah, great - thank you.

DevneyHamilton (commented):
Thank you for all this work - it's a LOT! I did a first pass on reviewing. The changes to question sets and response sets pages are a big improvement and will make future work easier.  I haven't QAed or taken a really close look at the migrations of response set data yet. I can do that when I'm back and hope you're not blocked on this before then. I saw you still need to talk to the La Base Argentina team too.

db/migrate/20210907201954_add_division_to_question_sets.rb
smoyte: I'm not sure what you mean. The down process is trivial since it's just removing a column. I guess we could make separate up and down methods. Would that work?

db/migrate/20210908204646_add_loan_prefix_to_response_set_kinds.rb
smoyte: :)

db/migrate/20210909142344_add_question_set_to_response_sets.rb
smoyte: Similar answer to above :)

db/migrate/20210907201954_add_division_to_question_sets.rb
DevneyHamilton: Yes, please - separate up and down methods would be great. I know Rails added this reversible block, but for future devs reading this (in the unfortunate case we have to revisit this migration) i think it's better to have down and up separate.

db/migrate/20210909142344_add_question_set_to_response_sets.rb
DevneyHamilton: Same here, yes please to defining up and down methods.

db/migrate/20210910170107_add_response_sets_loans_foreign_key.rb
```diff
@@ -0,0 +1,11 @@
+class AddResponseSetsLoansForeignKey < ActiveRecord::Migration[6.1]
+  def change
+    reversible do |dir|
```
DevneyHamilton: same here, if the up/down code is different, it's easier to read if they are separate methods.

db/schema.rb
```diff
@@ -417,18 +419,20 @@
...
-    t.string "kind"
+    t.jsonb "custom_data"
     t.integer "loan_id", null: false
     t.integer "lock_version", default: 0, null: false
+    t.bigint "question_set_id", null: false
     t.datetime "updated_at", null: false
     t.integer "updater_id"
-    t.index ["loan_id", "kind"], name: "index_response_sets_on_loan_id_and_kind", unique: true
+    t.index ["loan_id", "question_set_id"], name: "index_response_sets_on_loan_id_and_question_set_id", unique: true
+    t.index ["question_set_id"], name: "index_response_sets_on_question_set_id"
```
DevneyHamilton, line 437: very very happy to have response sets belong to loans and qsets, and not rely on 'kind' anymore. thank you!

app/models/project.rb
```diff
@@ -41,12 +41,6 @@ class Project < ApplicationRecord
     exclude_association :copies
     exclude_association :transactions
     exclude_association :sync_issues
-    # below we exclude criteria and post analysis associations (see loan.rb) because they refer
```
DevneyHamilton, line 44: nice to be rid of this too!

DevneyHamilton:
Hi Tom, 
Thanks for patience with this review. As I said before, it's a ton of work and leaves things in much better shape overall - thank you! If you could please do those changes to have up/down instead of reversible blocks, that would be great for future readability.  Am I understanding right that these changes are prereqs for the argentina migration, and they will be merged into develop AND into a branch that you're putting on the argentina migration qa server? Is there any way they can go to the qa server first? If not, I want to think more carefully about merging this into develop and release timeline. My concern is that I want TWW staff to look over these changes on staging before they go to prod, which they are not in the habit of doing right now.

PR #921, opened 2021-09-20
11958: Add inception field

smoyte:
New field to track how the organization was formed. Replaces the old `is_recovered` flag (which was never used) with something slightly more general and usable.


DevneyHamilton (approved):
Thank you!

DevneyHamilton (changes requested):
Thank you - do you have time to add specs about this? I know it's a pretty straightforward form field, but it'd be great to keep increasing our spec coverage as we go.

DevneyHamilton:
> Thank you!

Oops, approved a little to fast - I think it'd be good to add a quick spec here or add this field to existing specs.

smoyte:
@DevneyHamilton good call. I added a system spec. Does that cover it sufficiently? Covering model stuff would really fall to coverage of the option set system, which I'm assuming we have somewhere, and if not, I think I'd say that's out of scope for right now.

PR #922, opened 2021-09-22
11959 add byte order mark

smoyte:


DevneyHamilton (commented):
Thanks for figuring out that we need this! Mainly I'd just like to know more context for this in case a similar issue shows up in the future. Did this come from an issue testing on Madeline data, or only in Argentina team's data? Is this solution specific to an operating system, or anything like that? (feel free to point me to documentation about this fix as an answer to that).

smoyte:
This BOM is required any time you want to use CSV with accented chars as far as I know. I knew about it from NEMO.

PR #923, opened 2021-09-22
11961: Remove representative field/column

smoyte:


DevneyHamilton (approved):
Thanks! Excited we can remove a field :)

PR #925, opened 2021-09-27
Add legacy_id to response_sets and questions

smoyte:
This is to support the Argentina migration. Thanks!


DevneyHamilton:
> This is to support the Argentina migration. Thanks!

@smoyth just for future reference, can you please just confirm (or correct) my assumption that the legacy ids in this case are the response set and question ids from the labase argentina database?

DevneyHamilton (approved):
Glad we're keeping these just in case. I do have a question in the "conversation" for you. Thanks!

smoyte:
> > This is to support the Argentina migration. Thanks!
> 
> @smoyth just for future reference, can you please just confirm (or correct) my assumption that the legacy ids in this case are the response set and question ids from the labase argentina database?

Yes, correct. Just keeping them around in case we need to trace something down later.

PR #929, opened 2021-10-04
Allow br tag in questionnaire show view

smoyte:
Very small change to allow br tags in the show view to support full text data imported from old system. Also to support when user presses <Shift+Enter> in rich text editor to insert a line break as opposed to a paragraph break. Not sure if this would work properly in current system.

### DIFF
--- a/app/views/admin/loans/questionnaires/_answer.html.slim
+++ b/app/views/admin/loans/questionnaires/_answer.html.slim
@@ -42,7 +42,7 @@
   / Display response as rich text.
 
   / Specify which html tags are allowed.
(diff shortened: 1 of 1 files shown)


DevneyHamilton:
> Very small change to allow br tags in the show view to support full text data imported from old system. Also to support when user presses <Shift+Enter> in rich text editor to insert a line break as opposed to a paragraph break. Not sure if this would work properly in current system.

I tried it in the rich text editor and it looks like its all working. Thanks!
Screen Shot 2021-10-05 at 11 28 41 AM
:

DevneyHamilton (approved):

PR #930, opened 2021-10-04
11993 zero interest txn match txn fix

DevneyHamilton:
See https://docs.google.com/document/d/1oy7WHpMWaqdrsHu3JDJ1lC8DGp4HEI2GcSqjxlJERBk/edit for context. 

QB API turns any 0.00 line item's posting type to Debit even if we send Credit. This was causing problems with matched txns because Madeline was trying to change the posting type on matched txns which is understandably not allowed.  This PR changes interest calculator, which handles making/editing line items we send to qb, set posting type to debit when amt is zero. The journal entry extractor was already set up tohandle a zero debit on interest when setting the repayment type on the txn's return trip from qb to madeline.

### DIFF
--- a/spec/models/accounting/qb/journal_entry_extractor_spec.rb
+++ b/spec/models/accounting/qb/journal_entry_extractor_spec.rb
@@ -352,7 +352,7 @@ describe Accounting::QB::JournalEntryExtractor, type: :model do
           {'line_items' =>
             [{'id' => '0',
               'description' => 'Eba',
-              'amount' => '10.99',
+              'amount' => '0.00',
               'detail_type' => 'JournalEntryLineDetail',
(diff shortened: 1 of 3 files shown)


spec/models/accounting/qb/journal_entry_extractor_spec.rb
```diff
@@ -480,7 +480,7 @@ def expect_line_item_amounts(amounts)
           {'line_items' =>
             [{'id' => '0',
               'description' => 'Eba',
-              'amount' => '10.99',
+              'amount' => '0.0',
```
DevneyHamilton, line 483: this is an unneeded change, but doesn't affect the logic of this spec.

PR #932, opened 2021-10-07
11994 notifications subdivisions

smoyte:
Moves log notification into a background job for separation of concerns.

Adds system specs for log create and person create & update, both previously uncovered.

Adds notification source attribute to User.

### DIFF
--- a/app/jobs/log_notification_job.rb
+++ b/app/jobs/log_notification_job.rb
@@ -1,0 +1,10 @@
+class LogNotificationJob < ApplicationJob
+  def perform(log)
+    division = log.division
+    return unless division.notify_on_new_logs?
+
+    division.users.each do |user|
+      NotificationMailer.new_log(log, user).deliver_now
+    end
+  end
+end
--- a/app/jobs/log_notification_job.rb
+++ b/app/jobs/log_notification_job.rb
@@ -3,8 +3,14 @@ class LogNotificationJob < ApplicationJob
     division = log.division
     return unless division.notify_on_new_logs?
 
-    division.users.each do |user|
-      NotificationMailer.new_log(log, user).deliver_now
+    ancestors = division.ancestors.pluck(:id)
+    base_scope = Person.with_system_access.joins(:user).where.not(users: {notification_source: "none"})
+    own_division = base_scope.where(division: division)
+    subdivision = base_scope.where(division: ancestors).where(users: {notification_source: "home_and_sub"})
+    people = own_division.or(subdivision)
+
+    people.each do |person|
+      NotificationMailer.new_log(log, person.user).deliver_now
     end
   end
 end
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -33,7 +33,7 @@ class Person < ApplicationRecord
   validate :user_valid
 
   # Transient attributes to facilitate user management
-  attr_writer :access_role
+  attr_writer :access_role, :notification_source
   attr_accessor :password, :password_confirmation
 
   before_save :update_full_name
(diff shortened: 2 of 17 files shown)


spec/jobs/log_notification_job_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      it "sends no emails" do
+        expect { described_class.perform_now(log) }.to change { deliveries.size }.by(0)
+      end
+    end
+  end
+
+  context "when log is on subdivision" do
+    let(:log_division) { subdivision }
+
+    context "when user has notification_source home_only" do
```
DevneyHamilton, line 44: why do we not set the notification source to home_only here?  is it possible this spec is passing with notification source actually being "none"?

app/jobs/log_notification_job.rb
```diff
@@ -0,0 +1,16 @@
+class LogNotificationJob < ApplicationJob
+  def perform(log)
+    division = log.division
+    return unless division.notify_on_new_logs?
+
+    ancestors = division.ancestors.pluck(:id)
+    base_scope = Person.with_system_access.joins(:user).where.not(users: {notification_source: "none"})
+    own_division = base_scope.where(division: division)
+    subdivision = base_scope.where(division: ancestors).where(users: {notification_source: "home_and_sub"})
+    people = own_division.or(subdivision)
```
DevneyHamilton, line 10: cool!

app/models/person.rb
```diff
@@ -60,6 +60,19 @@ def agent_projects
...
 
+  def notification_source
+    # I am just following how other user attributes are persisted and accessed here, but it is not
+    # a very nice way of doing things. For models that are nested inside one another,
+    # accepts_nested_attributes_for is the better way to do this. But I also don't really see the
+    # benefit of having separate Person and User classes in this project, whereas this split adds
+    # a lot of extra complexity. Ultimately I think the two should be combined.
+    # So I didn't think it was worthwhile to refactor this class
+    # to more gracefully accept nested attributes at this time.
+    return @notification_source if defined? @notification_source
```
DevneyHamilton, line 71: I understand this choice, thank you for noting it. I also do not like having these separate classes. For now could you please move this larger explanation into a comment on the PR? And leave a shorter comment here like "Todo: Refactor to merge User and Person classes. Below is workaround for user attires being persisted and accessed in Person class."?

DevneyHamilton: i do use prs a lot for forensics for 'why is the code the way it is?' so if i have questions about this stuff, which i probably will,  i will come back to this PR and so this info won't be lost in a PR comment.

spec/factories/users_factory.rb
```diff
@@ -1,3 +1,7 @@
+# WARNING: This factory may generate double users in some cases. It seems the better way to get users
+# is to go through the Person factory. The create_member and create_admin methods attest to this.
+# Those should be removed since they are not easy to find (principle of least surprise).
+# This factory should perhaps be removed for the same reason.
 FactoryBot.define do
```
DevneyHamilton, line 5: thanks for flagging this, ok to leave this whole comment in the code for now.

spec/jobs/log_notification_job_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  let!(:decoy2) { create(:person, :with_member_access, :with_password, division: division) }
+  let!(:decoy3) { create(:person, :with_member_access, :with_password, division: division) }
+  let(:log) { create(:project_log, division: log_division) }
+  let(:deliveries) { ActionMailer::Base.deliveries }
+
+  before do
+    [person1, person2, decoy2].each { |p| p.user.update!(notification_source: notification_source) }
+    decoy2.update!(has_system_access: false)
+    decoy3.user.update!(notification_source: "none")
+  end
```
DevneyHamilton, line 20: thanks for the decoys in this spec!

DevneyHamilton (dismissed):
Thank you for adding this feature, and for adding spec coverage. Cool scope logic in the job :)

DevneyHamilton (commented):
Thank you for adding this feature, and for adding spec coverage. Cool scope logic in the job :). I made one request to please pare down a comment, and otherwise this is good to merge.

app/models/person.rb
```diff
@@ -60,6 +64,12 @@ def agent_projects
     Project.where("primary_agent_id = ? OR secondary_agent_id = ?", id, id)
   end
 
+  def notification_source
```
smoyte, line 63: In creating this method I am just following how other user attributes are persisted and accessed here, but it is not a very nice way of doing things. For models that are nested inside one another, accepts_nested_attributes_for is the better way to do this. But as noted above, I think the two should be combined. So I didn't think it was worthwhile to refactor this class to more gracefully accept nested attributes at this time.

PR #933, opened 2021-10-08
12000 per division data exports

smoyte:
- Refactors data export classes to eliminate copied code.
- Improves performance of data exports by replacing O(N) header index lookup with O(1).
- Adds text, boolean, not applicable to data exports.
- Cleans up and comments on confusing sets of different data types and value types in Question and Response classes.
- Adds missing spec for enhanced export
- Restricts exported data to current division
- Restricts exported question sets to those related to the exported loans

Could use some QA to ensure that the division gets correctly passed into the export object and that the export looks ok with real live data.

### DIFF
--- a/app/models/data_export.rb
+++ b/app/models/data_export.rb
@@ -15,11 +15,27 @@ class DataExport < ApplicationRecord
     "enhanced_loan_data_export" => "EnhancedLoanDataExport"
   }
 
-  # Process data should be defined on subclasses,
-  # generate a 2D array with the desired data,
-  # and save it in the `data` field
+  def self.model_name
+    ActiveModel::Name.new(self, nil, "DataExport")
+  end
+
   def process_data
-    raise NotImplementedError
+    @child_errors = []
+    data = []
+    data.concat(header_rows)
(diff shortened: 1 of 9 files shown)


DevneyHamilton:
QAed this exporting from SeedCommons. Notes: we will need to rely on TWW admin team to communicate amongst themselves that exports can be done be division from the division. SeedCommons enhanced export had 314 columns, which is a lot, but not too bad considering. We may want to add a note in the UI that enhanced data exports are not recommended for root division (since they will have > 1000 columns now with the questions separated, right?)

DevneyHamilton (approved):
Thank you very much Tom for this work. I really appreciate your commitment to leaving Madeline's codebase in better shape with these refactors.

QAed with SeedCommons and it all looked good!

smoyte:
> We may want to add a note in the UI that enhanced data exports are not recommended for root division (since they will have > 1000 columns now with the questions separated, right?)

Yes I think you're right. Can I leave the note adding part to you all, perhaps after we demo this to Brendan?

PR #938, opened 2021-10-29
Fix division refs in dropdown scoping

smoyte:
I realized that I should have been using the record's division and not the currently selected division when scoping this list. Sorry about the oversight. This should fix it.


DevneyHamilton (approved):

smoyte:
Failure is a pesky flapper. 😢  Merging.

DevneyHamilton:
> Failure is a pesky flapper. 😢 Merging.

passed on my local.

PR #939, opened 2021-11-02
Fix timeline step moves bug

smoyte:


smoyte:
Spec failure is flapper

PR #941, opened 2021-12-03
10486 remove old migration code

smoyte:


DevneyHamilton:
For people looking at this in the future: several years ago there had been an intention to migrate data off of a mysql db into Madeline. The timeline for that migration ended up being pushed back by several years  and it actually happened in fall 2021. this pr is removing some mysql code and some code referring to the old data that was kept around for this purpose.

PR #952, opened 2022-02-16
12118: set interest amt to 0 if prin bal is zero, add spec coverage o…

DevneyHamilton:
…f this


DevneyHamilton:
QA'ed on staging

PR #958, opened 2022-03-03
12134 fix linked document markup always showing

DevneyHamilton:
hotfix level bug - answer.html.slim was checking response.has_url. has_url was true for all responses because url (and start cell and end cell) are added as types to all responses.


DevneyHamilton:
This was merged earlier from https://github.com/sassafrastech/madeline/pull/959/files.

